PR #38, opened 2016-04-06
Provide XCode project file for iOS sources and harmonize directory structure

philikon:
See individual commit messages for details.

If you're good with this change, I'll be happy to update the iOS installation docs.

Thanks!


zxcpoiu:
after this commit, can we simply use `rnpm link` now?

it wasn't also on android, maybe due to the module name.

anyway, +1

philikon:
> after this commit, can we simply use `rnpm link` now?

I'm not familiar with `rnpm`, so I can't answer that. As I pointed out in the commit message for 80851f9, you'll still have to manually add `libjingle` to your app's XCode project and add all the frameworks that WebRTC needs to work.

oney:
Thanks! I will publish to nom later

oney:
@zxcpoiu Using rnpm should be good, but I don't try it yet.

oney:
@philikon Already update README https://github.com/oney/react-native-webrtc/commit/05e13caa8c29cbbd08e91210259a694fac0d2c9b
Please take a look, thanks.

philikon:
> @philikon Already update README 05e13ca

Looks great! Thanks!

PR #42, opened 2016-04-09
Make PeerConnection and MediaStream proper EventTargets

philikon:
Implements the `EventTarget` spec for PeerConnection and MediaStream, so that code that uses `addEventListener`/`removeEventListener` will work out of the box. With this, `peerConnection.onfoobar = handler` and `peerConnection.addEventListener('foobar', handler)` are equivalent, just like in the browser (and the spec).


philikon:
Can we get this merged?

PR #43, opened 2016-04-09
Consolidation, cleanup, and spec compliance

### LINKED ISSUE #41: What's with the Foobar v. FoobarImpl stuff?
I don't quite see the point of `RTCPeerConnectionBase` deferring `someMethod()` to `someMethodImpl()` which is then implemented in `RTCPeerConnection`. This seems to be modeled on React Native's `WebSocketModule` implementation, which only does this so it can reuse the code for the websocket-based debugging stuff. What's the reasoning do introduce this complication in react-native-webrtc? Can we get rid of it?

philikon:
- This builds upon #42, which makes `PeerConnection` and `MediaStream` proper event targets
- Add or correct several Flow type annotations. -- Not that it matters right now since Flow isn't turned on for this project, but it would be nice to get there... In the meantime it's good source code documentation.
- Rename `RTCMediaStream` to `MediaStream` as per spec
- Fix #41 by folding `RTCPeerConnectionBase` into `RTCPeerConnection` and `getUserMediaImpl` into `getUserMedia`
- Ensure `DeviceEventEmitter` handlers are removed when the `RTCPeerConnection` is closed.


zxcpoiu:
you are the man!

philikon:
I'm just getting started :)

oney:
Look good. I will test it later.

MediaStreamTrack.js
```diff
@@ -7,37 +7,20 @@ var {
 
 var WebRTCModule = NativeModules.WebRTCModule;
 
-class SourceInfo {
+type SourceInfo = {
```
oney: What is `type` syntax? Can you give me a hint?

philikon: It's a Flow thing. See http://flowtype.org/docs/objects.html.

philikon:
Ping?

oney:
Merge!

PR #50, opened 2016-04-14
Complete MediaStreamTrack with Spec

### LINKED ISSUE #28: Can you add this method for stream object?
This would allow the voice mute，I really need this method，Thank you very much.
```javascript

localMediaStream.getAudioTracks()[0].enabled = false;

```

### LINKED ISSUE #48: RTCMediaStream releaseTrack or stopTrack
Hi, 

I was looking at adding a releaseTrack method to WebRTCModule+RTCMediaStream.m so that I can mute video/audio individually however RTCMediaStream is unavailable to edit or see the internals of. It seems that this part of the library is closed-source?

I see it has a [mediaStream addVideoTrack:videoTrack]; method..  Is there an equivalent releaseVideoTrack or stopVideoTrack method inside? Other WebRTC implementations implement stream.getVideoTracks()[0].enabled = false. However any JS changes made on the MediaStreamTracks are not passed into the objective C WebRTC implementation. 

Between releaseVideoTrack, stopVideoTrack or track.enabled = false, is there anyway for me to submit a PR for you given the limited visibility into the RTCMediaStream.m internals?

Or, if this is in development already, can you provide a status update? 

Thanks,

zxcpoiu:
see comment in commit:

this PR:
close #51 
close #48
close #28 

### Complete MediaTrack member and method

1. pass correspond properties to js when create track.

2. implement track.enabled = true/false which can be used to mute a track.
3. MediaStream.removeTrack() will invoke native method now.
4. implement track.stop()
   maybe unnecessary to invoke it if we explicitly release media stream.
   but some implementation would use it when hangup.
   this also supported in chrome and firfox.

   I'm not exactly understand what this means, but here are some hints:
   [Google Developers - MediaStream Deprecations](https://developers.google.com/web/updates/2015/07/mediastream-deprecations?hl=en)
   [W3C track.stop() procedure](https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack-stop)
   [W3C detached meaning](https://www.w3.org/TR/mediacapture-streams/#track-detached)

   what exaclty `detached` means in doc?

### NOTICE:

I'm confusing about what's the differences of behavior between `MediaStream.removeTrack()` and `MediaStreamTrack.stop()` ?

Currently my understanding is, `MediaStream.removeTrack()` will invoke native one `removeTrack`, thus the track will remove from stream. In the webrtc java source code, it iterate all tracks from mediaStream and `dispose()` ( aka `free()` ) each track.

`MediaStreamTrack.stop()` is just stop the track itself, will not be removed from mediaStream.

in short: both `track.stop()` and `mediaStream.removeTrack(track)` will STOP playing the track, and set `enable = false`;

if someone have a clear understanding, please fix those actions to consistent with the spec. thanks!

### Error Handling Object and Event

[js] add MediaStreamError Event type
this is just added the type object at the very first stage and nothing
todo with currently code base for now.

for future compatible with spec. maybe when we started to implement
[constrainable pattern interface](https://www.w3.org/TR/mediacapture-streams/#constrainable-interface)

and

[error
    handling](https://www.w3.org/TR/mediacapture-streams/#error-handling)

### DIFF
--- a/MediaStream.js
+++ b/MediaStream.js
@@ -40,6 +40,7 @@ class MediaStream extends EventTarget(MEDIA_STREAM_EVENTS) {
     if (index === -1) {
       return;
     }
+    WebRTCModule.mediaStreamTrackRelease(this.id, track.id);
     this._tracks.splice(index, 1);
     this.dispatchEvent(new MediaStreamTrackEvent('removetrack', {track}));
   }
--- a/MediaStreamError.js
+++ b/MediaStreamError.js
@@ -1,0 +1,16 @@
+'use strict';
+
+class MediaStreamError {
+
+  name: string;
+  message: ?string;
+  constraintName: ?string;
+
+  constructor(error) {
+    this.name = error.name;
+    this.message = error.message;
+    this.constraintName = error.constraintName;
+  }
+}
+
+module.exports = MediaStreamError;
--- a/MediaStreamErrorEvent.js
+++ b/MediaStreamErrorEvent.js
@@ -1,0 +1,14 @@
+'use strict';
+
+import type MediaStreamError from './MediaStreamError';
+
+class MediaStreamErrorEvent {
+  type: string;
+  error: ?MediaStreamError;
+  constructor(type, eventInitDict) {
+    this.type = type.toString();
+    Object.assign(this, eventInitDict);
+  }
+}
+
+module.exports = MediaStreamErrorEvent;
--- a/MediaStreamTrack.js
+++ b/MediaStreamTrack.js
@@ -1,11 +1,20 @@
 'use strict';
 
-var React = require('react-native');
-var {
-  NativeModules,
-} = React;
+const EventTarget = require('event-target-shim');
+const WebRTCModule = require('react-native').NativeModules.WebRTCModule;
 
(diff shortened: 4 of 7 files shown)


oney:
OK! Thanks.
I will spend some time implementing iOS part. Or any kind person would like to do it?

zxcpoiu:
haha, I will look into it.
Will ping you if I encounter any trouble.

dguillamot:
I added some of the ios code here : https://github.com/oney/react-native-webrtc/pull/51

I only had time to do what was necessary for my project (mediatrack setEnabled method). So it might be best to just use the objective c code I have and merge it into this architecture here. 

Cheers,

zxcpoiu:
I updated the PR, commit together with @dguillamot
please see the NOTICE in the PR comments.

zxcpoiu:
@oney 

I don't know how to get `readyState` on ios. ( method not found, btw it's readonly according to spec )
so hard-coded as `"live"` when we got the track first time.
since we don't use `readyState` for now so it won't break or hurt anything.

oney:
Awesome PR! I will test it later.

PR #51, opened 2016-04-14
ios+ js track setEnabled and media stream release track

dguillamot:
ios+ js track setEnabled and media stream release track

### DIFF
--- a/MediaStream.js
+++ b/MediaStream.js
@@ -40,6 +40,7 @@ class MediaStream extends EventTarget(MEDIA_STREAM_EVENTS) {
     if (index === -1) {
       return;
     }
+    WebRTCModule.mediaStreamTrackRelease(this.id, track.id);
     this._tracks.splice(index, 1);
     this.dispatchEvent(new MediaStreamTrackEvent('removetrack', {track}));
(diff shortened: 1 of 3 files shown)


philikon:
This will break the Android version if you don't at least add stub methods for `mediaStreamTrackRelease` and `trackSetEnabled` to the Android code. But perhaps you can just implement them? It's pretty straight forward.

dguillamot:
Thanks Phillikon. I only have time to do the ios version right now, but could look at Android version in 1-2 weeks. However there is a separate PR for android that was prepared yesterday from another contributor : https://github.com/oney/react-native-webrtc/pull/50

So, I think combining these two should get the job done. I think the other contributor did more JS-side infrastructure if you wanted to use that, and then you can just take the objective c code from this PR as needed.

philikon:
Oh, you're right, that's great news! Would be great if @zxcpoiu could combine your iOS code with his Android stuff.

zxcpoiu:
ok, I will do that!

zxcpoiu:
ok, I updated the PR, commit together with @dguillamot

PR #69, opened 2016-05-10
[android] prevent crash if object is null

zxcpoiu:
check object is null when get() from SparseArray, such as

* mPeerConnections
* mMediaStreams
* mMediaStreamTracks
* mDataChannels

if object is null, log without touch anything to prevent crash.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -403,8 +403,9 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
     @ReactMethod
     public void mediaStreamTrackStop(final String id) {
         final int trackId = Integer.parseInt(id);
-        MediaStreamTrack track = mMediaStreamTracks.get(trackId);
+        MediaStreamTrack track = mMediaStreamTracks.get(trackId, null);
         if (track == null) {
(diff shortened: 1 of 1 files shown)


zxcpoiu:
@oney 
I want to implement ios side as well, to be clear, will following checks work on objective C?

```
RTCMediaStreamTrack *track = self.tracks[trackID];
if (track) {
    // do something with track
} else {
    NSLog(@"track is null");
    return;
}
```

```
RTCPeerConnection *peerConnection = self.peerConnections[objectID];
if (peerConnection) {
    // do something with peerConnection
} else {
    NSLog(@"peerConnection is null");
    return;
}
```

zxcpoiu:
seems ios will not crash, is it correct?

oney:
I prefer using below codes.
```objective-c
RTCPeerConnection *peerConnection = self.peerConnections[objectID];
if (!peerConnection) {
    NSLog(@"peerConnection is nil");
    return;
}
```
I think we should check it on iOS, too.
```objective-c
RTCMediaStream *stream = self.mediaStreams[streamID]; // stream is nil
RTCPeerConnection *peerConnection = self.peerConnections[objectID];
BOOL result = [peerConnection addStream:stream]; // this should crash
```

zxcpoiu:
:ok_hand:

PR #71, opened 2016-05-17
Initialize RTCMediaStream with UUID

lyubomir:
The native counterpart of the implementation of getUserMedia in
WebRTCModule+RTCMediaStream.m initializes RTCMediaStream with 1 constant
label. That's a problem when RTCMediaStream instances gathered from
multiple getUserMedia (e.g. 1 getUserMedia for audio and 1 getUserMedia
for video)  calls are to be added to 1 RTCPeerConnection. The method
webrtc::PeerConnection::AddStream will not add (and return false for) a
MediaStreamIterface (i.e. RTCMediaStream) with a label equal to the
label of a previously added MediaStreamInterface (the function
CanAddLocalMediaStream implements the check.) Eventually, what the app
using react-native-webrtc will see is that only the first local
MediaStream is streamed to its remote peer.

The standard at https://www.w3.org/TR/mediacapture-streams/#mediastream
states that a good practice is to use a UUID conforming to RFC 4122.
This commit uses such a UUID in the implementation of getUserMedia and,
consequently, allows RTCMediaStreams from different getUserMedia calls
to be streamed through 1 RTCPeerConnection.

### DIFF
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -64,7 +64,13 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints callback:(RCTResponse
   
   NSMutableArray *tracks = [NSMutableArray array];
 
-  RTCMediaStream *mediaStream = [self.peerConnectionFactory mediaStreamWithLabel:@"ARDAMS"];
+  // Initialize RTCMediaStream with a unique label in order to allow multiple
+  // RTCMediaStream instances initialized by multiple getUserMedia calls to be
+  // added to 1 RTCPeerConnection instance. As suggested by
+  // https://www.w3.org/TR/mediacapture-streams/#mediastream to be a good
+  // practice, use a UUID (conforming to RFC4122).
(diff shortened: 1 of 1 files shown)


oney:
I am very curious about why they use `ARDAMS` https://github.com/gandg/webrtc-ios/blob/master/ios-example/AppRTCDemo/APPRTCAppDelegate.m#L269

lyubomir:
I believe ARDAMS is just an example the author of AppRTCDemo picked up. Most of names of the source files of the project AppRTCDemo begin win ARD. I'm pretty sure that's an abbreviation of [A]pp[R]TC[D]emo. MS is obviously an abbreviation of [M]edia[S]tream.

Anyway, there's nothing in WebRTC that requires such a MediaStream label/id. In our (as in Jitsi community) experience of multiple years with WebRTC and browsers (e.g. Chrome, Firefox, Temasys), the string ARDAMS has no significance.

zxcpoiu:
great lightening talk :smile:

oney:
@lyubomir What should we use for label id on Android?

lyubomir:
@oney, I'd recomment using a UUID on Android as well in accord with the good practice cited by the standard (that I mentioned above). (A colleague of mine should start working on the Android side this week so I'll ask him to please come up with a solution sooner rather than later.)

PR #88, opened 2016-06-08
Send null icecandidate when gathering complete

### LINKED ISSUE #87: 'onicecandidate' not fired on ICE gathering completion
According to the spec, `onicecandidate` should be fired with `newCandidate == null` when the ICE gathering process is done. Currently, `onicecandidate` is not fired when gathering is completed.

https://w3c.github.io/webrtc-pc/#h-operation
```
When a new ICE candidate is available or when the ICE gathering process is done , the user agent must queue a task to run the following steps:
 (...)
3. If the gathering process is done: Update connection's ICE gathering state to completed and let newCandidate be null.
4. Fire an event named icecandidate with newCandidate at connection.
```

https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onicecandidate
```
If the event's candidate property is null, ICE gathering has finished.
```

maxhawkins:
Closes #87

### DIFF
--- a/RTCIceCandidateEvent.js
+++ b/RTCIceCandidateEvent.js
@@ -5,7 +5,11 @@ class RTCIceCandidateEvent {
   candidate;
   constructor(type, eventInitDict) {
     this.type = type.toString();
-    Object.assign(this, eventInitDict);
+
+    this.candidate = null;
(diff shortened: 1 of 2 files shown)


maxhawkins:
@oney Does this look OK?

PR #90, opened 2016-06-13
specify OpenGL version in AndroidManifest.xml

### LINKED ISSUE #18: No Video appearing on Android (also in demo)
Please see issue in demo app https://github.com/oney/RCTWebRTCDemo/issues/7

Also have issue when using react-native-webrtc core library in 2 other apps I tested.

maxhawkins:
Running without this was causing no video to appear in RTCViews on my Android device.

May fix #18.

### DIFF
--- a/Documentation/AndroidInstallation.md
+++ b/Documentation/AndroidInstallation.md
@@ -6,6 +6,7 @@
 <uses-permission android:name="android.permission.CAMERA" />
 <uses-feature android:name="android.hardware.camera" />
 <uses-feature android:name="android.hardware.camera.autofocus"/>
+<uses-feature android:glEsVersion="0x00020000" android:required="true" />
 
 <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
 <uses-permission android:name="android.permission.RECORD_AUDIO" />


zxcpoiu:
Which device and version you used has the problem?

@reactor81 @JoeLogan1981 do you guys have time to check?

maxhawkins:
Looking again, I think this was unrelated to the problem. You can close
this pull request.

Instead, the issue seems to happen because in using Live Reload to test my
code.

When I compile and push a new binary it works the first time. Whenever I
make a change and reload the element it shows no video and produces no
error.

Perhaps there's some global state that's not getting cleared? I'll keep
looking.

zxcpoiu:
@maxhawkins 

thanks!
So this PR did fix your problem when you are using Live Reload to test code?
If it fixed your problem as long as did not break anything, I think it's worthy to be included

zxcpoiu:
my concern is: will specifying OpenGL Version in Manifest breaks the user who use webrtc audio only on a older device?

Given that react-native supports only API >= 17 , seems it won't hurt anything.

https://developer.android.com/guide/topics/graphics/opengl.html

>OpenGL ES 1.0 and 1.1 - This API specification is supported by Android 1.0 and higher.
OpenGL ES 2.0 - This API specification is supported by Android 2.2 (API level 8) and higher.
OpenGL ES 3.0 - This API specification is supported by Android 4.3 (API level 18) and higher.
OpenGL ES 3.1 - This API specification is supported by Android 5.0 (API level 21) and higher.


https://developer.android.com/guide/topics/graphics/opengl.html#version-check

> There are several versions of OpenGL ES available on Android devices. You can specify the minimum version of the API your application requires in your manifest, but you may also want to take advantage of features in a newer API at the same time. For example, the OpenGL ES 3.0 API is backward-compatible with the 2.0 version of the API, so you may want to write your application so that it uses OpenGL ES 3.0 features, but falls back to the 2.0 API if the 3.0 API is not available.

JoeLogan1981:
Hi Max,

Does this happen with the local video, remote video, or both when using LiveReload?

Joe

maxhawkins:
I thought it fixed my problem but it didn't. With the new xml entry it only works the first time, as before.

@JoeLogan1981 My test app shows a local video stream from the front camera only. It shows the image the first time I load. When I reload the JS the image is replaced with a gray rectangle.

zxcpoiu:
@maxhawkins 

can you reproduce the problem using [demo](https://github.com/oney/RCTWebRTCDemo/)?

p.s. you may have to change package.json to the latest version accordingly

JoeLogan1981:
@maxhawkins does the same thing happen with remote video?

maxhawkins:
Sorry I got pulled off onto another project. I'll have time to test this in a few weeks.

saghul:
Inactive, closing. Please holler is this is still needed. (but test the latest version first, since we added some EGL rendering related fixes)

PR #95, opened 2016-06-29
Android build: don't depend on specific RN version binaries

philikon:


zxcpoiu:
Is there any difference when use 23.0.3?

philikon:
Functionally no, just 23.0.3 is the latest and thus generally more available in CI environments.

philikon:
Although, hmm, I just see that react-native itself uses 23.0.1. So we can stick with that for now.

zxcpoiu:
Yeah, that's why I was asking.
It's a Deja vu that I used to see discussion about issues of version number somewhere.

philikon:
Thanks! :)

PR #97, opened 2016-07-02
Native to JS plumbing for DataChannel on Android

philikon:
I never implemented the Android code that will let us receive DataChannel messages in JS. Whoops. Here it is!


zxcpoiu:
thanks man!

PR #120, opened 2016-08-23
AndroidManifest.xml update - Package name fix

MicronXD:
`react-native link` was failing to update the android project with the proper import class path.
This resolves the issue.


zxcpoiu:
thx

PR #134, opened 2016-09-14
ios: use WebRTC.framework

philikon:
WebRTC nowadays comes with build scripts for building an iOS framework, obviating https://github.com/pristineio/webrtc-build-scripts. The API provided by WebRTC.framework is somewhat more pleasant and I'd expect it to be better maintained going forward.


philikon:
Note: I didn't do the LFS thing for the dynamic library. I can still do that if it's requested.

philikon:
Updated demo here: https://github.com/oney/RCTWebRTCDemo/pull/28

zxcpoiu:
wow, thank you @philikon , looks promising.
would you please share thoughts/experiences to the following items:

**1. build scripts comes from official repo supports ios only for now?**
I didn't see build script in ` webrtc/build/android/` like in ` webrtc/build/ios/build_ios_libs.sh`

**Q:** is it true? 

**2. library size**
I checked both build scripts and target architectures are the same: `"arm" "arm64" "ia32" "x64"`
and noticed that the size of libraries between **master** and **this PR** has significant differences
on master it is: `72.3 MB`, and this PR is: `20.9 MB`

**Q:** is that expected because the differences between static library and framework?

**3. specified revision number when building**
`webrtc/build/ios/build_ios_libs.sh -r $REV`

I noticed that the build script would not checkout out to specific $REV branch automatically but does the folloing in the end of the build process:

```bash
  # Modify the version number.
  INFOPLIST_PATH=${OUTPUT_DIR}/WebRTC.framework/Info.plist
  MAJOR_MINOR=$(plistbuddy -c "Print :CFBundleShortVersionString" \
                ${INFOPLIST_PATH})
  VERSION_NUMBER="${MAJOR_MINOR}.${POINT_VERSION}"
  echo "Substituting revision number: ${VERSION_NUMBER}"
  plistbuddy -c "Set :CFBundleVersion ${VERSION_NUMBER}" ${INFOPLIST_PATH}
  plutil -convert binary1 ${INFOPLIST_PATH}

  # Copy pod file.
  FORMAT_STRING=s/\${FRAMEWORK_VERSION_NUMBER}/${VERSION_NUMBER}/g
  sed -e ${FORMAT_STRING} ${WEBRTC_BASE_DIR}/webrtc/sdk/objc/WebRTC.podspec > \
      ${OUTPUT_DIR}/WebRTC.podspec
```

**Q:** would `$REV` matter while building? or it just a comment in `info.plist` ?
can we specify a string for example: `'branch_53'`?

**4. correct revision number**

I was considering `13855` `Cr-Commit-Position: refs/heads/master@{#13855}` is the revision number of branch_53.

but turns out it should be: `13317` `Cr-Branched-From: 65f47275cbb5ad3aa5146b9e75afa9894467daff-refs/heads/master@{#13317}`

because `branch_53` does branched from 13117, with additional 6 cherry-picks to it. the last fixed is from `13855`

**Q.** is it true? if so, and I won't affect building process, I can take care of it after merged

**5. git lfs**

to be honest, git lfs is annoying. but if I don't use it, sometime I can't upload or push to a remote branch.
but no worries, I will take care of it if needed after merge.

zxcpoiu:
thanks again @philikon and sorry of the long words.

@oney @lyubomir 
any thoughts?

zxcpoiu:
just for memo:

this is command to extract `Cr-Branched-From` instead `Cr-Commit-Position`
`git log -1 | grep -E 'Cr-Branched-From:.*refs/heads/master@{#' | grep -v '>' | egrep -o "[0-9]+}" | tr -d '}'`

philikon:
> 1. build scripts comes from official repo supports ios only for now?
> I didn't see build script in webrtc/build/android/ like in webrtc/build/ios/build_ios_libs.sh

Yeah, I don't know how Google builds WebRTC for Android as a library. Perhaps they don't at all. In either case, it seems that https://github.com/pristineio/webrtc-build-scripts is our best bet for Android libraries for now.

> 2. library size
> I checked both build scripts and target architectures are the same: "arm" "arm64" "ia32" "x64" and noticed that the size of libraries between master and this PR has significant differences on master it is: 72.3 MB, and this PR is: 20.9 MB

Yeah, the static library has debug symbols that are not present in the framework. If we wanted to include debug symbols for the framework, we would have to ship the separate dSYM bundle, but that's huge (~250MB). I can include it, but it seems largely unnecessary.

> 3. specified revision number when building
> webrtc/build/ios/build_ios_libs.sh -r $REV
> Q: would $REV matter while building? or it just a comment in info.plist ?
can we specify a string for example: 'branch_53'?

It doesn't matter at all, it's just a line in `Info.plist`. The version number ends up being `1.0.$REV`.

> 4. correct revision number

Yeah, I'm happy to call it `13317`, though IMHO saying it's WebRTC version 53 revision 13855 is a more correct way because that way you include the 6 cherry-picks. By saying it's version 53, it should hopefully be clear that not all revisions since the branch point are included.

> 5. git lfs
> to be honest, git lfs is annoying.

It's incredibly annoying. I'd love it if we can avoid using it.

philikon:
Ping?

zxcpoiu:
LGTM by the

1. smaller size ( and we can try to eliminate annoying git lfs )
2. officially more maintainable build process.

ping @oney @lyubomir 

@lyubomir has lot more use cases in jitsi community than me, may have some valuable opinions.

zxcpoiu:
about revision, I will update doc to for example: `branch_53 13855 (branched from 13317)`

for reason to avoid:

1. people may misunderstand and build directly targeting `13855` which doesn't equal to `branch_53 (13855)`

2. weird collision in some cases a future release branch doesn't have any cherry picks, like `branch_54 (13855) branched from 13855`

lyubomir:
@zxcpoiu, I agree with all thoughts you've expressed on this PR.

zxcpoiu:
@lyubomir  thanks for review!
ok, I will finalize it this week ( hopefully :astonished: )

zxcpoiu:
I also removed `git-fls` as well

thanks guys

philikon:
Awesome, thank you guys!

zxcpoiu:
@lyubomir thanks for mention that and my bad for not reviews closely.

I'm going to merge commits below as well:

https://github.com/jitsi/react-native-webrtc/commit/706b40d71696fd2ec8390a045c507322d65d7933
https://github.com/jitsi/react-native-webrtc/commit/2f7934be2227a6c52244d6d2f5e9ff1cc2f95809
https://github.com/jitsi/react-native-webrtc/commit/ad26addaeaa23edac227fa5047907a79de4740ce

I think it would be great that if you can commit to master directly if you found any bugs, since you have the write access and lot more use cases than most of folks.

philikon:
@lyubomir sorry about that, it wasn't my intention to break your code. It must've gotten lost in the refactoring to the .framework and I never ended up resolving the TODO. Thanks for fixing it!

minhtc:
why don't use cocoa pod to embed webrtc lib?

saghul:
@minhtc Because we want land cherry-picked patches if need be.

PR #135, opened 2016-09-16
Added a podspec file for iOS projects.

robblovell:
Added podspec file for easy install into react-native projects iOS that use a podfile.
Updated the documentation for iOS.


robblovell:
Last minute changed turned out to be faulty, doing some more tests.

PR #136, opened 2016-09-16
Added podspec file 

robblovell:
Added a podspec file for easy install into react-native projects that use iOS project podfiles.
Updated the documentation for iOS projects.


robblovell:
Note that it assumes that releases are tagged with "release x.y.z", where x.y.z is the package.json version.

This line in the podspec does this:
```
s.source  = { :git => "git@github.com:oney/react-native-webrtc.git", :tag => "release #{s.version}" }
```

robblovell:
note it also reads the package.json file to get the version and description

robblovell:
okay, there might be an issue with the version I was working with locally and the version that exists now.  This podfile works with 0.10.2.  I need to test 0.12.0.

robblovell:
This works with 0.10.2 but not with version 0.11.0 and 0.12.0.

robblovell:
I reopened this after getting my demo project to work with version 0.12.0.

zxcpoiu:
ahh, I totally don't have any ideas on it.

Are there any conflicts with #134? or can it work together with it?

@philikon would you mind take a look?

zxcpoiu:
@robblovell 

would this affect the installation process if xcode project didn't use `pod`?

zxcpoiu:
Hi @robblovell 

We have changed iOS to use `WebRTC.framework` instead `libjingle_peerconnection`
would you please rebase on master?

saghul:
Ping?

PR #147, opened 2016-10-14
Add support for RTCPeerConnection.setConfiguration()

philikon:
See https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/setConfiguration


zxcpoiu:
hello and thanks again @philikon

Issue #158, opened 2016-10-28
Android: Media device stream not released


stwiname:
Trying to dispose of the local media stream doesn't seem to work correctly.
I am calling:

``` js
localStream.getTracks().forEach(t => t.stop());
localStream.release();
```

But signs are pointing to it still being used.
- I can see camera related logging (`org.webrtc.Logging: CameraStatistics: Camera fps: 30`) continuing on my main device (Nexus 5X, Android 7.1.1) and similar behaviour on another device (Samsung Tablet, Android 5.1.1).
- Device continues to feel hot.
- Battery stats show the usage for the app continues to use the camera.

I think https://github.com/oney/react-native-webrtc/issues/101 might be related to this.

zxcpoiu:
@stwiname which react-native-webrtc version you are using?

stwiname:
Version 0.35. Ill have a quick look to see if i can get it working with 0.36 when i get the chance.

zxcpoiu:
Oh, I mean the version of this library

stwiname:
Ah. Sorry.

We are using the latest master branch

zxcpoiu:
thanks, but I don't have any clue currently.
maybe you should close peerConnection as well?

stwiname:
It seems to be happening even if I don't add the stream to the peer connection.

maximovicd:
I got the same problem and it seems to be happening due to the fact that VideoCapturer is never saved so it can be stopped at later time. Can you guys confirm that that's the root cause of the problem?

zxcpoiu:
@maximovicd 
What do you mean by saved VideoCapturer?
And did you found a solution or workaround?

maximovicd:
I mean in WebRTCModule.java, `VideoCapturer` is created and reference to it is not stored. It has `stopCapture` method, that I tried and successfully used to release the camera capture. I think it's a solution for the problem. Let me know if you need more details.

stwiname:
@maximovicd I have managed to reproduce your solution.

`stopCapture` only exists on `VideoCapturerAndroid` so the `VideoCapture` instance needs to be cast back.

@zxcpoiu This is the `VideoCapturer` reference that needs to be kept and have `stopCapture` called on.
https://github.com/oney/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java#L323 

It also seems to fix https://github.com/oney/react-native-webrtc/issues/101 for me.

[...]

PR #172, opened 2016-11-20
Android: Keep reference to VideoCapturers and dispose of them with media stream

stwiname:
Fixes https://github.com/oney/react-native-webrtc/issues/158

And might also fix (I think this might be device specific): https://github.com/oney/react-native-webrtc/issues/116 https://github.com/oney/react-native-webrtc/issues/101

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -56,6 +56,7 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
     private final SparseArray<PeerConnectionObserver> mPeerConnectionObservers;
     public final Map<String, MediaStream> mMediaStreams;
     public final Map<String, MediaStreamTrack> mMediaStreamTracks;
+    private final Map<String, VideoCapturer> mVideoCapturers;
     private final MediaConstraints pcConstraints = new MediaConstraints();
 
     public WebRTCModule(ReactApplicationContext reactContext) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -327,11 +330,14 @@ public void getUserMedia(ReadableMap constraints,
...
                     videoSource = mFactory.createVideoSource(
                             videoCapturer, videoConstraints);
+
+                    trackId = getNextTrackUUID();
+
+                    mVideoCapturers.put(trackId, videoCapturer);
                 }
             }
 
             if (videoSource != null) {
```
zxcpoiu: never mind, I've made a PR to your branch

zxcpoiu:
I have tested it on `RCTWebRTCDemo`.

* compilation passed
* switched back and force passed

but I don't have any further video cases.

zxcpoiu:
@maximovicd

can you test it on this?

zxcpoiu:
another question:

does anyone know:

1. should we do `removeVedioCapturer` in `peerConnectionClose` and `peerConnectionRemoveStream` also?

2. would `String onRemoveStream()` be called automatically after `peerConnectionRemoveStream`?

stwiname:
I'm not sure these are valid use cases but i think:

1. Users might want to keep displaying a local stream after a peer connection has been removed
1. Users might attach a local stream to multiple peer connections

So cleaning up a local stream at the end of a peer connection might not be desired behavior

zxcpoiu:
@stwiname you are right, makes scene. thanks your reply and PR.

I think this PR is ready for merge.

ping @lyubomir 

about comments by @maximovicd from #158 
>  it might be possible to create multiple tracks from one capturer, in which case when all tracks are dropped, we should drop the main capturer.

this doesn't seem to be happen based on current source code. ( we always create `VideoCapturer` before create video track )

but it this behavior changes in the future, we should have this case in mind and might be change to counter-based refrences

maximovicd:
Great job everyone, this seems to fix it for me.

@zxcpoiu Yes, you are right, but I am not sure if you should be able to create another `VideoCapturer` on the same camera. I don't think you will be allowed to do that because one capturer takes hold of one camera, in that case, we should maybe think of just creating one more track from existing capturer (if that's possible too) and do reference counting. This is just me guessing things, but I can try that out when I have more time and let you know.

zxcpoiu:
thanks @maximovicd 

OK, let's get this merged.

PR #192, opened 2017-01-12
bump version

sagivo:
as for change https://github.com/oney/react-native-webrtc/pull/190 we need to bump the version so people will get it via npm update.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-webrtc",
-  "version": "0.54.4",
+  "version": "0.54.5",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/oney/react-native-webrtc.git"


zxcpoiu:
thanks
I'm going to close this because I use automatic script to bump and publish
`0.54.5` is release and npm published via scripts

Kenny-House:
The react-native release 0.40 recommends a major version bump, since I believe the header updates are backwards incompatible.

sagivo:
good point. maybe should be 0.55

zxcpoiu:
hmm...

version naming of this library follows webrtc branch closely or folks would have no ideas what branch is.

how about 1.54.5 ? does that sounds strange?

zxcpoiu:
or maybe `54.5.0`?

Kenny-House:
It is convenient having the minor version follow the WebRTC branch. 

I think having the 1.x.x for RN >= 0.40 would be clear. Users of RN < 0.40 can track the 0.x.x versions.

zxcpoiu:
sounds good.

so the first digit indicating any major breaking changes,
the second one is webrtc minor version
the third is patch or bug fix.

not entirely follows semver but that's do this

zxcpoiu:
1.54.5 released

jhen0409:
@zxcpoiu could you also release `0.54.6` as revert `0.54.4`? I hope we can still keep `^0.54.4` instead of lock `0.54.4` for old RN versions. Thanks!

Kenny-House:
It might be good to maintain a separate branch for the 0.x.x versions, with the current tip right at 0.54.4 (https://github.com/oney/react-native-webrtc/commit/a722cd17b92e068ec9f88d6d49574e7af7ae3d8e).

Until the actual React Native API's change, it shouldn't be too hard to merge updates into master and the 0.x.x branch.

zxcpoiu:
@Kenny-House 

what do you mean by `Until actual React Native API's change`?

zxcpoiu:
@jhen0409

I had created a branch [0.54.6-oldrn](https://github.com/oney/react-native-webrtc/tree/0.54.6-oldrn)
and published on npm as `0.54.6`
now the master is `1.54.6`

Kenny-House:
@zxcpoiu By `Until the actual React Native API's change` I meant that although the headers have changed on ios for >= 0.40, updates to react-native-webrtc should be fairly easy to merge into the 0.54.6-oldrn branch if anyone is in a situation where upgrading to 0.40 isn't feasible.

PR #193, opened 2017-01-12
Allow for null facingMode and null constraint maps in getUserMedia (Android)

Kenny-House:
GUM constraints like the following example from the readme:
```
getUserMedia({
  audio: false,
  video: {
    optional: [{sourceId: selectedCamera.id}]
  }
}
```
Are crashing on Android when attempting to get non-existent `mandatory` key from map-type constraints and attempting to get non-existent `facingMode` key for video constraints.

This PR adds `hasKey()` checks to ensure the keys exist before getting the values.


zxcpoiu:
thanks!

Issue #196, opened 2017-01-19
Audio output is forced to speaker when closing a peer connection


gmahomarf:
I'm working on an app that has conference calling, and I keep running into an issue where, whenever a party leaves the call, all the other parties' audio is forced out through the device speakers, even if they're using headphones.

I checked the code and noticed [this call to resetAudio](https://github.com/oney/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java#L880) and that `resetAudio` calls `audioManager.setSpeakerphoneOn(true)` [here](https://github.com/oney/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java#L901). What's the reasoning behind this behavior (i.e. resetting audio whenever we close a peer connection)?

I'm using a separate library to handle audio outputs, (I even tried calling [setAudioOutput](https://github.com/oney/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java#L905) directly) but if I manually reset the audio to headphones after closing a peer connection, there's a brief moment during which audio is output through the speakers before being redirected back to the headphones. It would be great if I could get rid of that brief period of speaker output.

zxcpoiu:
yes, thanks.
it's a old implementation before `react-native-incall-manager`

#198 fixed it.

gmahomarf:
@zxcpoiu would you mind adding this to 0.54.X versions? I noticed this is only available in 1.54.7 and I'm still not using RN 0.40. Thanks for the quick response!

gmahomarf:
Never mind. I was looking at github tags not NPM versions. Sorry

PR #198, opened 2017-01-20
android: remove unused API methods

saghul:
These are better handled by a module such as
react-native-incall-manager.

Only resetAudio() was used, which was arguably incorrect, since it's up to
the user to decide what audio mode they want for their application.


zxcpoiu:
fix #196

PR #219, opened 2017-02-25
Update README.md: add another demo

thoqbk:
Hi,

I have another demo with good UI, can you consider to add it in README?

- Web app and signaling server: https://rewebrtc.herokuapp.com/
- React native app repository: https://github.com/thoqbk/rewebrtc


zxcpoiu:
thanks, I will add a section

Issue #223, opened 2017-03-07
Non-public API usage


siemya:
hey @oney 

When i submit my app to appstore, it gets denied with the error;

> The app references non-public selectors in yourApp: channelId, initWithURLStrings:, isBinary, reportId, sdp, sendData:, setChannelId:, stringForType:, typeForString:

When i search my project, it looks like webrtc is using them.

siemya:
did anyone succeeded to submit an app to appstore?

saghul:
Yes, we have succeeded: https://itunes.apple.com/us/app/jitsi-meet/id1165103905?mt=8

I wasn't the one submitting the build, but I'm not aware of such problems.

antoniotuzzi:
this is webrtc/chromium bug 6382 https://bugs.chromium.org/p/webrtc/issues/detail?id=6382#c19
has been solved in M58

siemya:
hey @antoniotuzzi,

when i search the project in github, those keywords still appear.

siemya:
@saghul interesting.

does your react-native-webrtc module contain those keywords?

antoniotuzzi:
I'm quite sure that webrtc build after feb 8 2017 are ok

saghul:
We use current master, so M54.


-Saúl

siemya:
@saghul @antoniotuzzi 

the main problem is we are using react version 0.39.2 atm and we are not ready to upgrade.

thats why we are using version 0.54.4 :/

saghul:
You can probably take master, fix up imports and make it work with RN < 0.40, basically undoing what was done to upgrade, it shouldn't be a huge effort.


-Saúl

shaneneuerburg:
We are still running into a similar issue as this. The error back from Apple is:

> Non-public API usage:
> 
> The app references non-public selectors in relianceconnect: addStream:, channelId, initWithURLStrings:, isBinary, sdp, sendData:, setChannelId:, setIsEnabled:, stringForType:

In the comment above (https://github.com/oney/react-native-webrtc/issues/223#issuecomment-286428269), it doesn't look like the WebRTC library changes actually fixed this particular issue but a different one.

harish50:
+1

saghul:
Just submitted a build to Apple last week with no issues. Tell them those symbols are way too generic or to point out the offending framework.

harish50:
[...]

PR #184, opened 2016-12-27
M55

zxcpoiu:
upgrade webrtc to branch 55 stable 14500 (+8 15071)

Issue #183, opened 2016-12-27
default iOS architecture library 


zxcpoiu:
according to discussion in #141 ( thanks to @besarthoxhaj )
current iOS native webrtc framework includes 4 arch:

* arm / armv7
* arm64
* x64 / x86_64 ( seems for iPad simulators )
* x86 / i386 ( seems for iPhone simulators )

currently, if you want to submit your app to AppStore, you may need to strip x86 and x64 manually before submission via script provided in #141 

**Question:**
is it necessary to include x64/x86 archs in the iOS library?
or just have arm/arm64 included?

davidperrenoud:
If `i386` and `x86_64` aren't included in the framework, we need to make sure that we don't call a method from this framework, otherwise it won't compile for the simulator. Weak linking could help but I don't think it's needed.

You can add compiler macros in order to leave out methods:

```objective-c
#if !(TARGET_IPHONE_SIMULATOR)
    RTCDataChannelConfiguration *init = [RTCDataChannelConfiguration new];
#endif
```

zxcpoiu:
@davidperrenoud 

thanks for mentioned that, though I don't have any idea on it. :cry: 
would you please tell me where should I add those macros?

and what is your opinion about exclude `i386` and `x86_64`?

my current thinking is,
we should make this library prodution ready means that we exclude supports for simulators by default in order to reduce size and avoid any bugs when submission to apple store

saghul:
I move for close. The simulator is useful for testing, stripping the unneeded architectures is trivial and already documented.

Issue #175, opened 2016-12-01
Compile iOS WebRTC.framework with support for vp9 codec


bennyguitar:
VP9 is explicitly not compiled into webrtc with iOS by default (though it is for Android). I know CPU usage is a bit higher with VP9 vs VP8 - but mobile bandwidth makes using VP9 a very exciting possibility.

It's also pretty easy to modify `build_ios_libs.sh` in webrtc to make this a possibility.

zxcpoiu:
@bennyguitar

thanks for recommend.
yes, it would be better we sync codecs between two platforms as possible.

1. have you tried runs vp9 on iOS and is everything alright?
2. do you know is there any reason webrtc doesn't compile VP9 on iOS by default?
3. would you might provide how to specify compiling codecs in `build_ios_libs.sh`?

and you reminds me another question, I had open an issue for it. see #177

bennyguitar:
* VP9 runs on iOS just fine. The reason it is not included by default in iOS webrtc builds is because it has a higher CPU usage to decode/encode (it is not a hardware encoder/decoder). VP9 also has majorly increased quality for bandwidth, which makes it compelling for mobile uses despite the high CPU. Mobile connections can be less than stellar, so getting more out of each packet that comes in is a good thing.

As far as `build_ios_libs.sh`, it's super simple to compile it with VP9. In fact, it's a 1 line change. In that file (on the `branch-heads/54` branch of webrtc), you will see:

```shell
# TODO(tkchin): Add ability to pass in a flag to build vp9. Odds are mobile
# clients will not want it though because of the higher CPU usage.
LIBVPX_BUILD_VP9=0
```

Basically just change the last line to this:

`LIBVPX_BUILD_VP9=1`

And then re-run the build script, and the WebRTC.Framework that is output will now have VP9 codec support.

zxcpoiu:
thanks @bennyguitar I will try to include it on `55`

zxcpoiu:
this is landed in latest master

Proxy08:
Hi, I'm using Webrtc 1.67.1  on iphone 6/7  cpu goes up and my phone gets warm up, can you explain me the process  of Compiling iOS WebRTC.framework with support for vp9 codec? please @zxcpoiu

PR #228, opened 2017-03-15
M56

zxcpoiu:
ios/android: upgrade webrtc to branch 56 stable 15101 (+12 16107)

This is based on previous M55 with **rebased** latest master changes.

## MAJOR CHANGES:

* about M55, see #184 
* (android/ios) can supports custom constraints passed in `create offer/answer` by it's `options` param
* (android only) can supports rest of custom configuration options passed into `PeerConnectionInit`, include some hidden options.
* (android only) use sourceId to select video device is NOT necessary but optional, you can just pass `facingMode` and it will select for you.

## NOTE:

* ios VP9 codec included #175
* ios now compiled with arm/arm64 (#141 #183), so iPad and Simulator are not supported.

## TODO

* ios: implements the rest of configuration options like android
* android: should normalize video resolution params, now in `java` side, we still use `minWidth` ...etc, should be normalized according to spec ( the `exact` key word ? )

## Later TODO

* M57 is releasing
* consider cherry picks ios submission bug described in #223 which fixed till M58
* refactor DEMO Project


saghul:
Damm. I have a M57 PR almost ready. I'll see where we can meet in the middle.

zxcpoiu:
@saghul 

does your PR includes android as well?
or you can point me where your branch is and I can try to merge into this PR as a follow up

hmm...next time we should open an issue and create a branch for upgrade PR to compare

saghul:
@zxcpoiu Yep, Android too. I can make builds with the 4 architectures for both Android and iOS. I'll base my PR on yours, I think that's the safest bet now.

I'm in the process of testing Android, once I got it working I'll PR my changes.

On hindsight I should have opened an issue, but I thought the update process was kinda stalled, so I didn't bother. My bad!

zxcpoiu:
@saghul 

awesome!
I am going to implement the rest of peer connection configurations till then :+1: 
modify my code/coding convention/naming directly if anything you think better

I didn't open an issue as well , what I mean is, I can open an issue/empty PR indicating working process to avoid duplicate efforts and provide a one commit point branch :smile: 

so the architecture would be:

* android: armv7/x86
* ios: arm/arm64 with vp9 codec

right?

saghul:
> so the architecture would be:
> 
> android: armv7/x86
> ios: arm/arm64 with vp9 codec
> right?

Iv'e got armv7, arm64, x86 and x64 builds for both :-) I'm recompiling now with VP9 support enabled.

saghul:
@zxcpoiu I merged your PR onto M56 and I'm now going to PR M57 after some fixes which were required.

Issue #141, opened 2016-09-27
Unsupported Architectures


besarthoxhaj:
When I try to submit the bundle to the TestFlight I get this error.

```
[Transporter Error Output]: ERROR ITMS-90087: "Unsupported Architectures. The executable for NativeRTC.app/Frameworks/WebRTC.framework contains unsupported architectures '[x86_64, i386]'."
[13:44:54]: [Transporter Error Output]: ERROR ITMS-90209: "Invalid Segment Alignment. The app binary at 'NativeRTC.app/Frameworks/WebRTC.framework/WebRTC' does not have proper segment alignment. Try rebuilding the app with the latest Xcode version."
[13:44:54]: [Transporter Error Output]: ERROR ITMS-90125: "The binary is invalid. The encryption info in the LC_ENCRYPTION_INFO load command is either missing or invalid, or the binary is already encrypted. This binary does not seem to have been built with Apple's linker."
[13:44:54]: [Transporter Warning Output]: WARNING ITMS-90080: "The executable 'Payload/NativeRTC.app/Frameworks/WebRTC.framework' is not a Position Independent Executable. Please ensure that your build settings are configured to create PIE executables. For more information refer to Technical Q&A QA1788 - Building a Position Independent Executable in the iOS Developer Library."
[13:44:55]: Transporter transfer failed.
[13:44:55]: WARNING ITMS-90080: "The executable 'Payload/NativeRTC.app/Frameworks/WebRTC.framework' is not a Position Independent Executable. Please ensure that your build settings are configured to create PIE executables. For more information refer to Technical Q&A QA1788 - Building a Position Independent Executable in the iOS Developer Library."
```

I'm looking at this [stackoverflow question](http://stackoverflow.com/questions/30547283/submit-to-app-store-issues). Any help much appreciated :)

**Edit 1:**

As far as I can see **x86_64** architecture is for iPad simulators while **i386** is for iPhone simulators, both are 32 and 64bit compatible.

**Edit 2:**

[Here](http://stackoverflow.com/questions/26190693/xcode-6-01-archive-validation-error-invalid-segment-alignment) they talk about _Archive validation error: Invalid Segment Alignment_ and one of the answers refers to:

> In my case I was submitting an app that include a framework I had built that was a universal framework - created using lipo (for use with both device and simulator).
> 
> After replacing the universal compiled framework with device compiled framework, validation was successful.

**Edit 3:**

[...]

Issue #121, opened 2016-08-25
Shared object loading issue on Android x64


MicronXD:
I'm came across issues with react-native failing to load `.so` files on Android x64 when react-native-webrtc is added to a fresh project. 

When the application starts up, it's looking for an **x64** version of `libgnustl_shared.so` which I think is purposefully omitted by the default `/app/build.gradle` in a fresh react project:

```gradle
    defaultConfig {
        ...
        ndk {
            abiFilters "armeabi-v7a", "x86"
        }
    }
```

I experimented with tweaks to `build.gradle` and `gradle.properties`, but couldn't get the app to live through loading all `.so` files.

After reverting and starting from scratch, I upgraded gradle to 2.14 and the build started failing.
gradle began complaining that `libjingle_peerconnection_so.so` was missing from the arm64 directory somewhere in `/node_modules/react-native-webrtc/build/......`

--------

Starting from a fresh project:

```
> react-native init Demo
> cd Demo/
> react-native run-android
```
![](https://dl2.pushbulletusercontent.com/znLf9rGE0uBD8FGIDkkFFXvwvIUPZS8z/Screenshot_2016-08-25-14-15-39.png)

```
> npm install react-native-webrtc --save
> react-native link
-  rnpm-install info Linking react-native-webrtc android dependency
-  rnpm-install info Android module react-native-webrtc has been successfully linked 
-  rnpm-install info Linking react-native-webrtc ios dependency 
-  rnpm-install info iOS module react-native-webrtc has been successfully linked 
> react-native run-android
```
![](https://dl2.pushbulletusercontent.com/3UZrKh280en3gPIlHKt6pH5MNOyPTgjK/Screenshot_2016-08-25-14-16-23.png)

It launches and dies immediately:

```
Shutting down VM
FATAL EXCEPTION: main
Process: com.demo, PID: 7013
java.lang.UnsatisfiedLinkError: dlopen failed: "/data/data/com.demo/lib-main/libgnustl_shared.so" is 32-bit instead of 64-bit
	at java.lang.Runtime.load(Runtime.java:332)
	at java.lang.System.load(System.java:1069)
	at com.facebook.soloader.DirectorySoSource.loadLibraryFrom(DirectorySoSource.java:71)
	at com.facebook.soloader.DirectorySoSource.loadLibrary(DirectorySoSource.java:42)
	at com.facebook.soloader.SoLoader.loadLibraryBySoName(SoLoader.java:299)
[...]

Issue #79, opened 2016-05-25
WebRTC library version 


zxcpoiu:
**Last Update: 2016-09-10 +0**

**This thread is for discussion about upgrade/maintain webrtc version**

The webrtc library of this repository is based on [webrtc-build-script](https://github.com/pristineio/webrtc-build-scripts) opensourced and maintained by [pristine](http://tech.pristine.io/) and it seems pristine is no longer maintain the [auto-build](http://tech.pristine.io/automated-webrtc-building/) but [webrtc-build-script](https://github.com/pristineio/webrtc-build-scripts)
## WebRTC Library git revision build Info

**webRTC git branch mapping revision:**

[wertc git repo](https://chromium.googlesource.com/external/webrtc)
[chrome release history wiki](https://en.wikipedia.org/wiki/Google_Chrome_release_history)

45 -> #9565
46 -> #9745
47 -> #10155
48 -> #10622
49 -> #11252
50 -> #11765
51 -> #12279
52 -> #12798
53 -> #13317 ( Stable )
54 -> #13869 ( Beta )
## Latest pre-build binary from pristine

The version of webrtc we use in this repository currently is likely between chrome 47-rc and 47-stable.  

**ANDROID:**

```
webRTC svn version: 11139
GIT COMMIT: 0c7e9f540b282d60b94081f601a1694054d8646e
maven: http://mvnrepository.com/artifact/io.pristine/libjingle
download: https://repo1.maven.org/maven2/io/pristine/libjingle/
```

**IOS:**

```
webRTC svn version: 11177
GIT COMMIT: 893505d0fb41a840be5e4a44a1250dba83d79bf5
cocoapods: https://cocoapods.org/pods/libjingle_peerconnection/
cocoapos: https://github.com/CocoaPods/Specs/blob/master/Specs/libjingle_peerconnection/
download: https://s3.amazonaws.com/libjingle/11177/Release/0/libWebRTC.tar.bz2
```

zxcpoiu:
**Side Note:**

iOS will generate crash report such as

```
Exception Type:  EXC_RESOURCE
Exception Subtype: WAKEUPS
Exception Message: (Limit 150/sec) Observed 202/sec over 300 secs
Exception Note:  NON-FATAL CONDITION (this is NOT a crash)
```

it will be treated as a crash on iOS8 and be treated as a warning on iOS9

see:
https://bugs.chromium.org/p/webrtc/issues/detail?id=4212

fix at webrtc revision 12186:
[...]

PR #12, opened 2016-01-07
Fixes Android RTCVideoView.onLayout issue

snoronha:
Android was crashing in the simulator (OS versions 5.1 and 6.0) and device (version 5.1.1). This fixes the simulator issue.

PR #229, opened 2017-03-15
M57

saghul:
This PR sits atop the M56 branch (see #228), so all that work is included, thanks @zxcpoiu!

All M57 builds here were made with GN. The build has VP9 support compiled in.

## iOS

Works fine AFAICT.  Architectures: arm, arm64, x86, x86_64.

## Android

Works fine AFAICT.  Architectures: armv7, x86.

I had to disable the Camera2 enumerator because it crashes on received video and doesn't render the local video. I think the EGL renderer needs updating, I'll look into that as a follow up.

The builds includes a cherry-pick from master which fixes a crash on Android when video is received at an odd resolution.

## TODO

* Add arm64 and x86_64 to the Android build (I just have to package those) and update the gradle file
* Consider changing the versioning scheme since revision numbers are not very representative now that git is used. I suggest we update the wiki with the branch cutoff point that was made for the release, plus the hashes for the cherry-picks
* Check other items in #228


zxcpoiu:
here comes our new muscle @saghul 🎊 

All looks good!
To be honest, I have no ideas about video at all, but will learn. Thanks for your rescue :)
Can be merged once you're ready.

I have some statements just for discussion and will wikify it accordingly

## architectures

> Add arm64 and x86_64 to the Android build (I just have to package those) and update the gradle file

I was thinking to exclude unnecessary architectures since RN android doesn't support 64bit libs well (#121) and iOS has submission bug (#141).

for people doing test, usually has to recompile to enable some debugging flags, so they can adds libs at the same time.

apart from above, exclude necessary libs reduce download sizes.

But it might have issues indicated by @davidperrenoud in #183

On the other hand, it feels right as well that to includes all architecture for easier to test on simulator or for different usage (it can be used even by different projects ), then strip it manually based on different needs.

what's your opinion on it?


## revision

> Consider changing the versioning scheme since revision numbers are not very representative now that git is used. I suggest we update the wiki with the branch cutoff point that was made for the release, plus the hashes for the cherry-picks

Yeh, I always trying to find a good way for versioning.

The scheme we have is, for example,
56 stable 15101 (+12 16107) will be:
1.56.0, the first digit is for trying to following semver when major changes.

Do you mean we use: 56.12.0 instead, and if we really encounter some thing like RN breaking change, we bump to 56.13.0?

I totally agree and prefer the later one.

## additional webrtc cherry picks

Could you post the commit hash or position picked from webrtc master?

Then I can do the paper works to make it clearly to know what salts we've added.

It's good for future contributors who might compile/re-compile libs while keeping functionality/archs consistent.

saghul:
Oops, looks like I missed this notification!

> here comes our new muscle @saghul 🎊

Glad to be here!

> I was thinking to exclude unnecessary architectures since RN android doesn't support 64bit libs well

Can do. They are currently not included in the build, I'm happy to leave them as is.

> iOS has submission bug

I haven't seen this one happen to us, and we just launched Jitsi Meet in the store... Maybe @lyubomir knows if we do something special about it.

> Do you mean we use: 56.12.0 instead, and if we really encounter some thing like RN breaking change, we bump to 56.13.0?

Nope, I mean that "56 stable 15101 (+12 16107)" doesn't clearly convey what commit did we branch on (since git is used now) and what cherry picks are included. See my commit message, I included the hashes.

> Could you post the commit hash or position picked from webrtc master?

See the commit message ;-)

zxcpoiu:
@saghul 

okey, about archs, we can discuss it later and see what @lyubomir think.

about revision, hmm..I must missed something, I didn't quite get what you are referring to.
if you are not talking about npm version, then my understanding is you referring to documentation itself ?

if it's document describing the native webrtc lib itself, then yes, add hashes would be much more precise.
but I think the original `56 stable 15101 (+12 16107)` is necessary to keep, as lots of people didn't bother webrtc git repo directly, they just grab binaries from some folks web-hosting repository.

for example, almost every one call `webrtc release branch_56` as version `16107` on the internet, but it seems strange, because it's branch from `15101`.

so that's the reason I use that scheme.
with only hashes, it would be harder to know what branch it is on with a glance.

since we all build from stable release branch, 
is it look to you if I change it to:

```
M56 15101 (+12 16107)
branched from: 613152af11d6e9a4d046af3c48a7be7642dfcc68
additional cherry picks: 0e22a4cfd3790d80ad1ae699891341fe322cb418
```

( I saw your hashes, thanks )

saghul:
@zxcpoiu 

> so that's the reason I use that scheme.
> with only hashes, it would be harder to know what branch it is on with a glance.

Is that so? If we are building M57 the branch would obviously be 57, at a given commit (hence the need for the hash) plus a number of cherry-picks.

> but I think the original 56 stable 15101 (+12 16107) is necessary to keep, as lots of people didn't bother webrtc git repo directly, they just grab binaries from some folks web-hosting repository.

Can you explain what those numbers mean, and more importantly, how I can get them for the builds I'm currently doing? Thanks!

zxcpoiu:
@saghul 

yes, I think it's a historical reason.

below is the last cherry pick from M56:


```
commit 10b65c02893c09ba6a51cd5a3b62f64aa348136e
Author: philipel <philipel@webrtc.org>
Date:   Fri Jan 20 13:21:38 2017 +0100

    Initiate mid-call probing even if estimated bitrate is at max configured bitrate.

    BUG=chromium:681546

    Review-Url: https://codereview.webrtc.org/2634883003
    Cr-Commit-Position: refs/heads/master@{#16107}
    (cherry picked from commit da5e9d04f59ed88d1db4a364ca23ac8466d7ec8b)

    R=stefan@webrtc.org

    Review-Url: https://codereview.webrtc.org/2639423004 .
    Cr-Commit-Position: refs/branch-heads/56@{#12}
    Cr-Branched-From: 613152af11d6e9a4d046af3c48a7be7642dfcc68-refs/heads/master@{#15101}
```

as you can see, in the scheme`M56 15101 (+12 16107)`,
`15101` is `Cr-Branched-From` and `16107` is the last cherry pick which `Cr-Commit-Position` is `16107`
those number, I think, is a shorthand ( their version control software ) to represent a commit hash.

`+12` is located in `Cr-Commit-Position: refs/branch-heads/56@{#12}`

from my experience, some major webrtc lib hosting (maven or cocoapod), such as [pristineio](https://cocoapods.org/pods/libjingle_peerconnection), who use hooks to build continually, so their script use some `sed` to grab the revision targeted to `Cr-Commit-Position`.

some folks build on release branch and use the same `sed` snippet, therefore, folks call M56 release branch `16107`.

you can check #79 for previous discussions.

but no worries, I don't think it's something to do with this PR, just a discussion.
because we are not talking about npm version, so it's just a documentation work.

we can happily describe our version with both, like above, can be:

```
M56 15101 (+12 16107)
branched from: 613152af11d6e9a4d046af3c48a7be7642dfcc68
additional cherry picks: 0e22a4cfd3790d80ad1ae699891341fe322cb418
```

saghul:
@zxcpoiu Sounds good!

I'll be testing the builds today with Jitsi Meet, and report back.

saghul:
Sorry for the merge / revert, fat fingers :-S

zxcpoiu:
wow, the build script looks delicious.

thanks a lot for your decent work!
you can merge it by yourself when done.

saghul:
Ran some tests on both iOS and Android and things seem to be working fine. I'll merge towards the end of the week if I don't run into anything unexpected. @lyubomir and I will be running this build during the week. If others can test it too that would be great!

> wow, the build script looks delicious.

Thanks! I'm glad you like it :-)

> you can merge it by yourself when done.

Merge it to master or to the M57 branch?

zxcpoiu:
This PR merged to M57.
 It would be fine to merge into master directly since it's tested.

Then I would release a version `1.57.0` and do some documentation work.

saghul:
Landed it in master. Let's have some more people test it before making a release.

PR #235, opened 2017-03-22
Add ability to switch cameras on the fly

saghul:
This patch adds a private custom API for switching the cameras on a
video track on the fly, without the need for add / removing tracks or
renegotiating.

It's MediaStreamTrack._switchCamera() and it only works for local video
streams, it will throw otherwise.

This is how the reference application AppRTCMobile implements camera
switching.

### DIFF
--- a/MediaStreamTrack.js
+++ b/MediaStreamTrack.js
@@ -83,6 +83,23 @@ export default class MediaStreamTrack extends EventTarget(MEDIA_STREAM_TRACK_EVE
     this.muted = !this._enabled;
   }
 
+  /**
+   * Private / custom API for switching the cameras on the fly, without the
+   * need for adding / removing tracks or doing any SDP renegotiation.
+   *
+   * This is how the reference application (AppRTCMobile) implements camera
+   * switching.
+   */
+  _switchCamera() {
+    if (this.remote) {
+      throw new Error('Not implemented for remote tracks');
+    }
+    if (this.kind !== 'video') {
+      throw new Error('Only implemented for video tracks');
+    }
+    WebRTCModule.mediaStreamTrackSwitchCamera(this.id);
+  }
+
   applyConstraints() {
     throw new Error('Not implemented.');
   }
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -522,6 +522,19 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
+    public void mediaStreamTrackSwitchCamera(final String id) {
+        MediaStreamTrack track = mMediaStreamTracks.get(id);
+        if (track != null) {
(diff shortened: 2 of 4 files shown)


MediaStreamTrack.js
```diff
@@ -83,6 +83,16 @@ export default class MediaStreamTrack extends EventTarget(MEDIA_STREAM_TRACK_EVE
     this.muted = !this._enabled;
   }
 
+  _switchCamera() {
```
lyubomir: I do not claim that it's always been the case of the contributions of the Jitsi community but I myself made conscious efforts to at least try to have documentation comments in what I propose for merging into oney/react-native-webrtc. So please take it with many grains of salt. I believe in this case a jsdoc will be very useful because you're introducing non-standard JS WebRTC API. Something like why we decided to go about this way. You pretty much have it in the PR description, I just feel it'd be useful in the source code itself as well.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -522,6 +522,19 @@ public void mediaStreamTrackSetEnabled(final String id, final boolean enabled) {
     }
 
     @ReactMethod
+    public void mediaStreamTrackSwitchCamera(final String id) {
+        MediaStreamTrack track = mMediaStreamTracks.get(id);
+        if (track != null && track.kind().equals("video")) {
```
lyubomir: This one applies to the implementations of `mediaStreamTrackSwitchCamera` on both Android and iOS. The check whether track.kind is video has already been taken care of on the JS side and doesn't need to be repeated in the native side. If there's a problem in the source code which causes the check to pass through the JS side but then fail at the native side, I think it's better for us as developers to have it fail on the native side more glamorously than silently because the problem will be in the implementation. If we just silently allow the method call to do nothing, I think it will be a tad difficult to capture the problem at development time. So I'd personally remove the `kind` check on the native side.

MediaStreamTrack.js
saghul: You're absolutely right.  I'll also add a note to the readme, since it's hidden in the source otherwise.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
saghul: Yeah, I was trying to be overly cautious, i'll simplify.

saghul:
Thanks for the review @lyubomir!

saghul:
@lyubomir I believe I addressed your feedback, can you PTAL?

shaneneuerburg:
What is the reasoning behind having this be a private API? Seems like a common use-case that would be good to publicly expose.

HasnainAD:
I am facing an issue. When I call  `MediaStreamTrack.prototype._switchCamera()` It shows 

> Only implemented for video tracks

What am I missing here?

Nguyen-Thanh-Tung:
localStream.getVideoTracks().forEach((track) => {
      track._switchCamera()
    })

nadelicious:
@saghul Is this also working in mobile browser?

PR #244, opened 2017-04-04
ios: add mirroring capability to RTCView

saghul:


saghul:
This makes RTCView (on iOS) on feature par with Android. @lyubomir PTAL.

PR #247, opened 2017-04-06
ios: fix use of Apple non-public API

saghul:
This the same M57 build we have, plus this cherry-pick:
https://bugs.chromium.org/p/webrtc/issues/detail?id=6382


zxcpoiu:
once you merged, I will update README
https://chromium.googlesource.com/external/webrtc.git/+/1634e160426df926e14cf9f1e5346d2a1dc9c909

PR #249, opened 2017-04-07
M58

saghul:
The M58 branch was cut: https://groups.google.com/forum/#!topic/discuss-webrtc/13BE3fbHcLU

~This PR adds support for it. (iOS only for now).~

~There seems to be a build issue, so if you try to make your own build using the provided script you may run into https://groups.google.com/forum/#!topic/discuss-webrtc/rB-ntaaGXDA~

### TODO

- ~iOS: test RTCMTLVideoView, which is based on Metal~ (not all necessary commits are in M58)
- [x] Android: make an actual build


saghul:
Both iOS and Android builds are up for testing now.

saghul:
I switched our SurfaceViewRenderer to the builtin one, which got refactored and now handles many things we used to do manually. That means we get to delete ~1000 likes of code!

I'm still testing everything works as before, but so far so good!

android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
```diff
@@ -418,21 +187,12 @@ public void setObjectFit(String objectFit) {
...
-
-        synchronized (layoutSyncRoot) {
-            if (this.scalingType == scalingType) {
-                return;
-            }
-
+        if (this.scalingType != scalingType) {
             this.scalingType = scalingType;
 
             surfaceViewRenderer = getSurfaceViewRenderer();
```
lyubomir: Assigning to `surfaceViewRenderer` causes a compile-time error because the field is final. You didn't really want to assign to the field but it merely happened as a side effect of deleting `SurfaceViewRenderer surfaceViewRenderer;` at the beginning of the function. That was easy to fix because a local variable wasn't really anyway and it was OK to simply write `getSurfaceViewRenderer().setScalingType(scalingType)`. Anyway, I hit another compile-type error so I'm not amending your commit right away.

android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
```diff
@@ -237,106 +130,6 @@ protected void onDetachedFromWindow() {
...
-            // surfaceViewRenderer's render Thread.
-            post(requestSurfaceViewRendererLayoutRunnable);
-        }
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    protected void onLayout(boolean changed, int l, int t, int r, int b) {
```
lyubomir: @saghul, that's the second compile-time error I'm talking about in my previous comment. `WebRTCView` extends `ViewGroup` and `onLayout` is the implementation of an abstract method. I'm going to leave it for now so that you have a chance to look into it and I can switch back to my other tasks. If push comes to shove in the meantime, I may come back and look into it myself.

lyubomir:
@saghul, I merged https://github.com/oney/react-native-webrtc/pull/249/commits/86ed1abb963114269f523778b9be34b0211a372c (i.e. the first four commits of this PR) into jitsi/react-native-webrtc's master so that we can start testing 58 at least internally and see how it works. Technically, I did a rebase of this PR on top of oney/react-native-webrtc's master today so that hashes aren't the same but I didn't modify the commit's patches in any way.

saghul:
@lyubomir Thanks for taking a look! I'm not sure what happened there, I did test the code but I now wonder if somehow managed to make some of the changes after :-S

saghul:
@lyubomir I'm not sure what the hell I did with the last commit, `onLayout` was also missing 😮 At any rate, I tested it now and fixed those issues, PTAL.

lyubomir:
On the subject of merging 58 into master, should we be mindful of the crashes that 58 introduced on both desktop and mobile?

saghul:
@lyubomir The patch for those is already available apparently so I'm making new builds including that one. For reference: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/discuss-webrtc/pfwS9eeS23c

saghul:
Updated the iOS and Android builds including the fix for issue 679306, currently under embargo.

android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
```diff
@@ -237,102 +116,10 @@ protected void onDetachedFromWindow() {
...
-
-                    l = (width - frameDisplaySize.x) / 2;
-                    t = (height - frameDisplaySize.y) / 2;
-                    r = l + frameDisplaySize.x;
-                    b = t + frameDisplaySize.y;
-                }
-                break;
-            }
-        }
         surfaceViewRenderer.layout(l, t, r, b);
```
lyubomir: The line seems wrong because these `l`, `t`, `r`, `b` are relative to the parent. So this `WebRTCView` is being told to lay itself out in `l`, `t`, `r`, `b` and we want `surfaceViewRenderer` to fill this `WebRTCView` so the arguments provided to `sufraceViewRenderer.layout` should be like in the deleted `SCALE_ASPECT_FILL` case.

lyubomir:
I've merged the first 3 commits in jitsi/react-native-webrtc which bring WebRTC 58 w/o the second 3 which remove source code on Android so that we can prepare internal builds for testing and gather more experience with how the latest WebRTC 58 builds perform.

saghul:
@lyubomir Sounds good. Let's battle test 58 alone, I'll then focus on getting the Android renderer back in shape.

notedit:
have not test m58 yet.  and then  m59 comes.

saghul:
I removed the last 3 commits. I'll make a follow up PR after this one lands (after addressing your feedback).

lyubomir:
Both Android and iOS are crashing just like desktop. From my perspective, it's not worth it for jitsi/jitsi-meet to continue with WebRTC 58 until either jitsi/jitsi-videobridge manages to completely prevent the crashes or, most likely, Google fixes the causes of the crashes. People utilizing react-native-webrtc for audio and video (without video simulcast and video stream rewriting) may want to test this PR if it gives them a kick forward such as not experiencing crashes caused by videos with odd resolutions and have it merged in master if it is stable for them.

saghul:
@lyubomir Updated the commit logs with the exact commit info that I made the builds from.

So, I given we are not very happy with the quality in M58, so I'd propose the following:

- land this PR on master
- DON'T make a release
- those who want to test master are free to do so, in the meantime I'll make M59 builds and follow the bugtracker, looks like all crashes are gone in Canary (now 60), so maybe they'll backport some of the fixes

This way it's easier to iterate on WebRTC builds since the deltas will be smaller.

zxcpoiu:
@saghul 

how about release a 1.57.1 and publish to npm on current master,
then land this to master and pre-release a 1.58.0 without publish to npm?

saghul:
@zxcpoiu That sounds good to me. Can you make the 1.57.1 release? I'll discuss with @lyubomir merging this afterwards.

PR #256, opened 2017-04-16
Added a check on the sdpMid's default value in RTCPeerConnection.js

lightning-zgc:
If sdpMid doesn't exist，NoSuchKeyException will affect connection.

### DIFF
--- a/RTCPeerConnection.js
+++ b/RTCPeerConnection.js
@@ -181,6 +181,9 @@ export default class RTCPeerConnection extends EventTarget(PEER_CONNECTION_EVENT
   }
 
   addIceCandidate(candidate, success, failure) { // TODO: success, failure
+    if (!candidate.sdpMid) {
+      candidate.sdpMid = '';
+    }
     WebRTCModule.peerConnectionAddICECandidate(candidate.toJSON(), this._peerConnectionId, (successful) => {
       if (successful) {
         success && success();
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -1131,14 +1131,8 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
         PeerConnection peerConnection = getPeerConnection(id);
         Log.d(TAG, "peerConnectionAddICECandidate() start");
         if (peerConnection != null) {
-            String sdpMid;
-            try {
-                sdpMid = candidateMap.getString("sdpMid");
-            } catch (NoSuchKeyException e) {
-                sdpMid = "";
-            }
             IceCandidate candidate = new IceCandidate(
-                    sdpMid,
+                candidateMap.getString("sdpMid"),
                 candidateMap.getInt("sdpMLineIndex"),
                 candidateMap.getString("candidate")
             );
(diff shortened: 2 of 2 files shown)


zxcpoiu:
Hi @lightning-zgc 
thanks for your PR.

I think if these params are mandatory, then instead modify it in `native` side, this sort of things like `default key/value` or `pre-check stuff`, would be easier to maintain if we do it in `js` side. like `RTCConfiguration`.

you might put these check in https://github.com/oney/react-native-webrtc/blob/master/RTCPeerConnection.js#L183

something like we do in https://github.com/oney/react-native-webrtc/blob/master/RTCPeerConnection.js#L119

then we solid ios as well.

thanks!

lightning-zgc:
Thank you for your suggestion.The change has been moved to JS code side.

saghul:
I don't understand why the plugin needs to care about this. If the user if passing a broken ICE candidate it's the user's problem,not this library's IMHO.

zxcpoiu:
I'm not sure what it would happen if sdpmid is empty.

it seems doc says:

* if candidate is empty, indicate the end of remote candidates.
* both `sdpMid` and `sdpMLineIndex` is nullable (https://www.w3.org/TR/webrtc/#dom-rtcicecandidate-sdpmid)
* If candidate is not null and both sdpMid and sdpMLineIndex are null, return a promise rejected with a newly created TypeError. (https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-addicecandidate())


https://www.w3.org/TR/webrtc/#dom-rtcicecandidate-sdpmid

> @ sdpMid of type DOMString, readonly , nullable
>
>If not null, this contains the identifier of the "media stream identification" as defined in [RFC5888] for the media component this candidate is associated with.
>
> @sdpMLineIndex of type unsigned short, readonly , nullable
>
>If not null, this indicates the index (starting at zero) of the media description in the SDP this candidate is associated with.


https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-addicecandidate()

> If candidate.sdpMid is not null, run the following steps:
> 
> If candidate.sdpMid is not equal to the mid of any media description in remoteDescription , reject p with a newly created OperationError and abort these steps.
> 
> Else, if candidate.sdpMLineIndex is not null, run the following steps:
> 
> If candidate.sdpMLineIndex is equal to or larger than the number of media descriptions in remoteDescription , reject p with a newly created OperationError and abort these steps.
> 


I'm not exactly sure, but educated guessing that:

* one of `sdpMid` and `sdpMLineIndex` should present. the other one can be nullable.
* if `sdpMid` is not null, use it first.
* otherwise, use `sdpMLineIndex`

correct me if I'm wrong :pray: 

for my personal preference, in most situations, I would prefer warning than crash. because the user on the app store comments cruelly :scream:

zxcpoiu:
@lightning-zgc

if the review discussion finally passed. please rebase/squash your commit into one. or the old commit will still shows in the code base

saghul:
@zxcpoiu I think this doesn't fix anything. The user can pass whatever they want: a number, an object with whatever properties, undefined / null, etc. If we want to go the full nine yards and check the object has the required properties that's one thing, but the current fix is a corner case, probably specific to the user's code.

I would not merge this, but I leave it up to you.

zxcpoiu:
@saghul 

yes, you are right.
this probably specific to user's code.

@lightning-zgc 
I noticed this probably wont's happen because our `RTCIceCandidate` in `RTCIceCandidate.js` always has these keys.

you might need to use the following way to invoke `addIceCandidate`

`pc.addIceCandidate(new RTCIceCandidate(data.candidate));`
and thus, the key should present.

I'm going to close this

zxcpoiu:
@lightning-zgc thanks anyway for letting me notice the doc.

for future alignment with spec noted above, PR is still welcome :)

thanks for double check and review @saghul

lightning-zgc:
Thank you for your enthusiasm and concern, I will pay more attention to the need to submit.This problem is really due to my inadequate view, I found the problem in the use of the [peerjs](https://github.com/peers/peerjs) framework calls this framework under React Native. Now, I think it's peerjs's issue.Thank you again for your attention and serious.

PR #258, opened 2017-04-19
ios: set default video constraints to 720p

saghul:
By default the front camera  will be opened in 480p otherwise. This
aligns with the behavior in Android after
https://github.com/oney/react-native-webrtc/commit/f89f43c1799c2b65b99aa5d7a0114e0acf8eaf8d


saghul:
@lyubomir PTAL.

PR #261, opened 2017-04-23
ios: generate reactTag for remote streams/tracks

paweldomas:
Will generate react tags for remote streams and tracks in the form of
{RTCPeerConnection react tag}_{stream id|track id}

This allows to receive streams/tracks with the same IDs through different PeerConnections.

Required for "peer to peer".


saghul:
Tested and merged since it works fine.. We might need to revisit how we keep track of both local and remote streams as a whole, but that's on the edge of the scope of this PR. Thanks @paweldomas and sorry for the delay!

Issue #267, opened 2017-05-03
Support for max frame rate?


thehappycoder:
In our app it's critical to be able to specify maximum frame rate (e.g. 12). It seems react-native-webrtc does not support it at the moment.

Is there some workaround or unofficial branch / fork where this is already implemented? We need this in a week or two.

wumke:
FYI: I've seen articles of people who modify lines of the sdp manually after it has been generated to reduce the bandwidth (see https://webrtchacks.com/limit-webrtc-bandwidth-sdp/)... don't know if it can be fixed this easy though. I noticed that my iphone video stream is displayed very slowly and with lots of delay on my android device, so if limiting the frame rate/dimensions works by modifying te sdp please let me know, then it might also be a solution for me!

thehappycoder:
@wumke The other day I found out that on master branch kRTCMediaConstraintsMinFrameRate is specified https://github.com/oney/react-native-webrtc/blob/master/ios/RCTWebRTC/WebRTCModule%2BRTCMediaStream.m#L45

So I tried specifying kRTCMediaConstraintsMaxFrameRate. It worked. But would be nice to do it via Javascript.

wumke:
@thehappycoder 
Thank you for the hint, you saved me hours of searching and experimenting I had planned for today, I've set the constraints to:
```
NSDictionary *mandatoryConstraints
      = @{ kRTCMediaConstraintsMinWidth     : @"640",
           kRTCMediaConstraintsMinHeight    : @"480",
           kRTCMediaConstraintsMinFrameRate : @"15",
           kRTCMediaConstraintsMaxWidth     : @"1280",
           kRTCMediaConstraintsMaxHeight    : @"720",
           kRTCMediaConstraintsMaxFrameRate : @"30" };
```
Now it works great!

@oney I suggest setting up max values in the repository as long as it is not supported to set them via javascript... imho nobody needs full-hd 60fps video for a video call.

zxcpoiu:
it may needs to align as close as possible with https://www.w3.org/TR/mediacapture-streams/#constrainable-interface if we implement it in javascript api.

for the default value, I'm not sure which default value is ideal for the community.
you better call @saghul

saghul:
> it may needs to align as close as possible with https://www.w3.org/TR/mediacapture-streams/#constrainable-interface if we implement it in javascript api.

Indeed.

> for the default value, I'm not sure which default value is ideal for the community.

[...]

PR #288, opened 2017-06-06
Allow constraints to be passed.

georgewwindsor:
specifically integrated maximum and minimum constraints for  frame-rate, width, and height.

MINIMUM resolution is 240x160 on my Iphone 5

Usage in JS code:
constraints = {video:
            {
              mandatory: {
                maxHeight: 160, // Provide your own width, height and frame rate here
                maxWidth: 240,
                maxFrameRate: 10
              },
              facingMode: CM.facingMode // 'environment'
            }
        ,audio: true};
getUserMedia(constraints, function(stream){
//low resolution stream returned.
});

### DIFF
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -292,7 +292,7 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
       }
     }
     NSArray * mandatoryKeys = [mandatory allKeys];
-    if([mandatoryKeys count] == 0){
+    if ([mandatoryKeys count] == 0) {
       mandatory = [self defaultMediaStreamConstraints];
     }
     RTCMediaConstraints* customConstraints = [[RTCMediaConstraints alloc] initWithMandatoryConstraints:mandatory optionalConstraints:nil];
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -269,8 +269,8 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
 
   if (videoDevice) {
 
-     //Actually Filter the passed constraints, creating a new dictionary object to pass them as the correct format.
-     //Minimum Resolution 240 * 160
+
+      //Actually Filter the passed constraints, creating a new dictionary object to pass them as the correct format
 
       NSMutableDictionary *mandatory =[[NSMutableDictionary alloc]init];
       NSDictionary *video=constraints[@"video"];
@@ -287,7 +287,7 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
 
              if ([[frameRateObj valueForKey:@"max"] isKindOfClass:[NSNumber class]]) {
                  NSLog(@"framerate is a number");
-                 frameRateMax = [NSMutableString stringWithFormat:@"%d", (NSInteger)frameRateObj[@"max"]];
+                 frameRateMax = [NSMutableString stringWithFormat:@"%@", (NSInteger)frameRateObj[@"max"]];
 
 
              }
@@ -313,7 +313,7 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
 
             NSMutableString *heightMax = nil;
             if ([[heightObj valueForKey:@"max"] isKindOfClass:[NSNumber class]]) {
-                heightMax = [NSMutableString stringWithFormat:@"%d", (NSInteger)heightObj[@"max"]];
+                heightMax = [NSMutableString stringWithFormat:@"%@", (NSInteger)heightObj[@"max"]];
                 NSLog(@"height is number");
 
             }else{
@@ -334,8 +334,8 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
           NSMutableString *widthMax=nil;
 
           if(widthObj != nil){
-              if ([[heightObj valueForKey:@"max"] isKindOfClass:[NSNumber class]]) {
-                  widthMax = [NSMutableString stringWithFormat:@"%d", (NSInteger)widthObj[@"max"]];
+              if ([[widthObj valueForKey:@"max"] isKindOfClass:[NSNumber class]]) {
+                  widthMax = [NSMutableString stringWithFormat:@"%@", (NSInteger)widthObj[@"max"]];
 
                   NSLog(@"width is number");
               }else{
@@ -367,7 +367,6 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
 
 
       }
-
 
 
       RTCMediaConstraints* betterConstraints =
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -260,7 +260,6 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
       videoDevice = [AVCaptureDevice defaultDeviceWithMediaType:AVMediaTypeVideo];
     }
   }
-
   if (videoDevice) {
     NSMutableDictionary *mandatory = [[NSMutableDictionary alloc]init];
     if (videoConstraints != nil) {
@@ -293,7 +292,7 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
       }
     }
     NSArray * mandatoryKeys = [mandatory allKeys];
-    if([mandatoryKeys count]==0){
+    if([mandatoryKeys count] == 0){
       mandatory = [self defaultMediaStreamConstraints];
     }
     RTCMediaConstraints* customConstraints = [[RTCMediaConstraints alloc] initWithMandatoryConstraints:mandatory optionalConstraints:nil];
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -270,7 +270,7 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
   if (videoDevice) {
      NSMutableDictionary *mandatory =[[NSMutableDictionary alloc]init];
        NSDictionary *video=constraints[@"video"];
-       RTCMediaConstraints* betterConstraints = nil;
+       RTCMediaConstraints* customConstraints = nil;
        if (video != nil){
          NSDictionary *mandatoryParameters = video[@"mandatory"];
          if(mandatoryParameters != nil){
@@ -340,14 +340,14 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
                [mandatory setObject:minWidth forKey:@"minWidth"];
              }
            }
-           betterConstraints = [[RTCMediaConstraints alloc] initWithMandatoryConstraints:mandatory optionalConstraints:nil];
+           customConstraints = [[RTCMediaConstraints alloc] initWithMandatoryConstraints:mandatory optionalConstraints:nil];
          }else{
-         betterConstraints = [self defaultMediaStreamConstraints];
+         customConstraints = [self defaultMediaStreamConstraints];
          }
        }else{
-       betterConstraints = [self defaultMediaStreamConstraints];
+       customConstraints = [self defaultMediaStreamConstraints];
        }
-    RTCAVFoundationVideoSource *videoSource = [self.peerConnectionFactory avFoundationVideoSourceWithConstraints:betterConstraints];
+    RTCAVFoundationVideoSource *videoSource = [self.peerConnectionFactory avFoundationVideoSourceWithConstraints:customConstraints];
 
     switch (videoDevice.position) {
     case AVCaptureDevicePositionBack:
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -271,23 +271,23 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
           [mandatory setObject:maxFrameRate forKey:@"maxFrameRate"];
         }
         if ([mandatoryParameters valueForKey:@"maxHeight"] != nil && [[mandatoryParameters valueForKey:@"maxHeight"] isKindOfClass:[NSNumber class]]) {
-          NSMutableString *maxHeight = [NSMutableString stringWithFormat:@"%@", mandatoryParameters[@"maxHeight"]];
+          NSMutableString *maxHeight = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"maxHeight"]];
           [mandatory setObject:maxHeight forKey:@"maxHeight"];
         }
         if ([mandatoryParameters valueForKey:@"maxWidth"] != nil && [[mandatoryParameters valueForKey:@"maxWidth"] isKindOfClass:[NSNumber class]]) {
-          NSMutableString *maxWidth = [NSMutableString stringWithFormat:@"%@", mandatoryParameters[@"maxWidth"]];
+          NSMutableString *maxWidth = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"maxWidth"]];
           [mandatory setObject:maxWidth forKey:@"maxWidth"];
         }
         if ([mandatoryParameters valueForKey:@"minFrameRate"] != nil && [[mandatoryParameters valueForKey:@"minFrameRate"] isKindOfClass:[NSNumber class]]) {
-          NSMutableString *minFrameRate = [NSMutableString stringWithFormat:@"%@", mandatoryParameters[@"minFrameRate"]];
+          NSMutableString *minFrameRate = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"minFrameRate"]];
           [mandatory setObject:minFrameRate forKey:@"minFrameRate"];
         }
         if ([mandatoryParameters valueForKey:@"minHeight"] != nil && [[mandatoryParameters valueForKey:@"minHeight"] isKindOfClass:[NSNumber class]]) {
-          NSMutableString *minHeight = [NSMutableString stringWithFormat:@"%@", mandatoryParameters[@"minHeight"]];
+          NSMutableString *minHeight = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"minHeight"]];
           [mandatory setObject:minHeight forKey:@"minHeight"];
         }
         if ([mandatoryParameters valueForKey:@"minWidth"] != nil && [[mandatoryParameters valueForKey:@"minWidth"] isKindOfClass:[NSNumber class]]) {
-          NSMutableString *minWidth = [NSMutableString stringWithFormat:@"%@", mandatoryParameters[@"minWidth"]];
+          NSMutableString *minWidth = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"minWidth"]];
           [mandatory setObject:minWidth forKey:@"minWidth"];
         }
       }
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -268,98 +268,88 @@ RCT_EXPORT_METHOD(getUserMedia:(NSDictionary *)constraints
   }
 
   if (videoDevice) {
-     //Actually Filter the passed constraints, creating a new dictionary object to pass them as the correct format
-         NSMutableDictionary *mandatory =[[NSMutableDictionary alloc]init];
-         NSDictionary *video=constraints[@"video"];
-         RTCMediaConstraints* betterConstraints = nil;
-         //If the video constraints exists, continue to parse its properties.  These properties match the format that is currently passed to chrome.
-
-         if(video != nil){
-           NSDictionary *mandatoryParameters = video[@"mandatory"];
-             if(mandatoryParameters != nil){
-                 if([mandatoryParameters valueForKey:@"maxFrameRate"] != nil) {
-                     // The key existed...
-                     NSMutableString *frameRateMax = nil;
-                     if ([[mandatoryParameters valueForKey:@"maxFrameRate"] isKindOfClass:[NSNumber class]]) {
-                         frameRateMax = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"maxFrameRate"]];
-                     }else{
-                         frameRateMax =mandatoryParameters[@"maxFrameRate"];
-                     }
-                     if(frameRateMax != nil){
-                         [mandatory setObject:frameRateMax forKey:@"maxFrameRate"];
-                     }
-                 }
-                 if([mandatoryParameters valueForKey:@"maxHeight"] != nil) {
-                     // The key existed...
-                     NSMutableString *heightMax = nil;
-                     if ([[mandatoryParameters valueForKey:@"maxHeight"] isKindOfClass:[NSNumber class]]) {
-                         heightMax = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"maxHeight"]];
-                     }else{
-                         heightMax =mandatoryParameters[@"max"];
-                     }
-                     if(heightMax != nil){
-                         [mandatory setObject:heightMax forKey:@"maxHeight"];
-                     }
-                 }
-                 if([mandatoryParameters valueForKey:@"maxWidth"] != nil) {
-                     // The key existed...
-                     NSMutableString *widthMax = nil;
-                     if ([[mandatoryParameters valueForKey:@"maxWidth"] isKindOfClass:[NSNumber class]]) {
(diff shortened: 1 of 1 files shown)


m59peacemaker:
I'll try to test this today.

zxcpoiu:
anyone who are interested in video frame constraints, please take a look #267

georgewwindsor:
No problem, ill get those taken care of and get back to you. definately needs some cleanup. Also im not convinced framerate works. as far as android, is it simple a missing feature for max or it just hasnt been mapped correctly?

Max size is far more valuable to my team, because it reduces bandwidth, but i agree that both max and min being implemented would be ideal. ill take a look at passing some IOS min values.

georgewwindsor:
@zxcpoiu  I have cleaned up the code formatting, added 'minimum' constraint parsing, and changed the constraints to match the format that is specified in the documentation.

Please review and get back to me with any further changes you require to merge this in.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,13 +268,98 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
-    // starters because it is kind of a common and hence important feature on
-    // a mobile device.
-    switch (videoDevice.position) {
+     //Actually Filter the passed constraints, creating a new dictionary object to pass them as the correct format
+         NSMutableDictionary *mandatory =[[NSMutableDictionary alloc]init];
+         NSDictionary *video=constraints[@"video"];
+         RTCMediaConstraints* betterConstraints = nil;
+         //If the video constraints exists, continue to parse its properties.  These properties match the format that is currently passed to chrome.
+
+         if(video != nil){
```
saghul: style: here and elsewhere: spacing: `if (condition) {`

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,13 +268,98 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+           NSDictionary *mandatoryParameters = video[@"mandatory"];
+             if(mandatoryParameters != nil){
+                 if([mandatoryParameters valueForKey:@"maxFrameRate"] != nil) {
+                     // The key existed...
```
saghul: pointless comment

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,13 +268,98 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+                     NSMutableString *frameRateMax = nil;
+                     if ([[mandatoryParameters valueForKey:@"maxFrameRate"] isKindOfClass:[NSNumber class]]) {
+                         frameRateMax = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"maxFrameRate"]];
+                     }else{
+                         frameRateMax =mandatoryParameters[@"maxFrameRate"];
```
saghul: can't we just call this variable "maxFrameRate" ?

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,13 +268,98 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+                     if(widthMin != nil){
+                         [mandatory setObject:widthMin forKey:@"minWidth"];
+                     }
+                 }
+                 betterConstraints = [[RTCMediaConstraints alloc] initWithMandatoryConstraints:mandatory optionalConstraints:nil];
+             }else{
+               betterConstraints = [self defaultMediaStreamConstraints];
+             }
+         }else{
+         betterConstraints = [self defaultMediaStreamConstraints];
```
saghul: indentation is wrong here. Also, you can probably remove those else clauses by initializing constraints to the default and then overriding them if you find better ones.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,13 +268,98 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+         }
+         RTCAVFoundationVideoSource *videoSource = [self.peerConnectionFactory avFoundationVideoSourceWithConstraints:betterConstraints];
+
+      switch (videoDevice.position) {
```
saghul: broken indentation

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
georgewwindsor: I dont think we should do this because the current constraints are just minimum, and not maximum, so you would have to nullify the opposite direction to make sure you weren't scoping your range too small, in some cases.  Ie you would have constraints that you didn't realize you had.

georgewwindsor:
@zxcpoiu  @saghul  I have made the updates you requested, please review again and let me know if there any other changes needed.

georgewwindsor:
@zxcpoiu 
@saghul 

can you guys take a look at this? I have updated with the changes you requested...

saghul:
It looks pretty much the same as before to me. Many indentation and spacing issues. Also the 2 else clauses which can be simplified are still there.

georgewwindsor:
https://github.com/smiletime/react-native-webrtc/blob/master/ios/RCTWebRTC/WebRTCModule%2BRTCMediaStream.m

This is my updated code...looks pretty clean...? maybe i need to update the PR?

saghul:
> This is my updated code...looks pretty clean...? maybe i need to update the PR?

No, it doesn't look clean. Please review the diff instead of looking at the code in your editor.

georgewwindsor:
As far as the spacing issues, they are  a little off. as far the "if statements", they need to be there for logical reasons...if you guys would like to skype my ID is george.windsor5.

This requires discussion because im not convinced we are on the same page here. You CANNOT inherit the default "mandatory" constraints, because they MAY conflict with the constraints you pass in. simply "overwriting them" doesn't solve the problem, because it leaves constraints around that the "programmer" would not know were being applied.  

Please review more thoroughly, or explain what want more thoroughly.

zxcpoiu:
@georgewwindsor 

I understand your concern.
said, conflict may occurs when you pass only `max` constraints which are lower than the `default min` constraints, right?

it's true, but I think `to Pull Request a new feature` and `to change the default behavior` are two different things. It's better to make a PR's purpose as simple as possible, in this case, to add support for custom constraints.

because if you change the default behavior, it means all folks will have to change their settings accordingly.

besides, people who using `custom constraints` would probably knows constraints well, and are not likely to put constraints which will cause conflicts.

that's the reason I hope you can implement `min` and `max`, so that you can always put the right constraints to override the default one without any conflicts, and keep the default behavior the same.
everyone is happy.

looks like you did it, so to initialize the default constraints would not cause any problem as long as you understand the constraints and put the right one.

therefore, we can reduce lots of `if / else` logic to make the code `clean`.
`clean` generally is a subjective concept, the objective/relative speaking of clean, in this case, is `to make the code looks consistent to it's neighbor, to the entire repo, as possible`

zxcpoiu:
@georgewwindsor 

you aren't read closely what @saghul said, for those ` if(){ ` / `}else{`
should add space between them, like: `if () {` / `} else {`

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +268,87 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
-    // FIXME The effort above to find a videoDevice value which satisfies the
-    // specified constraints was pretty much wasted. Salvage facingMode for
-    // starters because it is kind of a common and hence important feature on
-    // a mobile device.
+     NSMutableDictionary *mandatory = [[NSMutableDictionary alloc]init];
+       NSDictionary *video = constraints[@"video"];
+       RTCMediaConstraints* customConstraints = nil;
+       if (video != nil){
+         NSDictionary *mandatoryParameters = video[@"mandatory"];
+         if(mandatoryParameters != nil){
```
zxcpoiu: add space `if(` and `){`

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +268,87 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+           if ([mandatoryParameters valueForKey:@"maxFrameRate"] != nil) {
+             NSMutableString *maxFrameRate = nil;
+             if ([[mandatoryParameters valueForKey:@"maxFrameRate"] isKindOfClass:[NSNumber class]]) {
+               maxFrameRate = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"maxFrameRate"]];
+             }else{
```
zxcpoiu: all `}else{` should be `} else {`

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +268,87 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+               maxFrameRate = mandatoryParameters[@"maxFrameRate"];
+             }
+             if (maxFrameRate != nil){
+               [mandatory setObject:maxFrameRate forKey:@"maxFrameRate"];
+             }
+           }
+           if ([mandatoryParameters valueForKey:@"maxHeight"] != nil) {
+             NSMutableString *maxHeight = nil;
+             if ([[mandatoryParameters valueForKey:@"maxHeight"] isKindOfClass:[NSNumber class]]) {
```
zxcpoiu: my opinion for type checking, is to do it on `js` side.
we can check or `toString`, do validation in javascript side
so on the native, we can just set constraint without validation hell.

it's ok to me if you didn't do it in js in the PR.

I can do it as a follow up. just to share my thoughts and my understanding to this repo's general preference.

reduce the line/logic/effort in native code not only make it easier to review/maintain, but also much more agile for adjust / refine / test / experiment. and reduce the size of necessary codes hard-codded in app.

it's the main reason we use `react-native`, isn't it?

saghul:
> As far as the spacing issues, they are a little off. as far the "if statements", they need to be there for logical reasons...if you guys would like to skype my ID is george.windsor5.

I don't understand what the problem is. Just make sure your code matches the style of the rest of the file.



> This requires discussion because im not convinced we are on the same page here. You CANNOT inherit the default "mandatory" constraints, because they MAY conflict with the constraints you pass in. simply "overwriting them" doesn't solve the problem, because it leaves constraints around that the "programmer" would not know were being applied.
> 
> Please review more thoroughly, or explain what want more thoroughly.

I'll rephrase: your code creates some custom constraints in `customConstraints`. All I said is that we should compact the code so you don't set it to the defaults in every "else" clause. Just reset it to null and check if there are no custom constraints at the end and set the defaults then. I never intended to imply merging defaults with those specified by the user.

georgewwindsor:
@saghul @zxcpoiu  I have made the changes requested. Please provide feedback.

1) Fixed Indentation
2) Removed type validation for strings/numbers etc. only strings allowed to be passed
3) created an mutable  mandatory Parameters with the default parameters.
4) No longer using defaultMediaStreamConstraints - Remove?

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +268,43 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
-    RTCAVFoundationVideoSource *videoSource = [self.peerConnectionFactory avFoundationVideoSourceWithConstraints:[self defaultMediaStreamConstraints]];
-    // FIXME The effort above to find a videoDevice value which satisfies the
-    // specified constraints was pretty much wasted. Salvage facingMode for
-    // starters because it is kind of a common and hence important feature on
-    // a mobile device.
+    NSMutableDictionary *mandatory = [[NSMutableDictionary alloc]init];
+    [mandatory setObject:@"720" forKey:@"maxHeight"];
+    [mandatory setObject:@"1280" forKey:@"maxWidth"];
+    [mandatory setObject:@"30" forKey:@"maxFrameRate"];
+    NSDictionary *video = constraints[@"video"];
```
zxcpoiu: we already have `id videoConstraints = constraints[@"video"];` above in https://github.com/smiletime/react-native-webrtc/blob/2836169d659e7f1fce66fce15ec9e386fae8f8ca/ios/RCTWebRTC/WebRTCModule%2BRTCMediaStream.m#L219

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +268,43 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
-    // TODO: Actually use constraints...
-    RTCAVFoundationVideoSource *videoSource = [self.peerConnectionFactory avFoundationVideoSourceWithConstraints:[self defaultMediaStreamConstraints]];
-    // FIXME The effort above to find a videoDevice value which satisfies the
-    // specified constraints was pretty much wasted. Salvage facingMode for
-    // starters because it is kind of a common and hence important feature on
-    // a mobile device.
+    NSMutableDictionary *mandatory = [[NSMutableDictionary alloc]init];
+    [mandatory setObject:@"720" forKey:@"maxHeight"];
+    [mandatory setObject:@"1280" forKey:@"maxWidth"];
+    [mandatory setObject:@"30" forKey:@"maxFrameRate"];
```
zxcpoiu: You can change `defaultMediaStreamConstraints`'s return type from `RTCMediaConstraints` to `NSDictionary` and initialize `mandatory` from it.

this might reviewed by @saghul

georgewwindsor: agree, if you guys think i should use the defaultmediastreamconstraints. It dosnt really do anything tho.

georgewwindsor: Resolved.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
georgewwindsor: Resolved.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +268,43 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+    NSDictionary *video = constraints[@"video"];
+    if (video != nil){
```
georgewwindsor: Resolved.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +268,43 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+      NSDictionary *mandatoryParameters = video[@"mandatory"];
+      if(mandatoryParameters != nil){
```
georgewwindsor: Resolved.

georgewwindsor:
Please review the latest changes.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +265,39 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+        if ([mandatoryParameters valueForKey:@"minHeight"] != nil) {
+          NSMutableString *minHeight = mandatoryParameters[@"minHeight"];
+          [mandatory setObject:minHeight forKey:@"minHeight"];
+        }
+        if ([mandatoryParameters valueForKey:@"minWidth"] != nil) {
+          NSMutableString *minWidth = mandatoryParameters[@"minWidth"];
+          [mandatory setObject:minWidth forKey:@"minWidth"];
+        }
+      }
+    }
```
zxcpoiu: you did not use `defaultMediaStreamConstraints` at all.

if `mandatory` is not touched at the end, `mandatory` should fallback to `defaultMediaStreamConstraints`  for compatibility.

georgewwindsor:
updated. seems real solid.

zxcpoiu:
@saghul 

it seems ok to me.
please check is there anything I've missed

georgewwindsor:
there is a problem with the "id video = constraints" That is not a valid variable. I will be updating shortly.

georgewwindsor:
I have added my fix, I think this is final, EXCEPT that you have to pass in strings. which does not match the android implementation. 
@saghul @zxcpoiu let me know if this works for you, I think we need the format to match on both android and ios though.


{ video:
            { mandatory: {
              minWidth: '500', // Provide your own width, height and frame rate here
              minHeight: '300',
              minFrameRate: '30'
            },

zxcpoiu:
How about use `toString` in JS side?
and accept only one type in native side?

I don't have video use case so can not be sure which is the best way.

georgewwindsor:
the native side only accepts a string right now. you can use a natural JS string ie var blah = '400' and its fine. toString would not be necessary. as long as people put in string values they will be fine.

I am indifferent to a solution. The only real problem is that the android version and the "documentation" for this project uses numbers and not strings. this will cause future confusion for people working on both...

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +265,43 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+    NSArray * mandatoryKeys = [mandatoryParams allKeys];
```
saghul: style: `*mandatoryKeys`

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +265,43 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+    if([mandatoryKeys count]==0){
```
saghul: style: space after if, space between ==

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -38,16 +38,13 @@ @implementation WebRTCModule (RTCMediaStream)
...
-  [[RTCMediaConstraints alloc]
-   initWithMandatoryConstraints:mandatoryConstraints
-   optionalConstraints:nil];
-  return constraints;
+
+- (NSMutableDictionary *)defaultMediaStreamConstraints {
+    NSMutableDictionary *mandatory = [[NSMutableDictionary alloc]init];
+    [mandatory setObject:@"720" forKey:@"maxHeight"];
+    [mandatory setObject:@"1280" forKey:@"maxWidth"];
+    [mandatory setObject:@"30" forKey:@"maxFrameRate"];
```
saghul: instead of repeating these string literals over and over I'd like to see a static constant.

saghul:
> EXCEPT that you have to pass in strings. which does not match the android implementation.

We need to have the implementations match, otherwise we cannot land it. Also, it must match the standard: https://www.w3.org/TR/mediacapture-streams/ So it's integers, NOT strings.

zxcpoiu:
@saghul is right.

* js spec: should pass a number ( type [long](https://heycam.github.io/webidl/#idl-long) )
* java: `videoCapturer.startCapture`, accept `int` as param.
* objc: `RTCMediaConstraints.initWithMandatoryConstraints` accept a `NSDictionary<NSString *, NSString *> *)`

I think the best way to match is to convert it on objc side. ios should only accept number then convert it to string. 

check
IF param is not `nil` AND `isKindOfClass:[NSNumber class]]` THEN
`[NSMutableString stringWithFormat:@"%d"`


this way, all the implementations are matched with the minimum changes. otherwise, we will have a verbose check on js.

georgewwindsor:
yup ill make those changes this evening.  i agree that its needed.

georgewwindsor:
Ok guys I have made the changes requested. I think we should be good to go.

georgewwindsor:
there are 2 other "constraint" related pull requests that can be removed, and about 10 "issues" related to constraints that can be removed.

zxcpoiu:
@georgewwindsor hey, you forgot the style spaces

zxcpoiu:
@georgewwindsor once this landed, will close related issues.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -268,12 +262,43 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+          [mandatory setObject:minHeight forKey:@"minHeight"];
+        }
+        if ([mandatoryParameters valueForKey:@"minWidth"] != nil && [[mandatoryParameters valueForKey:@"minWidth"] isKindOfClass:[NSNumber class]]) {
+          NSMutableString *minWidth = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"minWidth"]];
+          [mandatory setObject:minWidth forKey:@"minWidth"];
+        }
+      }
+    }
+    NSArray * mandatoryKeys = [mandatory allKeys];
+    if([mandatoryKeys count]==0){
```
zxcpoiu: spacing, should be `if ([mandatoryKeys count] == 0) {`

georgewwindsor: got it.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -266,14 +260,44 @@ - (void)getUserVideo:(NSDictionary *)constraints
...
+          NSMutableString *minHeight = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"minHeight"]];
+          [mandatory setObject:minHeight forKey:@"minHeight"];
+        }
+        if ([mandatoryParameters valueForKey:@"minWidth"] != nil && [[mandatoryParameters valueForKey:@"minWidth"] isKindOfClass:[NSNumber class]]) {
+          NSMutableString *minWidth = [NSMutableString stringWithFormat:@"%@", (NSInteger)mandatoryParameters[@"minWidth"]];
+          [mandatory setObject:minWidth forKey:@"minWidth"];
+        }
+      }
+    }
+    NSArray * mandatoryKeys = [mandatory allKeys];
```
zxcpoiu: should be `*mandatoryKeys`

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -293,7 +292,7 @@ - (void)getUserVideo:(NSDictionary *)constraints
       }
     }
     NSArray * mandatoryKeys = [mandatory allKeys];
-    if([mandatoryKeys count]==0){
+    if([mandatoryKeys count] == 0){
```
zxcpoiu: should be `if ([mandatoryKeys count] == 0) {`

space between `if` and `(` 
space between `)` and `{`

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
georgewwindsor: done

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
georgewwindsor: done. sorry my eyes are not trained to see it.:)

saghul:
@zxcpoiu This morning something clicked in my head. I was recently working on the Android permissions code, and noticed we do a similar thing as we do here.

I think we should move the parsing of resolutions to JS. That is, the native side would always get some 'mandatory' object with  'minHeight', 'minWidth' and 'minFps', which me make sure to set from JS. WDYT?

zxcpoiu:
@saghul 

Yes, that would be great.

ny preference is always put logic in js side and keep native side as simple as possible. just like we moved some default values from native to js.

putting logics and default values on js would speed up development process and easier for folks to test/customization/contributes.

so yes! as long as we keep `use case agnostic` and `flexible`  in mind.

we can use your PR as a follow up and close this PR and related issues

georgewwindsor:
kk let me know if you guys need me to make further changes.

georgewwindsor:
@zxcpoiu @saghul , if you guys dont have an existing PR for the way you would "like" to fix it, then i recommend merging this in. It is quite a blocking issue for literally dozens of peoples projects. It dosnt hurt me, im on a working fork.  i honestly think you guys need to prioritize getting this functionality merged in. This issue has been "open" for 18 months. Without it, your library is really only good for 2-way communication....and cannot support multiple parties at once with any reliability.

Additionally, I have made like 10 revisions doing exactly what you guys requested, so if you throw it away, I wont be contributing anything else, because you will have wasted my time.

zxcpoiu:
whoa, don't be mad mate.

we are only just a time limited enthusiast like you to dedicate our time in this repo.

it's unfair to said that "waste your time"
actually, why this PR rejected back and fourth several is mainly you did not look carefully and didn't modify what we repeatedly wrote in our review.

we all are not perfect and can not always toward the right direction. that's the reason we use github as a corporation platform to have discussions that make this repo better.

if you are also agree that doing constraints stuff in JS side is a better way, then you would not feel something "wasting time"

also, if you really agreed and fully understand why it's better to do it in JS, you can always fire a PR instead wrote emotional messages, isn't it?

saghul:
@georgewwindsor I won't enter any argument here. If this the attitude you want to have, good luck, I won't be looking at your PRs anymore. If you feel like calming down and finding the best way forward, let me know.

georgewwindsor:
ffs guys dont take my comments so personally. I stated some FACTS, there was a lot less emotion in it than you you guys are reading into it.  Why would i keep making PRs if you didnt merge them in? As I stated, I have a fork that I can keep working on, so none of this effects me. If you guys don't want help, thats fine. I have several other fixes for issues that I have developed, but ill just keep it on my fork.  Apparently you guys cant hear negativity without getting defensive.

zxcpoiu:
:expressionless: :expressionless: :expressionless:

we better stop meaningless discussion here. :cold_sweat:
I presume the sun will rise tomorrow and earth still circling. :alien:

saghul:
FTR, what needs to be done here is: perform all checks in JS land and make sure the "mandatory" section of the video constraints contains valid resolution / fps information, or the 720p/30fps defaults are applied.

The Android and iOS code would then read this, but without all the error checking.

PR #304, opened 2017-06-28
android: use the new permissions system on  Android API >= 23

saghul:
Do it dynamically, based on the target SDK version as specified by the
application.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
@@ -1,0 +1,133 @@
+package com.oney.WebRTCModule;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Build;
+import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
+import android.os.ResultReceiver;
+import android.support.v4.app.ActivityCompat;
+import android.support.v4.content.ContextCompat;
+
+
+/**
+ * Helper module for dealing with dynamic permissions, introduced in Android M (API level 23).
+ */
+public class PermissionUtils {
+    /**
+     * Constants for internal fields in the <tt>Bundle</tt> exchanged between the activity
+     * requesting the permissions and the auxiliary activity we spawn for this purpose.
+     */
+    private static final String GRANT_RESULT = "GRANT_RESULT";
+    private static final String PERMISSION = "PERMISSION";
+    private static final String REQUEST_CODE = "REQUEST_CODE";
+    private static final String RESULT_RECEIVER = "RESULT_RECEIVER";
+
+    /**
+     * Incrementing counter for permissions requests. Each request must have a unique numeric code.
+     */
+    private static int requestCode;
+
+    /**
+     * Requests the given permission. The callback will we called with the requested permission and
+     * the granted result. See: https://developer.android.com/reference/android/content/pm/PackageManager.html#PERMISSION_GRANTED
+     * for the result constants.
+     *
+     * @param context Application context / activity.
+     * @param permission Permission which the application is interested in requesting.
+     * @param callback Callback where the results will be reported.
+     */
+    public static void requestPermission(Context context,
+                                         String permission,
+                                         final Callback callback) {
+        ResultReceiver resultReceiver = new ResultReceiver(new Handler(Looper.getMainLooper())) {
+            @Override
+            protected void onReceiveResult (int resultCode, Bundle resultData) {
+                String permission = resultData.getString(PERMISSION);
+                int grantResult = resultData.getInt(GRANT_RESULT);
+
+                callback.invoke(permission, grantResult);
+            }
+        };
+
+        requestCode++;
+
+        Intent permissionsIntent = new Intent(context, PermissionHelperActivity.class);
+        permissionsIntent.putExtra(RESULT_RECEIVER, resultReceiver);
+        permissionsIntent.putExtra(PERMISSION, permission);
+        permissionsIntent.putExtra(REQUEST_CODE, requestCode);
+        permissionsIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+
+        context.startActivity(permissionsIntent);
+    }
+
+    /**
+     * Helper activity for requesting permissions. Android only allows requesting permissions from
+     * an activity and the result is reported in the <tt>onRequestPermissionsResult</tt> method.
+     * Since this package is a library we create an auxiliary activity and communicate back the
+     * results using a <tt>ResultReceiver</tt>.
+     */
+    public static class PermissionHelperActivity extends Activity {
+        int requestCode;
+        ResultReceiver resultReceiver;
+        String permission;
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public void onRequestPermissionsResult(int requestCode,
+                                               String permissions[],
+                                               int[] grantResults) {
+            Bundle resultData = new Bundle();
+            resultData.putString(PERMISSION, permissions[0]);
+            resultData.putInt(GRANT_RESULT, grantResults[0]);
+            resultReceiver.send(requestCode, resultData);
+            finish();
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        protected void onStart() {
+            super.onStart();
+
+            resultReceiver = this.getIntent().getParcelableExtra(RESULT_RECEIVER);
+            permission = this.getIntent().getStringExtra(PERMISSION);
+            requestCode = this.getIntent().getIntExtra(REQUEST_CODE, 0);
+
+            // Here we test for the target SDK version with which *the app* was compiled. If we use
+            // Build.VERSION.SDK_INT that would give us the API version of the device itself, not
+            // the version the app was compiled for. When compiled for API level < 23 we must still
+            // use old permissions model, regardless of the Android version on the device.
+            final int targetSdkVersion = getApplicationInfo().targetSdkVersion;
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M
+                    && targetSdkVersion >= Build.VERSION_CODES.M) {
+                ActivityCompat.requestPermissions(
+                    this,new String[]{permission}, requestCode);
+            } else {
+                // In the old permissions model the permissions are granted at install time, we can
+                // check for them directly.
+                Bundle resultData = new Bundle();
+                resultData.putString(PERMISSION, permission);
+                resultData.putInt(
+                    GRANT_RESULT, ContextCompat.checkSelfPermission(this, permission));
+                resultReceiver.send(requestCode, resultData);
+                finish();
+            }
+        }
+    }
+
+    /**
+     * Helper class for reporting back permission results.
+     * See: https://developer.android.com/reference/android/content/pm/PackageManager.html#PERMISSION_GRANTED
+     * for the result constants.
+     */
+    public interface Callback {
+        void invoke(String permission, int grantResult);
+    }
+}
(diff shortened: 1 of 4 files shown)


zxcpoiu:
@saghul 

is there any restriction that we can not use react-native provided permission module?

saghul:
Yes, we cannot use it. It relies on the fact that your Activity is a ReactActivity, but that need not be the case if you are embedding react-native on an Android app / library. This is the case with Jitsi Meet, so I implemented it this way, which makes it flexible and works regardless of the Activity type.

zxcpoiu:
thanks, good for flexible

8BallBomBom:
Getting this error.

`
Calling startActivity() from outside of an Activity context requires the FLAG_ACTIVITY_NEW_TASK flag. Is this really what you want?
`

Running wix react-native-navigation, api 22 along with launchmode as singletask.

saghul:
@8BallBomBom thanks for the report, I must have missed it. I'll take a look.

saghul:
@8BallBomBom I included a fix for that, can you please test again? Thanks!

saghul:
Thanks for testing!

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
```diff
@@ -0,0 +1,133 @@
...
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public void onRequestPermissionsResult(int requestCode,
+                                               String permissions[],
+                                               int[] grantResults) {
+            Bundle resultData = new Bundle();
+            resultData.putString(PERMISSION, permissions[0]);
+            resultData.putInt(GRANT_RESULT, grantResults[0]);
```
lyubomir: A few observations of mine about this PR will follow. I've taken care of them and I'm putting them here in writing for the sake of clarity and completeness.

The super's documentation says: It is possible that the permissions request interaction with the user is interrupted. In this case you will receive empty permissions and results arrays which should be treated as a cancellation.

Hence, the above seem opportunistic to me.

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
```diff
@@ -0,0 +1,133 @@
...
+                    && targetSdkVersion >= Build.VERSION_CODES.M) {
+                ActivityCompat.requestPermissions(
+                    this,new String[]{permission}, requestCode);
+            } else {
+                // In the old permissions model the permissions are granted at install time, we can
+                // check for them directly.
+                Bundle resultData = new Bundle();
+                resultData.putString(PERMISSION, permission);
+                resultData.putInt(
+                    GRANT_RESULT, ContextCompat.checkSelfPermission(this, permission));
```
lyubomir: https://developer.android.com/training/permissions/requesting.html says that `ContextCompat.checkSelfPermission` should be called in both branches of the above if in order to avoid a call to `ActivityCompat.requestPermissions` when the permission is granted already.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -445,270 +449,301 @@ private MediaConstraints parseMediaConstraints(ReadableMap constraints) {
...
-                //   2. all camera support level should greater than LEGACY
-                //   see: https://developer.android.com/reference/android/hardware/camera2/CameraCharacteristics.html#INFO_SUPPORTED_HARDWARE_LEVEL
-                // TODO Enable camera2 enumerator
-                if (false && Camera2Enumerator.isSupported(context)) {
-                    Log.d(TAG, "Creating video capturer using Camera2 API.");
-                    videoCapturer = createVideoCapturer(
-                        new Camera2Enumerator(context), isFacing, sourceId);
+        final PermissionUtils.Callback callback = new PermissionUtils.Callback() {
+            @Override
+            public void invoke(String permission, int grantResult) {
```
lyubomir: The implementation does not comply with the standard. If multiple permissions are to be requested and at least one is denied, the whole getUserMedia is to fail.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -445,270 +449,301 @@ private MediaConstraints parseMediaConstraints(ReadableMap constraints) {
...
-                // https://www.w3.org/TR/mediacapture-streams/#dom-mediadevices-getusermedia
-                // with respect to distinguishing the various causes of failure.
-                errorCallback.invoke(/* type */ null, "Failed to obtain audio");
-                return;
+        };
+
+        Context context = getReactApplicationContext();
+        for (String permission: requestPermissions) {
+            PermissionUtils.requestPermission(context, permission, callback);
+        }
```
lyubomir: 1. I don't understand how the above even works given that we issue multiple requests in a loop. I believe that, since we are limited by `PermissionUtils` to requesting 1 permission at a time, we should request the permissions in a sequence with the next permission being requested after the previous permission was granted. And as soon as 1 permission is denied, the whole sequence should break and `getUserMedia` should fail.

2. Moreover, `errorCallback` is never invoked to report a denied permission which makes the implementation incomplete.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -445,270 +449,301 @@ private MediaConstraints parseMediaConstraints(ReadableMap constraints) {
...
     @ReactMethod
-    public void getUserMedia(ReadableMap constraints,
-                             Callback    successCallback,
-                             Callback    errorCallback) {
-        AudioTrack audioTrack = null;
-        VideoTrack videoTrack = null;
-        WritableArray tracks = Arguments.createArray();
+    public void getUserMedia(final ReadableMap constraints,
+                             final Callback    successCallback,
+                             final Callback    errorCallback) {
```
lyubomir: The number of lines within this method is just too big and that hinders readability, understanding, and maintenance. I think it should've been broken into smaller methods. Moreover, the whole `getUserMedia` implementation along with its supporting methods is getting bigger and bigger so I believe it should be extracted out of WebRTCModule.

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
```diff
@@ -0,0 +1,133 @@
...
+     * Requests the given permission. The callback will we called with the requested permission and
+     * the granted result. See: https://developer.android.com/reference/android/content/pm/PackageManager.html#PERMISSION_GRANTED
+     * for the result constants.
+     *
+     * @param context Application context / activity.
+     * @param permission Permission which the application is interested in requesting.
+     * @param callback Callback where the results will be reported.
+     */
+    public static void requestPermission(Context context,
+                                         String permission,
```
lyubomir: I find `PermissionUtils` to be short sighted given that we know we can be in a situation in which we may have to request multiple permissions. I am aware that Jitsi Meet does not request both audio and video in 1 `getUserMedia` call but we're not in Jitsi Meet here.

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
```diff
@@ -0,0 +1,133 @@
...
+        context.startActivity(permissionsIntent);
+    }
+
+    /**
+     * Helper activity for requesting permissions. Android only allows requesting permissions from
+     * an activity and the result is reported in the <tt>onRequestPermissionsResult</tt> method.
+     * Since this package is a library we create an auxiliary activity and communicate back the
+     * results using a <tt>ResultReceiver</tt>.
+     */
+    public static class PermissionHelperActivity extends Activity {
```
lyubomir: `PermissionHelperActivity` seems very invasive to me because it "replaces" the `Activity` which effectively requests the permissions by making the `getUserMedia`. Google mentions a "headless fragment" which attaches to an `Activity` and is able to request runtime permissions. Given that `WebRTCModule` appears in source code to be able to give us a `currentActivity` to which it is attached, have "headless fragments" been explored as an alternative to `PermissionHelperActivity`?

lyubomir: A headless fragment instead of `PermissionHelperActivity` is exactly what we need here!

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
lyubomir: The cancellation scenario presents an unclear situation. For example:
1. JS calls `getUserMedia`
2. `PermissionUtils` requests a permission i.e. a dialog is presented to the user.
3. The user switches away from the app, possibly on a subject unrelated to the permission request.
4. The user comes back to the application.

At step 4, `onRequestPermissionsResult` will signal a cancellation. But not all permissions will have been reported as granted or denied to `callback` so `getUserMedia` will be stuck waiting for an answer to a question that is no longer posed to the user because `finish()` was invoked (presuming there's no unguarded access to elements at index 0, of course).

One question to be answered is whether `PermissionUtil` should ask for the permission again since technically it was not answered by the user. I guess we can look at what iOS gives as an experience in such a scenario.

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
```diff
@@ -0,0 +1,133 @@
...
+
+        requestCode++;
+
+        Intent permissionsIntent = new Intent(context, PermissionHelperActivity.class);
+        permissionsIntent.putExtra(RESULT_RECEIVER, resultReceiver);
+        permissionsIntent.putExtra(PERMISSION, permission);
+        permissionsIntent.putExtra(REQUEST_CODE, requestCode);
+        permissionsIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+
+        context.startActivity(permissionsIntent);
```
lyubomir: I do not like it that in order to check whether the requested permissions have already been granted the implementation starts an `Activity` (which let's not forget goes on top of the `Activity` which requests the permissions!) given that all requested permissions may have already been granted.

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
saghul: Good point. I was under the impression that when a permission i requested we'd always get an answer.

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
saghul: Good catch!

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
saghul: I don't disagree :-)

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
saghul: I handled that differently, this was checked further down, after all permissions were checked. Then the logic worked out if we had the ones that were needed vs the ones that were requested.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
saghul: It works because the callback is designed to do the processing once all permission requests have finished. This will fire 2 activities and once both are completed the code will look at the results.

The `errorCallback` would be executed after all permissions were evaluated.

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
saghul: The code does accommodate for that, though I designed that helper to handle them one by by one for simplicity in the implementation. This was deliberate.

android/src/main/java/com/oney/WebRTCModule/PermissionUtils.java
saghul: Interesting! I was not aware of that, sorry. Also, FWIW, my Google Plxel does it the sane way, that is, it doesn't replace the entire screen with a white full-blown view. 😞

PR #310, opened 2017-07-04
misc: add issue template

saghul:
Asking the same quations over and over again gets tiresome quickly. Let's try to have people file better reports.

@zxcpoiu what do you think?

For reference: https://github.com/blog/2111-issue-and-pull-request-templates


zxcpoiu:
hey, sorry for late.

it looks great, thanks!
I will find time to do some FAQ documents.

PR #311, opened 2017-07-05
android: create EGL context on render thread

saghul:
Port of [0] since we bundle our own SurfaceViewRenderer. It should fix
crashes on some Jelly Bean devices.

[0]:
https://chromium.googlesource.com/external/webrtc/codesearch/+/4ed5b9f62d7850fc5d9b2a006d3c368a9183c16d

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/SurfaceViewRenderer.java
+++ b/android/src/main/java/com/oney/WebRTCModule/SurfaceViewRenderer.java
@@ -155,8 +155,9 @@ public class SurfaceViewRenderer extends SurfaceView
    * |drawer|. It is allowed to call init() to reinitialize the renderer after a previous
    * init()/release() cycle.
    */
-  public void init(EglBase.Context sharedContext, RendererCommon.RendererEvents rendererEvents,
-      int[] configAttributes, RendererCommon.GlDrawer drawer) {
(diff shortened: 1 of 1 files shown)


zxcpoiu:
@saghul 

For some non-breaking changes' PR,  I believe you have tested it enough. Since we can never know if it breaks in any edge cases and thanks to git, revert process is not that complex.

I would suggest faster development circle that just merge these non-breaking PRs for more folks can test on it.

we can treat master is blooding-edge then release a new version if there is no bad news, said, two weeks or so.

what do you think?

saghul:
The PRs I make I double-check with @lyubomir, since we work together and communicate daily. If people are in a rush they can use their own forks and go to town with them, that's the beauty of Open Source. I won't compromise the quality because someone's comments about our speed, sorry.

zxcpoiu:
:laughing: easy man. I know you always keep quality.

Good to know that @lyubomir with you.
I thought you are waiting each other. and I don't have solid knowledge to help to review this kind of egl stuff. just want to let you know if I disappeared for few days, just go ahead.

saghul:
@zxcpoiu thanks for the support! Of course you know where my answer come from, you are not to blame for the snark, sorry about that! :-)

PR #316, opened 2017-07-11
Fix the compilation errors in pods project when using `use_frameworks!` in the pod file

mahmoud-cliqz:
Fix the compilation errors in resulting pods project (couldn’t find React headers) when `using use_frameworks!` in the referencing pod file


saghul:
Landed in 6313d29, cheers and sorry for the delay!

PR #317, opened 2017-07-11
android: HW video acceleration + camera2

saghul:
See the individual commit logs for details.


saghul:
Pushed an extra patch with an update to WebRTC M58 which includes a cherry-pick which fixes a crash when using textures, which is the case when camera2 is used. The details of the cherry-pick are in the commit message.

zxcpoiu:
I will do doc update once this landed.
ping me if a new release is required.

saghul:
Thanks, will do. We need to do more testing, but things are looking good! :-)

-Saúl

PR #326, opened 2017-08-02
handle RN 0.47 breaking change

brunsy:
RN 0.47 removed createJSModules from ReactPackage [commit](https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8)


saghul:
This is not a good fix, this package needs to work with earlier versions of RN. What did they replace it with?

lukecwilliams:
Removing just the @Override line should work yes?  Others have said they tested it with 0.46.x, and @Override is an optional declaration in Java, e.g. https://news.ycombinator.com/item?id=7866204

PR #334, opened 2017-08-23
[Android] Fix leaking remote streams

paweldomas:
Remote streams were leaking on Android, because when a peer connection was closed tracks were not removed from WebRTCModule. This PR stores tracks and streams in PeerConnectionObserver. Also adapts the iOS part to use similar strategy instead of special react tag syntax.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -3,6 +3,9 @@ package com.oney.WebRTCModule;
 import java.io.UnsupportedEncodingException;
 import java.lang.ref.SoftReference;
 import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
 
 import android.support.annotation.Nullable;
 import android.util.Base64;
@@ -33,6 +36,8 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         = new SparseArray<DataChannel>();
     private final int id;
     private PeerConnection peerConnection;
+    final Map<String, MediaStream> remoteMediaStreams;
+    final Map<String, MediaStreamTrack> remoteMediaStreamTracks;
     private final WebRTCModule webRTCModule;
 
     /**
@@ -47,6 +52,8 @@ class PeerConnectionObserver implements PeerConnection.Observer {
     PeerConnectionObserver(WebRTCModule webRTCModule, int id) {
         this.webRTCModule = webRTCModule;
         this.id = id;
+        this.remoteMediaStreams = new HashMap<String, MediaStream>();
+        this.remoteMediaStreamTracks = new HashMap<String, MediaStreamTrack>();
     }
 
     PeerConnection getPeerConnection() {
@@ -59,6 +66,10 @@ class PeerConnectionObserver implements PeerConnection.Observer {
 
     void close() {
          peerConnection.close();
+
+         remoteMediaStreams.clear();
+
+         remoteMediaStreamTracks.clear();
 
          // Unlike on iOS, we cannot unregister the DataChannel.Observer
          // instance on Android. At least do whatever else we do on iOS.
@@ -186,8 +197,9 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         MediaStreamTrack track = null;
         if (trackId == null
                 || trackId.isEmpty()
-                || (track = webRTCModule.mMediaStreamTracks.get(trackId))
-                    != null) {
+                || (track = webRTCModule.localMediaStreamTracks.get(trackId))
+                    != null
+                || (track = remoteMediaStreamTracks.get(trackId)) != null) {
             peerConnection.getStats(
                     new StatsObserver() {
                         @Override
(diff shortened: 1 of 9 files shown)


zxcpoiu:
@paweldomas this PR still checks with @saghul and @lyubomir right?

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -361,11 +360,49 @@ String getNextTrackUUID() {
...
+    }
+
+    MediaStreamTrack getTrackForId(String trackId) {
+        MediaStreamTrack track = localMediaStreamTracks.get(trackId);
+
+        if (track != null) {
+            return track;
+        }
+
+        for (int idx = 0; idx < mPeerConnectionObservers.size(); idx++) {
```
saghul: `for (PeerConnectionObserver pco : mPeerConnectionObservers) {` ?

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -327,34 +331,40 @@ - (void)peerConnection:(RTCPeerConnection *)peerConnection didChangeSignalingSta
...
-    NSString *tag = [self reactTagForID:track.trackId RTCPC:peerConnection];
-    self.remoteTracks[tag] = track;
+    peerConnection.remoteTracks[track.trackId] = track;
     [tracks addObject:@{@"id": track.trackId, @"kind": track.kind, @"label": track.trackId, @"enabled": @(track.isEnabled), @"remote": @(YES), @"readyState": @"live"}];
   }
 
-  NSString *streamReactTag =  [self reactTagForID:stream.streamId RTCPC:peerConnection];
-  self.remoteStreams[streamReactTag] = stream;
+  NSString *streamReactTag;
+  // Make sure ID does not exist across local and remote streams (for any peerConnection)
```
saghul: Is this check really necessary (same question for the Java side). I checked the docs on both platforms and the UUIDs are version 4, which uses a PRNG, so its unicity should be guaranteed.

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -327,34 +331,40 @@ - (void)peerConnection:(RTCPeerConnection *)peerConnection didChangeSignalingSta
...
-  if (mediaStream) {
-    for (RTCVideoTrack *track in mediaStream.videoTracks) {
-      [self.remoteTracks removeObjectForKey:[self reactTagForID: track.trackId RTCPC: peerConnection]];
-    }
-    for (RTCAudioTrack *track in mediaStream.audioTracks) {
-      [self.remoteTracks removeObjectForKey:[self reactTagForID: track.trackId RTCPC: peerConnection]];
-    }
-    [self.remoteStreams removeObjectForKey:streamReactTag];
+  NSArray *keysArray = [peerConnection.remoteStreams allKeysForObject:stream];
+  // We assume there can be only one object for 1 key
```
saghul: maybe log if there is more than one? that would be a bug

ios/RCTWebRTC/WebRTCModule.m
```diff
@@ -29,20 +30,29 @@ - (instancetype)init
...
     _localTracks = [NSMutableDictionary new];
   }
   return self;
 }
 
+- (RTCMediaStream*)streamForTag:(NSString*)reactTag
+{
+    RTCMediaStream *stream = _localStreams[reactTag];
+    if (!stream) {
+        for (NSNumber *peerConnectionId in _peerConnections) {
```
saghul: is there no nicer way to iterate this?

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
paweldomas: Maybe it's not really necessary, but since it's already there I don't see why not

ios/RCTWebRTC/WebRTCModule.m
paweldomas: I will take a look

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
paweldomas: "foreach not applicable to SparseArray"

paweldomas: hmmm it's IDE complaining, but it seems to work

paweldomas: nope, stupid me was installing webrtc-react-native from Github instead of local dev - not possible

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -284,10 +319,21 @@ public void onAddStream(MediaStream mediaStream) {
 
     @Override
     public void onRemoveStream(MediaStream mediaStream) {
-        String streamReactTag = webRTCModule.onRemoveStream(mediaStream);
+        String streamId = mediaStream.label();
```
paweldomas: Then I have to wrap the line - will change it but honestly this kind of changes are waste of time

ios/RCTWebRTC/WebRTCModule.m
paweldomas: I don't see any easier way as far as my Objective C and stackoverflow knowledge goes...

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
saghul: I think it may simplify the code in Android or iOS, because you wouldn't need access to all streams / tracks to check.

ios/RCTWebRTC/WebRTCModule.m
saghul: Ack

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
saghul: Ack

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
paweldomas: It won't because this access is necessary anyway for the view manager which needs to obtain stream for react tag. It needs to search all peerconnections an local streams.

paweldomas: But I was thinking the same at some point

saghul: Ack

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -242,9 +254,26 @@ public void onIceGatheringChange(PeerConnection.IceGatheringState iceGatheringSt
...
 
+    public String getReactTagForStream(MediaStream mediaStream) {
+        String reactTag = null;
+        for (Iterator<Map.Entry<String, MediaStream>> i
+             = remoteMediaStreams.entrySet().iterator();
+             i.hasNext();) {
+            Map.Entry<String, MediaStream> e = i.next();
+            if (e.getValue().equals(mediaStream)) {
+                reactTag = e.getKey();
+                i.remove();
```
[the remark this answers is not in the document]
paweldomas: Of course not, thanks ! It was a copy/paste mistake, but because it's used only in onRemoveStream the stream was supposed to be removed, but it should be done elsewhere.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -572,32 +609,9 @@ public void peerConnectionSetConfiguration(ReadableMap configuration, final int
         peerConnection.setConfiguration(parseRTCConfiguration(configuration));
     }
 
-    String onAddStream(MediaStream mediaStream) {
-        String id = mediaStream.label();
-        String reactTag = null;
-        // The native WebRTC implementation has a special concept of a default
-        // MediaStream instance with the label default that the implementation
-        // reuses.
-        if ("default".equals(id)) {
```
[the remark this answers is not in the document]
paweldomas: I wanted to ask you about that, but I forgot eventually. It is unclear to me how this should be handled. Can we discuss that outside of GH it would take too much time here ?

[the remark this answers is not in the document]
paweldomas: I added it here https://github.com/oney/react-native-webrtc/pull/334/files#diff-8f28661663601f7caa97d6fd5bf1bf30R276

PR #346, opened 2017-09-12
RTCDataChannel.send() doesn't support subarrays

jd20:
This change ensures that if the passed TypedArray is a subview on the underlying ArrayBuffer, that we preserve the byteOffset and byteLength when converting to an Uint8Array for base64 encoding.


zxcpoiu:
thanks! :+1: 

fixed #347

PR #350, opened 2017-09-14
Migrate to 'prop-types' module

jd20:
Fixes a PropTypes deprecation warning that keeps getting logged, see: https://facebook.github.io/react/warnings/dont-call-proptypes.html

### DIFF
--- a/RTCView.js
+++ b/RTCView.js
@@ -5,7 +5,7 @@ import {
   NativeModules,
   requireNativeComponent,
 } from 'react-native';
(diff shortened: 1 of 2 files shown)


saghul:
Thanks! Can you please also add a dependency for it? https://www.npmjs.com/package/prop-types#how-to-depend-on-this-package

jd20:
Done!

saghul:
Cheers!

PR #351, opened 2017-09-14
android: avoid storing a global context on RTCVideoViewManager

saghul:
It has a reference to the Activity, so this creates a (temporary) leak
because it keeps a reference to it until a new view is created.

Get a reference to the current context dynamically instead.


saghul:
@paweldomas I couldn't request your review for some reason. Can you PTAL, too?

PR #140, opened 2016-09-26
Add iOS media constraint support

Qard:
Probably could use some cleanup, but I got media constraints working on iOS.

PR #359, opened 2017-10-04
Ios media constraints

RangerMauve:
Same as #140 but merged with the latest version of master


ramneekhanda:
whats the status of this?

zxcpoiu:
sorry for late, I will find time to review it.

@ramneekhanda have you tried it ?

RangerMauve:
This is just #140 but with fixes for the merge conflicts

zxcpoiu:
@RangerMauve yes, I know. thank you for the rebasing :)

That PR does not get reviewed before.
We need to check this PR is compatible with current logic and the discussions in #288 

There are fairly large changes since then.

RangerMauve:
Looking at the code, this seems to be following a similar code style, and it's doing the conversion to strings in the Objective-C side instead of the JS side like y'all proposed in #288.

zxcpoiu:
@RangerMauve 

The way proposed in #288 should be:

* we parse all video constraints in JS side, so native side would always receive `mandatory constraints`
* `mandatory constraints` can only be either `false` or a `dictionary` contains `minHeight` / `minWidth` / `minFps` ...etc. And native can feed constraints directly to the underlying webrtc without bother parsing.


the main benefit would be we can change the JS api faster aligned with the spec without change the underlying webrtc api. And lot more easier to test/maintain/customization


pseudo code would be:

## JS:

```javascript
let videoConstraints = false;
if video:
    if typeof(video) === 'boolean':    
        videoConstraints = defaultVideoConstraints
    else if video is dictionary:
        // --- parse every thing we needed and some checks
        videoConstraints = {
            ...defaultVideoConstraints,
            ...video,
        }
```

## Native

just consume what JS passed in.

zxcpoiu:
In this PR:

* the `mandatory` constraints should be a `Dictionary` instead an `Array`.

since we only handle `sourceID` in `optionalConstraints`, I think we can temporary ignore `optionalConstraints`.

Then create a PR that do the JS parsing like above pesudo code.
The Native change should be small, maybe just one line to consume mandatory constraints, like:
`[[RTCMediaConstraints alloc] initWithMandatoryConstraints:mandatory optionalConstraints:nil];`

zxcpoiu:
I create a new PR do what I said above.

moved to #370

please help to test and review

Issue #396, opened 2018-01-01
RTCPeerConnection.getStats() returns a nested string blob


jd20:
While trying to integrate simple-peer with react-native-webrtc, I've found that the output of getStats() is non-standard, which can cause problems with other libraries that expect a browser-like output.

#### Expected behavior
Calling RTCPeerConnection.getStats() should return an array of structured objects. For example:

```
(18) [{…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}, {…}]
0: {id: "googLibjingleSession_7269088161857653593", type: "googLibjingleSession", timestamp: 1514770646094.905, values: Array(1)}
1: {id: "datachannel_0", type: "datachannel", timestamp: 1514770646094.905, values: Array(4)}
...
16: {id: "Conn-audio-1-4", type: "googCandidatePair", timestamp: 1514770646094.905, values: Array(21)}
17: {id: "Cand-00VNzHwh", type: "localcandidate", timestamp: 1514770646094.905, values: Array(6)}
length: 18
```

#### Observed behavior
Instead, it's returning a single element array, with a giant JSON string:

`["[{"id":"googLibjingleSession_7269088161857653593",…5"},{"portNumber":"54298"},{"transport":"udp"}]}]"]`

#### Steps to reproduce the problem
Create an instance of RTCPeerConnection, and then call getStats().

#### Platform information

* **React Native version**: 0.51.0
* **Plugin version**: ^1.58.3
* **OS**: iOS
* **OS version**: 9.0

#### Proposed Fix
In RTCPeerConnection.js:204, there's a long comment about how the getStats() data is passed across the bridge as a string, for performance reasons. However, the following check is `typeof stats === 'string'`, whereas what is coming across the bridge from WebRTCModule+RTCPeerConnection.m seems to always be an array with a single-element string, so that check is never matching. Not sure if the Android code has the same problem.

An easy fix would be to just check for an array with a single-element string, and convert it. I tried this and it worked OK for me locally, maybe there's a better solution, on the native side.

```javascript
        if (Array.isArray(stats) && stats.length === 1 && typeof stats[0] === 'string') {
          stats = stats[0]
        }
        if (typeof stats === 'string') {
          // Original code ...
        }
        return stats;
```

saghul:
Sounds about right. Care to send a PR?

jd20:
Sure, submitted #398.

[...]

PR #398, opened 2018-01-02
Fix getStats() output

jd20:
See #396. This checks if the value from getStats() is a single-element array containing a string, and if so will unbox it from the array, so it can be parsed as JSON.


zxcpoiu:
@jd20 
Thanks for your PR and detailed explanation in #396
I'm going to merge this since this doesn't break things.

Although github can squash and merge, it's better to squash it before send the PR to avoid `oops` from reviewer at the first glance :)

jd20:
Ah, thanks for the tip, I'm kind of new to using Github and kept wondering why my PR's were ending up like that, even though my branch would only have one or two commits on it.

PR #415, opened 2018-02-03
Tweak `getStats()` to return results in spec-compliant format

nazar-pc:
Read https://github.com/feross/simple-peer/issues/221 for background.

This PR converts previous `getStats()` results in format that is as close to the spec as possible.

### DIFF
--- a/RTCPeerConnection.js
+++ b/RTCPeerConnection.js
@@ -205,7 +205,23 @@ export default class RTCPeerConnection extends EventTarget(PEER_CONNECTION_EVENT
             throw e;
           }
         }
-        return stats;
+        // Should be RTCStatsReport, but as RTCStatsReport is Map-like and
+        // is not available for constructing directly, we can use regular
+        // Map instead
+        var specStats = new Map;
+        for (let item of stats) {
+          // Replicates RTCStats dictionary from the spec
+          let specItem = {
+            id: item.id,
+            type: item.type,
+            timestamp: item.timestamp
+          };
+          // Destructure weirds `values` array; each value is an object
+          // with single key, lets assign all of those keys
+          Object.assign(specItem, ...item.values);
+          specStats.set(specItem.id, specItem)
+        }
+        return specStats;
       });
     } else {
       console.warn('RTCPeerConnection getStats not supported');


saghul:
We are currently generating the legacy ones on iOS... how is this solving that?

nazar-pc:
Didn't get it, could you elaborate?
I've tested this on Android and one more person tested this on both Android and iOS.

nazar-pc:
Unfortunately I can't answer your questions fully, as I'm not familiar with native side of things.

From what I understand so far, the primary issue is how serialization of `getStats()` is happening on native side. But since I'm not Android or iOS developer, I'm not really qualified to fix that and I also only found Android stuff that generate it in this repository.

Regardless, I don't think this PR will cause much issues in future, as Chromium Nightly 64 (and few older versions, possibly even stable down to 58) return the `getStats()` results in the format that this PR provides (and this format corresponds to current version of the WebRTC spec), so you have to do this at some point, better sooner than later.

See 8.3 and 8.4 from the spec: https://www.w3.org/TR/webrtc/#rtcstatsreport-object

saghul:
> Regardless, I don't think this PR will cause much issues in future,

I respectfully disagree :-) In the future libWebRTC will remove legacy stat generation, so we'll probably have to undo your change, and make sure we generate them The Right Way on the native side of the plugin.

> Chromium Nightly 64 (and few older versions, possibly even stable down to 58) return the getStats() results in the format that this PR provides (and this format corresponds to current version of the WebRTC spec), so you have to do this at some point, better sooner than later.

I'm aware of this.

I'd like to see this properly fixed on the native side. By "properly" I mean in a future-proof manner. As it stands, this PR is just a stopgap, which doesn't fix the underlying issue, that is, (at least on iOS) we are generating "old fashioned" stats reports.

nazar-pc:
From my perspective what matters is to get spec-compliant result. As I don't know native languages in use, I've implemented workaround in JavaScript, but if you can fix underlying issue, that would be even better!

Either way, you will need to convert it to Map-like object from an Array, so at least part of this PR is still useful. I think this can be merged as workaround now to get spec-compliant result immediately and later, when underlying structure is fixed, `for` loop body can be reduced to a single line `specStats.set(item.id, item)`.

ibc:
@nazar-pc, I understand your aim with this PR, but it does not make much sense that the whole `react-native-webrtc` expects and produces legacy WebRTC API (Chrome ~2014) and, at the same time, we mangle just a few JS Objects to match the latest WebRTC spec.

The same would apply to the `getUserMedia` constraints: there is no `mandatory`/`optional` arrays anymore in the latest spec, but `react-native-webrtc` relies on them because that's how the underlying libwebrtc API (at least the ObjC wrapper) works.

saghul:
@nazar-pc I understand your angle, but as someone who is on the other side of the fence (i.e. the maintainer side) this is just more tech debt and doesn't fix the real problem, I'm sorry.

@ibc agreed!

nazar-pc:
Got it. Can we expect fundamental changes soon? Simple-peer doesn't work with `react-native-webrtc` that:
1) Doesn't follow the spec
2) Doesn't mimic behavior of any reasonably recent major browser

ibc:
You could ask Google to implement the new API in the libwebrtc, specially in the ObjC wrapper. That would help. Otherwise this is about doing lot of JS magic to map new API syntax to the old one (the one that libwebrtc implements) and vice-versa.

nazar-pc:
That would be a bit too much for me. I'm not even using `react-native-webrtc` myself, just got involved because this issue appeared to be related to the changes I've done in other project and I had to dig a few levels deeper.

It would be sad to see workaround rejected and proper fix not present. At the end of the day, you can always revert commit when it is not needed anymore.

nazar-pc:
With #449 merged I think this is not needed anymore

PR #436, opened 2018-04-06
Android: Fallback to any camera when constraints cannot be met.

guusdk:
When creating a videocapturer, allow one to be created that does not meet the provided constraints, if no camera matches these constraints. This mimics the IOS behavior.

Tread carefully: this is an (untested) backport of the (tested) fix in jitsi#24


saghul:
Closing since we are going to merge back soon.

Issue #338, opened 2017-08-27
Muting remote stream


konstantin-mohin:
Hi, 

I'm trying to mute stream when user gets phone call, 

`peer.getRemoteStreams()[0].getAudioTracks()[0].enabled = false`

but still hear the stream.

any ideas ?

----------------------------------
Also want to ask one question ,
I've tried to update webrtc to the last version and it works well,
except for one fact,
I get iceConnectionState = field almost immediately if it lost the network connection and I need to reconnect, 

while using the current version of webRTC,
I'm getting this state after 15 sec,
it means,
if I find the network before I get iceConnectionState = false,
I stay connected with another peer and that certainly pleases me,

 is this behaviour is intended ?

 

#### Platform information
Android 7,
RN 0.43.4,
RN webRTC 1.54.7

papigers:
@konstantin-mohin Any luck?

konstantin-mohin:
@papigers unfortunately no, maybe in a newer version of RNwebRTC it's working.

ippa:
@papigers @konstantin-mohin did you ever manage to solve this?

I have the very same problem, I can't mute audio on a videocall. I've tried toggling 'enabled' and 'muted' properties on both local and remote stream tracks.

I've also tried using the react-native-incall-manager plugin but they explicity say 'setMicrophoneMute()' won't work on iOS.. they even refer to the above method in webrtc, toggling enabled.

@oney do you know of any workarounds to this issue at this time?

emargollo:
Hi,

I've been trying to figure this out as well, I'm currently setting the enabled attribute to false, but it doesn't seem to work.

I've tried logging the tracks after setting enabled to false and I can see the attributes are being set correctly, _enabled is false and muted is true, yet I still listen to the remote audio.

#### Platform information

* **React Native version**: 0.55.2
* **Plugin version**: 1.67.1
* **OS**: Android
* **OS version**: 8

felixSchl:
You might want to put the peer on hold. Afaik this requires some SDP munging by creating an offer with `a=sendonly` on your end, which the peer will respond to with an `a=recvonly` or `a=inactive`. See more info in https://github.com/w3c/webrtc-pc/issues/142. You should still be able to just mute incoming audio though without going through any of this.

danjenkins:
I'm having this exact same issue. Muting a local stream works fine - muting a remote stream doesnt.

I can see its setting all the right properties and calling the underlying native function but I still hear remote audio.

[...]

Issue #179, opened 2016-12-15
Muting incoming audio doesn't work - 0.8.0


szymonzmyslony:
As I understand this code should mute incoming audio:
```
var config = this.props.mixertest.webrtcStuff
      if(config.remoteStream){
             var pc = config.pc
             var streams = pc._remoteStreams
        streams[0].getTracks().forEach(track => track.enabled = !track.enabled)
      }}}
```

It does change track.enabled value but audio doesn't get muted. What did I miss? Is there other way to mute incoming audio?

zxcpoiu:
yes, you are right.
but I only test on local tracks not sure about if we can mute the remote stream that way.

it's obvious there are only 3 ways:

1. find a way to mute remote track locally ( not sure about it )
2. turn off your speaker for that stream use system api. ( but webrtc controls audio player itself, so it may have unpredictable behaviors )
3. asking remote party to mute their local track remotely, just like "hold" in terms of telephony.

see how official webrtc app does is always a good direction to understand it's ability

szymonzmyslony:
so in your opinion it's impossible to do it locally?

zxcpoiu:
everything is possible :)

I just didn't find any implementations does that yet.

my guess is:

1. if you can get `AudioPlayer` instance which created by `webrtc`, you might able to use `AudioPlayer` API to *pause* it.

2. by webrtc documents: https://www.w3.org/TR/webrtc/#hold-functionality
    it shows how to implement "hold" functionality, but it seems like the same as my previous comments point 3. 


This module did not implement `RTCRtpTransceiver`, `RTCRtpReceiver`, `RTCRtpSender` yet.
and I believe those are interfaces which control local/remote media streams' rendering.

atyenoria:
@szymonzmyslony 
I tried streams[0].getTracks().forEach(track => track.enabled = !track.enabled) . But nothing changed. Have you succeeded in muting remote audio??

szymonzmyslony:
no, I dont know how to it yet. Any comments would be appreciated.

ippa:
@szymonzmyslony @atyenoria  did you ever manage to find a workaround?

@zxcpoiu you seemed knowledgeable with the details :) .. do you know a good work-around to this issue?

Toggling enabled (and muted)-properties on local and remote stream audio tracks won't work.

danjenkins:
Pretty sure this has been fixed now. Closing and  if its still an issue for someone can you make a new issue?

PR #440, opened 2018-04-16
muting/unmuting remote streams

weberjc:
updates mediaStreamTrackSetEnabled to work for remote tracks as well as by local tracks.

so setting .enabled on a remote track in js will mute / unmute the track if it is audio.
applies to issues #338 and #179

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -535,7 +535,7 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
 
     @ReactMethod
     public void mediaStreamTrackSetEnabled(final String id, final boolean enabled) {
-        MediaStreamTrack track = localTracks.get(id);
+        MediaStreamTrack track = getTrackForId(id);
         if (track == null) {
             Log.d(TAG, "mediaStreamTrackSetEnabled() track is null");
             return;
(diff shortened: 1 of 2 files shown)


weberjc:
@ippa yes this solution worked on 1.58.3 for muting remote audio tracks (local tracks already worked).  I updated to fix the merge conflict in 1.63.0.
With this change just call find the audio track in javascript (e.g. track = stream.getAudioTracks()[0]) and set track.enabled = false or track.enabled = true.
It's not tested with video tracks however.  I did see that in the jitsi fork there's a similar commit that  can mute video tracks too.  https://github.com/jitsi/react-native-webrtc/commit/49f678eed157ba95cbd7ec8bc19af1b002bb1513
Was wondering if perhaps that or the effort for M63 might have been keeping this from being merged.

saghul:
Sorry, this has fallen through the cracks again :-/ Can you please rebase? I'll include it for the 69 release.

weberjc:
Fixed the conflict, should be good to go now.

weberjc:
Actually it needs one fix first

saghul:
Hey @weberjc, seems like the iOS part is missing now?

weberjc:
@saghul it looks like the ios part is still there, however i'm not in a position to test it at the moment.  It gets those remoteTrack vars from the other files like WebRTCModule+RTCPeerConnection.m.

saghul:
I see it again now. I'll give it a test and merge it.

akashkrshukla:
Is there a reason why this PR is closed or some alternative solution has been added to library itself?

PR #445, opened 2018-04-19
Implement remaining RTCConfiguration options for iOS

bradleyjokinen:


shaneneuerburg:
Can this get merged? It would be very useful to have

zxcpoiu:
Will do! but would merge M67 first as planned previously. seems it's almost set.

saghul:
I'd like to have an equivalent fix for Android. having each platform behave differently is not really A Good Thing.

guitarultimate:
While not all, a great deal of this functionality appears to be already implemented in Android. I stumbled upon this pull request because my Android implementation was working and my iOS was not due to issues that required a change in bundle policy.  Applying this code fixes the issue for me with no regressions I noticed.

While I agree that platform discrepancies aren't good, since there already is one could we apply this since it ultimately gives us less broken functionality?

saghul:
I plan to get to this one soon, sorry for the delay everyone.

saghul:
Landed in 664440bb9e93c350d81ef60fd06156f1a86242ee, cheers!

PR #446, opened 2018-04-19
Fix MediaStreamTrack.getSources() result format on iOS

bradleyjokinen:
This fixes an issue with MediaStreamTrack.getSources() not resulting in the expected structure on iOS due to an extra layer of nesting.


bradleyjokinen:
It seems that this is no longer required as of 1.63.0.

PR #462, opened 2018-05-30
Update iOSInstallation.md

david0u0:
The original content of line 50
4-2.) add `libRCTWebRTC.a` 
is incorrect, and will cause some linker error. Just skip it and everything is alright.

### DIFF
--- a/Documentation/iOSInstallation.md
+++ b/Documentation/iOSInstallation.md
@@ -47,9 +47,8 @@ pod 'react-native-webrtc', :path => '../node_modules/react-native-webrtc'
 
 
 4-1.) click `Build Phases` tab, open `Link Binary With Libraries`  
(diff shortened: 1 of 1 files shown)


saghul:
Because it's incorrect. Use CocoaPods and you won't need to fiddle with Xcode linking stuff.

shirow-ozawa:
Then why isn't libRCTWebRTC included in the capture?

saghul:
That's old stuff. The docs need updating, sorry.

shirow-ozawa:
my settings is below.
Link_Binary_With_Libraries

Where should I fix it?

saghul:
Use CocoaPods, don't manually link. Ignore those instructions completely. Do as you would with any native plugin which uses CocoaPods.

PR #468, opened 2018-06-15
Added optional argument "constraints" to RTCPeerConnection constructor

vyshkant:
Currently the user cannot pass the needed "constraints" value to RTCPeerConnection when it is initializing.

This PR allows us to do that.

This PR does not break BC.


saghul:
Constraints have been dropped from the standard.

PR #485, opened 2018-07-25
Added Android Build Support for newer React Native Versions

cristiantx:
Added Android Build Support for newer React Native Versions which uses a new SDK and Build Tool version.


zxcpoiu:
moved to #494
thanks @cristiantx

PR #494, opened 2018-08-08
Fix Android build for RN56

maxhawkins:
Solution copy-pasted from here: https://github.com/crazycodeboy/react-native-splash-screen/pull/246/files


zxcpoiu:
This seems more agile to support future changes.

@cristiantx 

What do you think?
Is there any down side to use a function?

PR #507, opened 2018-08-24
Add bundlePolicy to the configuration parser in iOS 

YvanGuidoin:
Hi,

This fixes the issue to connect to Pexip MCU by setting the bundlePolicy: "balanced" directly from Javascript.

The RFC https://www.w3.org/TR/webrtc/#rtcbundlepolicy-enum for the enum.
By default in the RTCConfiguration it's supposed to be at "balanced" by default but I guess it would break the behavior for other people so I just left it as it is now by default (I guess the WebRTC Framework take care of it by default).

I'm not an Objective-C dev so tell me if I have done anything wrong. Tested on an Iphone and it works correctly.


saghul:
Already covered in 664440bb9e93c350d81ef60fd06156f1a86242ee

PR #508, opened 2018-08-31
Updated Android Instructions for M67

cristiantx:
Added steps to update the graddle and build tools to support Java 8.

### DIFF
--- a/Documentation/AndroidInstallation.md
+++ b/Documentation/AndroidInstallation.md
@@ -58,7 +58,7 @@ import com.oney.WebRTCModule.WebRTCModulePackage;  // <--- Add this line
 
 5.) Enable Java 8 support in your project. You will probably need to have React Native 0.55+ for this.
 
-5.a.) In `anroid/app/build.gradle` add:
+5.a.) In `android/app/build.gradle` add:
 ```gradle
 compileOptions {
     sourceCompatibility JavaVersion.VERSION_1_8
--- a/Documentation/AndroidInstallation.md
+++ b/Documentation/AndroidInstallation.md
@@ -56,14 +56,38 @@ import com.oney.WebRTCModule.WebRTCModulePackage;  // <--- Add this line
      .setUseDeveloperSupport(BuildConfig.DEBUG)
 ```
 
-5.) Enable Java 8 support in your project. In `anroid/app/build.gradle` add:
+5.) Enable Java 8 support in your project. You will probably need to have React Native 0.55+ for this.
 
+5.a.) In `anroid/app/build.gradle` add:
 ```gradle
 compileOptions {
     sourceCompatibility JavaVersion.VERSION_1_8
     targetCompatibility JavaVersion.VERSION_1_8
 }
 ```
+
+Replace to:
+
+```gradle
+    compileSdkVersion 27
+    buildToolsVersion '27.0.3'
+```
+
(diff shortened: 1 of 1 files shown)


cristiantx:
@saghul changed to minimal required gradle versions. Thanks!

saghul:
@cristiantx Sweet! Can you confirm it works now if you start from scratch?

cristiantx:
@saghul I just checked and it does work ok. But just noticed that instead of changing `compileSdkVersion` and `buildToolsVersion` in `android/app/build.gradle` we could just update the variables inside `ext` in `android/build.gradle` what do you think?

Documentation/AndroidInstallation.md
```diff
@@ -56,15 +56,39 @@ import com.oney.WebRTCModule.WebRTCModulePackage;  // <--- Add this line
      .setUseDeveloperSupport(BuildConfig.DEBUG)
 ```
 
-5.) Enable Java 8 support in your project. In `anroid/app/build.gradle` add:
+5.) Enable Java 8 support in your project. You will probably need to have React Native 0.55+ for this.
 
+5.a.) In `anroid/app/build.gradle` add:
```
k0pernicus: Typo: `anroid` instead of `android` :)

Documentation/AndroidInstallation.md
```diff
@@ -56,15 +56,39 @@ import com.oney.WebRTCModule.WebRTCModulePackage;  // <--- Add this line
...
 ```gradle
 compileOptions {
```
k0pernicus: Apparently, we need to add this configuration option in a section - otherwise, add them in android/app/build.gradle results in:

Could not find method compileOptions()

Those configuration options must be in the android part, right?
Just after defaultConfig?

cristiantx: @k0pernicus That's right it should be inside `android` section. Will add that clarification

k0pernicus: Thanks 👍

saghul:
> we could just update the variables inside ext in android/build.gradle what do you think?

👍

saghul:
@cristiantx Yeah, please make the change. This will mean less instructions then :-)

cristiantx:
All done I think!

Documentation/AndroidInstallation.md
```diff
@@ -66,21 +66,23 @@ compileOptions {
...
 
-Replace to:
-
-```gradle
-    compileSdkVersion 27
-    buildToolsVersion '27.0.3'
-```
-
-Also inside of `dependencies` replace all the `compile` by `implementation` otherwise you'll get deprecation warnings.
+Inside of `dependencies` replace all the `compile` by `implementation` otherwise you'll get deprecation warnings.
```
saghul: This is not needed (yet) and has really nothing to do with how this plugin functions. Can you please drop it?

saghul:
Nice! Left one last comment, otherwise it's ready! Thank you!

k0pernicus:
I confirm those build steps helped me in building the Android app :+1: 
Thanks again @cristiantx :)

Documentation/AndroidInstallation.md
cristiantx: Sure!

cristiantx:
@saghul Ready to go!

saghul:
@cristiantx Landed, cheers!

PR #516, opened 2018-09-22
`const i` cannot be modified inside the loop

divyanshunumrah:
changed the Readme doc error.


saghul:
Thanks for catching it! But please, use `let`, not `var` it's 2018 :-)

PR #528, opened 2018-10-17
doc: remove demos from README

saghul:
They are not up to date and are source of useless bug reports.


saghul:
@zxcpoiu Please take a look.

Issue #401, opened 2018-01-05
New remote tracks with same stream.id (in a=msid) are ignored in iOS (known problem)


ibc:
I have a react-native-webrtc `peerconnection` and, initially, call `pc.setRemoteDescription(offer)`  with an offer as follows:

```
v=0
o=mediasoup-client 56448044 1 IN IP4 0.0.0.0
s=-
t=0 0
a=ice-lite
a=fingerprint:sha-512 2B:19:7E:44:2A:92:E8:49:EE:FA:8E:C5:D3:EC:79:A3:E3:22:A3:E8:B3:C7:7C:9D:A5:6A:F3:F7:B3:BA:50:AF:58:1B:36:96:05:67:D8:6A:EE:F6:89:09:01:E4:98:6B:10:03:75:AD:14:7F:36:A6:6E:FD:9A:49:0A:DB:DF:BC
a=msid-semantic: WMS *
a=group:BUNDLE audio
m=audio 7 RTP/SAVPF 100
c=IN IP4 127.0.0.1
a=rtpmap:100 opus/48000/2
a=fmtp:100 useinbandfec=1
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=setup:actpass
a=mid:audio
a=sendonly
a=ice-ufrag:zzacfqcdz53izo9w
a=ice-pwd:8w3tkz1u8i0rg9ul9xwaxgavp3xc5mve
a=candidate:udpcandidate 1 udp 1082702079 192.168.1.92 40159 typ host
a=end-of-candidates
a=ice-options:renomination
a=ssrc:61428989 msid:recv-stream-61369657 consumer-audio-93108211
a=ssrc:61428989 mslabel:recv-stream-61369657
a=ssrc:61428989 label:consumer-audio-93108211
a=ssrc:61428989 cname:aJ9EdjJjV0S/hP05
a=rtcp-mux
a=rtcp-rsize
```

As expected, once applied the `pc` has a single remote `stream` (pc.getRemoteStreams()[0])  with id `recv-stream-61369657`. And such a `stream` has a single audio track with id `consumer-audio-93108211`.

After that, I run `pc.createAnswer()` and `pc.setLocalDescription(answer)`.

Later, I call `pc.setRemoteDescription(offer)` with a new offer as follows:

```
v=0
o=mediasoup-client 56448044 2 IN IP4 0.0.0.0
s=-
t=0 0
a=ice-lite
[...]

PR #564, opened 2019-01-07
allow tracks to be added and removed during renegotiation

gmondada:
During WebRTC renegotiation, some tracks can appear or disappear. Current implementation doesn't update the list of tracks stored in the javascript environment and, after renegotiation, getRemoteStreams() returns the wrong tracks.

### DIFF
--- a/RTCPeerConnection.js
+++ b/RTCPeerConnection.js
@@ -361,6 +361,32 @@ export default class RTCPeerConnection extends EventTarget(PEER_CONNECTION_EVENT
         // checking for ResourceInUse.
         this._dataChannelIds.add(id);
         this.dispatchEvent(new RTCDataChannelEvent('datachannel', {channel}));
+      }),
+      DeviceEventEmitter.addListener('peerConnectionAddTrack', ev => {
+        let i;
+        if (ev.id !== this._peerConnectionId) {
+          return;
+        }
+        for (i = 0; i < ev.streams.length; i++) {
+          const streamRef = ev.streams[i];
+          const stream = this._remoteStreams.find(s => s.reactTag === streamRef.reactTag);
+          if (stream) {
+            stream.addTrack(new MediaStreamTrack(ev.track));
+          }
+        }
+      }),
+      DeviceEventEmitter.addListener('peerConnectionRemoveTrack', ev => {
+        let i;
+        if (ev.id !== this._peerConnectionId) {
+          return;
+        }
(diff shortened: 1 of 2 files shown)


saghul:
Thanks for doing this? I assume an Android equivalent is missing?

gmondada:
Correct, I didn't implement the Android part, sorry.

saghul:
Are you able to? We try to minimize the differences between platforms so nobody is surprised.

gmondada:
I understand you want to minimize the differences between platforms, but I'm not able to do that on Android, sorry.

ibc:
So, does it solve the problem I reported in #401? To clarify, #401 is a very old and well known issue in ObjcC bindings of libwebrtc. The problem was that when a new remote track was added to a remote stream (via SDP renegotiation), the ObjC `MediaStream` did not realize of the new track and did not emit the "ontrack" event.

However, the new approach in WebRTC 1.0 is the PeerConnection "ontrack" event, which happens when a new remote track is generated after a O/A renegotiation. I assume that it's also implemented in modern libwebrtc ObjC bindings. Otherwise your PR would not work, am I right?

Also, may you clarify if your PR would make this work in iOS?

* Call `pc.setRemoteDescription()` with a SDP containing a single `m=audio` section with `a=msid: MMMM TTT1`.
* So `pc.getRemoteStreams()` returns an array with a single `MediaStream`.
* `stream.id` is "MMMM".
* `stream.getTrackById("TTT1")` obtains a remote audio `MediaStreamTrack` with `id` "TTT1".
* Later renegotiation is done and the remote SDP includes a new `m=video` section with `a=msid: MMMM TTT2`.
* So `pc.getRemoteStreams()` still returns an array with a single `MediaStream`.
* `stream.id` is "MMMM" (of course).
* `stream.getTrackById("TTT1")` obtains the remote audio `MediaStreamTrack` with `id` "TTT1".
* `stream.getTrackById("TTT2")` obtains the remote video `MediaStreamTrack` with `id` "TTT2".

Is it accurate?

gmondada:
I think you are right. My intention was not to solve or workaround any problem in the objc bindings. The objc bindings does notify new tracks added during renegotiation (I don't remember through which event it does this). This PR just forwards the notifications to the RN/JS side in order to update the MediaStream object.
The process described by your bullet list should work.

saghul:
Closing since this will no longer apply in unified plan.

PR #565, opened 2019-01-07
allow using custom codec factories

gmondada:


saghul:
Landed with fixes in 7c32ebea35062b33b88285eb1a850774d48b72a8. It didn't compile.

PR #566, opened 2019-01-08
Use ES6 export

piranna:
All files are using ES6 imports and exports except this one that using CommonJS `module.exports`, and in fact is mixing it with ES6 imports. This pull-request fix it having all files using ES6 imports and exports.


brunsy:
@saghul This can introduce in breaking changes to how the project is imported...
You might want to keep a default export here too.

Issue #575, opened 2019-01-28
Could not find @ReactModule annotation in com.oney.WebRTCModule.WebRTCModule


mbloody:
![screenshot](https://user-images.githubusercontent.com/1745865/51846116-a293d100-2321-11e9-89fd-09d016851133.png)

#### Platform information

* **React Native version**: 0.58.1
* **Plugin version**: 1.67.1
* **OS**:  Android
* **OS version**: all

mbloody:
On real device:

01-28 17:53:55.856 25351 25351 W ResourceType: No known package when getting name for resource number 0xffffffff
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: Error while updating prop streamURL
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: java.lang.reflect.InvocationTargetException
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: 	at java.lang.reflect.Method.invoke(Native Method)
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: 	at com.facebook.react.uimanager.ViewManagersPropertyCache$PropSetter.updateViewProp(ViewManagersPropertyCache.java:83)
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: 	at com.facebook.react.uimanager.ViewManagerPropertyUpdater$FallbackViewManagerSetter.setProperty(ViewManagerPropertyUpdater.java:132)
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: 	at com.facebook.react.uimanager.ViewManagerPropertyUpdater.updateProps(ViewManagerPropertyUpdater.java:51)
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: 	at com.facebook.react.uimanager.ViewManager.updateProperties(ViewManager.java:34)
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: 	at com.facebook.react.uimanager.NativeViewHierarchyManager.createView(NativeViewHierarchyManager.java:269)
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: 	at com.facebook.react.uimanager.UIViewOperationQueue$CreateViewOperation.execute(UIViewOperationQueue.java:200)
01-28 17:53:55.860 25351 25351 E unknown:ViewManager: 	at com.facebook.react.uimanager.UIViewOperationQueue$1.run(UIViewOperationQueue.java:911)
[...]

PR #576, opened 2019-01-28
Fix for Could not find @ReactModule annotation issue #575

mbloody:


saghul:
Landed in 8def7d0dc9370934a4d56770d63d742c2a203fea, cheers!

PR #602, opened 2019-03-15
android: add support for 64bit builds 

saghul:
The supported Android architectures are now: arm, arm64, x86 and x86-64.


saghul:
Those interested in 64bits please give this a go! Preliminary testing shows it's working just fine.

Aubynj:
MediaStreamTrack.getSources crashes on android with **React Native version 0.58.6** Error log shows that MediaStreamTrack.getSources is not a function. Is that am doing something wrong? I have uninstall the lib and reinstall it back but the problem persist

brunsy:
@Aubynj That is not related to this PR (have a look at https://github.com/oney/react-native-webrtc/issues/581)

@saghul Thanks, been using this for a while now and seems to running fine.

PR #619, opened 2019-04-10
Change code sample.

DimitrisTzimikas:
This is a proposal for changes in the code of sample. I know its the whole thing but I think its more useful this way. 
If found that the existing code has some errors like in if statement, there is no ```sourceInfo.kind == "video"``` but ```sourceInfo.kind === 'videoinput'```, and some others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,53 +32,30 @@ Please see [wiki page](https://github.com/react-native-webrtc/react-native-webrt
 Now, you can use WebRTC like in browser.
 In your `index.ios.js`/`index.android.js`, you can require WebRTC to import RTCPeerConnection, RTCSessionDescription, etc.
 
-```javascript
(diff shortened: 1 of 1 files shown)


saghul:
Looks good! How about you turn this into a sample application in a new `example/` directory instead?

DimitrisTzimikas:
Yes of course, but do you mean to create the directory in main repo, like in the same level with 

- Documentation 
- andoid 
- ios 
- exapmle

etc? I can create a pull request with a working sample for android and until Friday I will do it for iOS too.

saghul:
Yes, at the top level. 

> I can create a pull request with a working sample for android and until Friday I will do it for iOS too.

No rush, I appreciate you wanting to do this!



> After that do you think its a good idea to change the code in README.md (same format as you have it) but with the update API?

After that I'd say we simply remove the code and point at the example.

PR #623, opened 2019-04-11
Add example 

DimitrisTzimikas:
This is a working example that uses the current API. The only thing it does is to configure the devices media and display the from camera (or rear) stream to the RTCView.  It also give access to RTCView props "change camera', 'mirror' and 'object fit' option. 

Its configured only for Android. 

EDIT: 
I configured the iOS and build successfully but only on emulator. I'll test it atMonday on a real device. Maybe it would be better to wait until then to pull then with working configuration for both iOS/Android.


DimitrisTzimikas:
Final example, working on Android & iOS

example/rnWebRTC/android/build.gradle
```diff
@@ -0,0 +1,34 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
+buildscript {
+    ext {
+        buildToolsVersion = "28.0.3"
+        minSdkVersion = 16
```
aghahuseynov: really ?

aghahuseynov: I think the following versions of Android 24 are not supported or I couldn't run

saghul: It must be something else. We use 19 with no problems. 16 is indeed too old, Google no longer supports it.

piranna: 16 is the minimum version needed by ReactNative itself, newer versions would be needed by specific native modules if they use more recent APIs, but they would need to specify it in their docs.

aghahuseynov: I get this error when I make minSDK 23 in the project.

saghul: Try adding this to your gradle.properties file: android.enableDexingArtifactTransform.desugaring=false

aghahuseynov: I am trying...

aghahuseynov: YES it is worked! Thanks for support. Can you write about the "code" ?

saghul: What "code" ?

aghahuseynov: android.enableDexingArtifactTransform.desugaring=false

saghul: What is the question? This disables some ccode transformations that seem to break WebRTC. It started to appear with the gradle plugin 3.5.0.

aghahuseynov: Thanks

PR #624, opened 2019-04-21
Provide custom capturing capability

keshavkaul:
Refactor code base to support adding custom video capturing capability.
At the moment only iOS is supported. Will add support for Android later.


saghul:
Loos good! There are quite a few indentation and extraneous whitespace issues. Can you fix those please?

keshavkaul:
@saghul Done. I wonder is there any easy way to sync text format across team in Xcode. (editorconfig)[https://editorconfig.org] is one way but xcode is not supported.

PR #628, opened 2019-05-06
Add support for mediasoup

Sharcoux:
Signed-off-by: François Billioud <fbillioud@cutii.io>

The track needs to keep track of the stream that generated it, otherwise you will never be able to display it in an RTCView.

### DIFF
--- a/MediaStream.js
+++ b/MediaStream.js
@@ -47,6 +47,7 @@ export default class MediaStream extends EventTarget(MEDIA_STREAM_EVENTS) {
   }
 
   addTrack(track: MediaStreamTrack) {
+    if(track.kind==='video' && this.getVideoTracks().length===0 && track.streamReactTag) this.reactTag = track.streamReactTag;
     this._tracks.push(track);
     this.dispatchEvent(new MediaStreamTrackEvent('addtrack', {track}));
   }
@@ -59,6 +60,7 @@ export default class MediaStream extends EventTarget(MEDIA_STREAM_EVENTS) {
     WebRTCModule.mediaStreamTrackRelease(this.reactTag, track.id);
     this._tracks.splice(index, 1);
     this.dispatchEvent(new MediaStreamTrackEvent('removetrack', {track}));
+    if(this.getVideoTracks().length===0) this.reactTag = undefined;
   }
 
   getTracks(): Array<MediaStreamTrack> {
--- a/getUserMedia.js
+++ b/getUserMedia.js
@@ -141,7 +141,9 @@ export default function getUserMedia(constraints = {}) {
         /* successCallback */ (id, tracks) => {
           const stream = new MediaStream(id);
           for (const track of tracks) {
-            stream.addTrack(new MediaStreamTrack(track));
+            const mediaStreamTrack = new MediaStreamTrack(track);
+            mediaStreamTrack.streamReactTag = stream.reactTag;
+            stream.addTrack(mediaStreamTrack);
           }
     
           resolve(stream);


Sharcoux:
It could be a good idea to change MediaStream for it to read the streamReactTag of the tracks we add to it. Or at least find a solution for users to use `new MediaStream()`, add tracks to it, and still be able to display it inside an `RTCView`. You see what I mean?

saghul:
I'm not sure how this fixes anything, since we only read `streamReactTag` from a stream, IIRC.

> The track needs to keep track of the stream that generated it, otherwise you will never be able to display it in an RTCView.

This is working now without this change, so is certainly not needed. Can you please clarify?

> Or at least find a solution for users to use `new MediaStream()`, add tracks to it, and still be able to display it inside an `RTCView`

Yes, we definitely need to fix that. I'll have time in the upcoming months for it. Alas is not a simple change, because creating the stream on the native side happens asynchronously.

Sharcoux:
As long as the tracks keep the reference to the stream that generated it, we can always find a way to display it. If we don't, MediaStreamTracks are completely useless and so is using `MediaStream` to represent them.

> This is working now without this change, so is certainly not needed. Can you please clarify?

First of all, it won't work with mediasoup. But ok, you may say that it is not the problem of react-native-webrtc to handle that.

We should probably also update the MediaStream class to do something like:

```javascript
  addTrack(track: MediaStreamTrack) {
+    if(this.kind==='video' && this._tracks.length===0) this.reactTag = track.streamReactTag;
    this._tracks.push(track);
    this.dispatchEvent(new MediaStreamTrackEvent('addtrack', {track}));
  }
```

By doing this, we ensure that `a = new MediaStream();a.addTrack(videoTrack);` will always be rendered by an `RTCView`.

ibc:
> I'm not sure how this fixes anything, since we only read `streamReactTag` from a stream, IIRC.

It makes it possible for the application to know the `reactTag` of the stream (which is what react-native-webrtc unfortunately uses internally to reference a single video track), so the application can later construct a new `MediaStream` from scratch and assign the same `reactTag` to it (it can read it from the video track it's gonna append to the that new stream) and things "will work".

Yes. It's an ugly solution for an ugly design.

> > The track needs to keep track of the stream that generated it, otherwise you will never be able to display it in an RTCView.
> 
> This is working now without this change, so is certainly not needed. Can you please clarify?

The native `MediaStream` just assumes that the video it must render is referenced by the annoying `reactTag` value. If a new created `MediaStream` shares the same `reactTag`, both streams will represent/render the same video. That's all.

 
> > Or at least find a solution for users to use `new MediaStream()`, add tracks to it, and still be able to display it inside an `RTCView`
> 
> Yes, we definitely need to fix that. I'll have time in the upcoming months for it. Alas is not a simple change, because creating the stream on the native side happens asynchronously.

This lib could create the JS `MediaStream` synchronously in JS land and asynchronously in native land (similar to how `cordova-plugin-iosrtc` creates a [PeerConnection](https://github.com/BasqueVoIPMafia/cordova-plugin-iosrtc/blob/master/js/RTCPeerConnection.js#L30)). Not a big deal IMHO.

saghul:
> By doing this, we ensure that `a = new MediaStream();a.addTrack(videoTrack);` will always be rendered by an `RTCView`.

I see what you mean now. The proposed fix, is not addecuate though. While that works for just displaying a local stream, the native side won't have been told about it, so it won't be present in the local description if you add the newly created stream to a peer connection.

Sharcoux:
It is true that if you create your own `MediaStreamTrack` outside of `getUserMedia`, there is no way right now to stream it. However, it doesn't mean that we should not make us able to manipulate MediaStream containing valid MediaStreamTracks for now on.

ibc:
> It is true that if you create your own `MediaStreamTrack` outside of `getUserMedia` and not coming from the `peerConnection`, there is no way right now to stream it. However, it doesn't mean that we should not make us able to manipulate MediaStream containing valid MediaStreamTracks for now on.

Yes. This PR is not the big refactor this lib needs but just a workaround that enables a single and legit use case. Since such a required refactor may take long, accepting this PR would at least improve the lib a bit.

saghul:
I disagree. Right now we have bad, but at least consistent behavior. Creating streams and manually adding tracks to them doesn't work, period. If we land this PR iw will work sometimes, which is harder to reason about.

If you want to take a stab at fixing this the right way, I'm happy to help.

ibc:
> I disagree. Right now we have bad, but at least consistent behavior. Creating streams and manually adding tracks to them doesn't work, period. If we land this PR iw will work sometimes, which is harder to reason about.
> 
> If you want to take a stab at fixing this the right way, I'm happy to help.

That would mean refactoring almost everything and fixing the "1 stream == 1 video track" assumption. Or we can just stay with this assumption and the WebRTC API from early 2013 for next 5 years.

saghul:
> That would mean refactoring almost everything and fixing the "1 stream == 1 video track" assumption.

Absolutely not.

ibc:
> > That would mean refactoring almost everything and fixing the "1 stream == 1 video track" assumption.
> 
> Absolutely not.

You know the internals better than me, but it's easy to verify that this lib assumes 1 stream == 1 video track. In fact, internally it does not handle tracks but just streams.

Sharcoux:
> I disagree. Right now we have bad, but at least consistent behavior. Creating streams and manually adding tracks to them doesn't work, period. If we land this PR iw will work sometimes, which is harder to reason about.

I don't get your position. It would work for all tracks generated by react-native-webrtc. If you are worried about people creating tracks outside of it, why not throwing an error when they attempt to do so?

The issue right now is that when a peer shares a track, I receive a notification about that track. Not the stream. So what am I supposed to do with that track if I have no MediaStream where I can legitimately store it?

The only solution I found was to monkey patch the MediaStream the way I described above.

ibc:
> > I disagree. Right now we have bad, but at least consistent behavior. Creating streams and manually adding tracks to them doesn't work, period. If we land this PR iw will work sometimes, which is harder to reason about.

@saghul: the only tracks in react-native-webrtc are those generated by react-native-webrtc itself. Now imagine this:

* react-native-webrtc creates a stream (via gUM or `pc.addstream` event).
* react-native-webrtc checks if there is a video track on it.
* If so, it generates a `reactTag` and assigns it to the video track.
* And of course the track exists in the stream, so the stream knows its "video reactTag".
* Later the app creates a new `stream2` and does `stream2.addTrack(videoTrack)`.
* `stream2` verifies that the given `videoTrack` has a `reactTag` and adopts it.

In summary: the `reactTag` that a MediaStream reports to the native code is the `reactTag` of its video track (if any). Problem solved.

Sharcoux:
And you should probably throw a warning or an error if people attempt to add a second video track to the same stream as there is no way this could be represented anyway (that's why @ibc was saying 1 Stream === 1 video Track).

saghul:
I know how the code works.

What you seem to not like to hear is what I said in https://github.com/react-native-webrtc/react-native-webrtc/pull/628#issuecomment-490069859

If you create a stream by hand and add tracks from another stream coming from gUM, and then add it to a PC, there will be no tracks in the local description.

That is a problem and this is not the right fix for it.

ibc:
> If you create a stream by hand and add tracks from another stream coming from gUM, and then add it to a PC, there will be no tracks in the local description.

If both streams share the same reactTag wouldn't it work?

saghul:
> If both streams share the same reactTag wouldn't it work?

I'm not sure, but what would that accomplish? addTrack and removeTrack don't operate on the native side so you wouldn't be modifying the underlying stream anyway.

ibc:
> > If both streams share the same reactTag wouldn't it work?
> 
> I'm not sure, but what would that accomplish? addTrack and removeTrack don't operate on the native side so you wouldn't be modifying the underlying stream anyway.

Yes, `addTrack` and `remoteTrack` does nothing. The only relationship/mapping between JS and native is the `stream.reactTag` which is transmitted from JS to native and it's used in native to find the associated video track, right?

If so, I insist that the solution/workaround above would work even in your suggested scenario. to summarize:

1. Call `getUserMedia` => get stream1 with `reactTag: 1234`.
1. Let's assume that, as proposed above, the video track in stream1 also has a `streamReactTag: 1234` attribute.
1. `const stream2 = new MediaStream();` => stream2 has no `reactTag`.
1. `stream2.addTrack(stream1.getVideoTracks()[0])` => stream2 copies `track.streamReactTag` into its `reactTag` attribute.
1. Now both stream1 and stream2 are two separate streams (in fact stream2 just exists in JS) that point to the same native stream + single video track.
1. So you can use **any** of them, stream1 or stream2, into a video element to render your local video track or call `pc.addStream()` with both. Same result.

Sharcoux:
The current code in this PR does make the local stream being displayed, and enable sending the local stream to remote peers. However, it is not possible to display the remote peer video. @ibc or @saghul do you know what I should add to make this work ?

If I remember correctly, I had to add the remote streams to the peerConnection or to the localStreams, but I don't get why I have to do that or how I can do it.

I succeeded to hack that easily in mediasoup-client, but I don't see how to do that in react-native-webrtc.

ibc:
I'm really sorry but I don't have time now to get involved into this neither to spend time figuring out what must be done to make it work. I just don't remember the details.

Sharcoux:
@ibc of course, sorry.
I was doing this in ReactNative handler in mediasoup-client:

`this._pc.addStream(this);`

This was made to add remote streams to the PeerConnection, in order to read the remote stream when using the RTCView. But, I don't know how to do that from react-native-webrtc as I won't have access to the PeerConnection. I think that it should probably be fixed in the ReactNative handler...

Sharcoux:
Ok, now the PR makes react-native-webrtc 100% working with mediasoup.
@saghul was the answer of @ibc clear enough? Do you think you could merge the PR?

saghul:
I'm closing this. Luckily I got some time to work on The Proper Fix (TM): support for MediaStream constructor. You can take a look and test it here: https://github.com/jitsi/react-native-webrtc/pull/55 - it may still take a bit of time to land while we test it.

ibc:
I took a look to the MediaStream constructor and it looks nice. Good work.

PR #629, opened 2019-05-14
Add optional toJSON for candidate

### LINKED ISSUE #582: Support for standard in RTCPeerConnection.addIceCandidate
Similar to https://github.com/oney/react-native-webrtc/pull/524 it probably makes sense to support null | undefined | RTCIceCandidateInit dictionary in RTCPeerConnection.addIceCandidate as some libraries are not passing RTCIceCandidate but rather RTCIceCandidateInit structure as a parameter value

#### Expected behavior
pc.addIceCandidate(null) or pc.addIceCandidate({ candidate, sdpMid, sdpMLineIndex }) with correct values should result into resolved promise

#### Observed behavior
rejected promise due to undefined candidate or candidate.toJSON in https://github.com/oney/react-native-webrtc/blob/5ea877734facf0857febe673f6cddfc6b0c3f9ec/RTCPeerConnection.js#L189

#### Steps to reproduce the problem
pass null | undefined | object conforming to RTCIceCandidateInit structure (one without toJSON method)

#### Platform information
any

* **Plugin version**:
1.69.0 and older

Also (this probably should be filed as a separate issue), at least in Android implementation  sdpMLineIndex can not be null while it's actually valid to have it as null along with sdpMid for end-of-candidates indication. This is due to the following call, which throws exception: https://github.com/oney/react-native-webrtc/blob/5ea877734facf0857febe673f6cddfc6b0c3f9ec/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java#L855

bighitbiker3:
This fixes #582 and allows for `pc.addIceCandidate(null)` or `pc.addIceCandidate({ candidate, sdpMid, sdpMLineIndex })` with correct values resulting in a resolved promise.

### DIFF
--- a/RTCPeerConnection.js
+++ b/RTCPeerConnection.js
@@ -186,7 +186,7 @@ export default class RTCPeerConnection extends EventTarget(PEER_CONNECTION_EVENT
   addIceCandidate(candidate) {
     return new Promise((resolve, reject) => {
       WebRTCModule.peerConnectionAddICECandidate(
-        candidate.toJSON(),
+        candidate.toJSON ? candidate.toJSON() : candidate,
         this._peerConnectionId,
         (successful) => {
           if (successful) {


saghul:
Thanks!

PR #637, opened 2019-05-27
Add camera position in facing property

KarlosQ:
- Add camera position in device list. Now we can get it in facing property. 
- Fix readme
- Add RTCView explication with props

### DIFF
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -120,8 +120,14 @@ RCT_EXPORT_METHOD(enumerateDevices:(RCTResponseSenderBlock)callback) {
     NSMutableArray *devices = [NSMutableArray array];
     NSArray *videoDevices = [AVCaptureDevice devicesWithMediaType:AVMediaTypeVideo];
     for (AVCaptureDevice *device in videoDevices) {
+        NSString *position = @"";
+        if (device.position == 1) {
+            position = @"back";
+        } else if (device.position == 2) {
+            position = @"front";
+        }
         [devices addObject:@{
-                             @"facing": device.position,
+                             @"facing": position,
                              @"deviceId": device.uniqueID,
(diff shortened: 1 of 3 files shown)


ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -121,6 +121,7 @@ - (RTCVideoTrack *)createVideoTrack:(NSDictionary *)constraints {
     NSArray *videoDevices = [AVCaptureDevice devicesWithMediaType:AVMediaTypeVideo];
     for (AVCaptureDevice *device in videoDevices) {
         [devices addObject:@{
+                             @"facing": device.position,
```
saghul: AFAIS Position is an NSInteger, we need a "front" or "back" string instead: https://developer.apple.com/documentation/avfoundation/avcapturedeviceposition?language=objc

KarlosQ: Thanks you.
It's correct now? I can't try in a real phone actually, so please try it if you can.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -120,8 +120,14 @@ - (RTCVideoTrack *)createVideoTrack:(NSDictionary *)constraints {
     NSMutableArray *devices = [NSMutableArray array];
     NSArray *videoDevices = [AVCaptureDevice devicesWithMediaType:AVMediaTypeVideo];
     for (AVCaptureDevice *device in videoDevices) {
+        NSString *position = @"";
+        if (device.position == 1) {
```
saghul: Please don't use numbers, there are defined constants.

KarlosQ: Ok, sorry. Can you check it now?

saghul:
Nice work, thanks!

PR #641, opened 2019-06-06
Fix mirror on landscape view

bighitbiker3:
I've confirmed this bug with a fresh project.  If you are mirroring the video and are in landscape mode the `subview.transform` logic flips the video vertically.  

It's unclear to me if this is the best approach as it seems like the subview should realize it's now horizontal and still be able to use `CGAffineTransformMakeScale` on its X axis.

If this isn't the best approach please point me in the right direction and I'd be happy to take a look.


bighitbiker3:
So I looked into this more and this weirdly only happens when there aren't any peers...not really sure where to start but closing this for now.

PR #642, opened 2019-06-11
add sending of dtmf ability

danjenkins:
Enable us sending DTMF

### DIFF
--- a/RTCPeerConnection.js
+++ b/RTCPeerConnection.js
@@ -104,6 +104,10 @@ export default class RTCPeerConnection extends EventTarget(PEER_CONNECTION_EVENT
     this._registerEvents();
   }
 
+  sendDTMF(tone) {
+    WebRTCModule.peerConnectionSendDTMF(tone, this._peerConnectionId);
+  }
+
   addStream(stream: MediaStream) {
     WebRTCModule.peerConnectionAddStream(stream.reactTag, this._peerConnectionId);
     this._localStreams.push(stream);
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -664,6 +664,25 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
+    public void peerConnectionSendDTMF(String tone, int id) {
+        ThreadUtils.runOnExecutor(() ->
+        peerConnectionSendDTMFAsync(tone, id));
+    }
+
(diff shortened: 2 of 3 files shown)


saghul:
Hi Dan! Sorry it took me a bit to review this. AFAICT the implemented API does not follow any spec, ie it's custom. That's not something this project aims to do, except when there is no "standard" way to do it.

Now, given the current state of our APIs (stream based, now considered deprecated) I think the best course of action is to implement it as RTCPeerConnection.createDTMFSender(). See: https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Using_DTMF Here is how I implemented it for Cordova back in the day: https://github.com/BasqueVoIPMafia/cordova-plugin-iosrtc/commit/427d80e28b62a60235507bf727c574e3b7be6cc0

redexp:
@saghul `createDTMFSender()` is deprecated. Instead of it `getSenders()` should be used like this

```javascript
let sender = pc.getSenders()[0];
sender.dtmf.insertDTMF("1", 500, 50);
```
[getSenders()](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/getSenders), [RTCRtpSender](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender), [RTCDTMFSender](https://developer.mozilla.org/en-US/docs/Web/API/RTCDTMFSender), [insertDTMF()](https://developer.mozilla.org/en-US/docs/Web/API/RTCDTMFSender/insertDTMF)

saghul:
I know. But we don't have `getSenders` yet, so that was a suggestion for a stopgap.

redexp:
@saghul As I understand all of this api are ready on android and ios side, all we need is to proxy them to js side, right? I can try to do that.

saghul:
iOS is also ready on the native side IIRC. The devil is always on the details ;-) Feel free to open a PR, that would be much appreciated!

harishgupta01:
Hi, Are you guys planning to this pull request?  

https://github.com/react-native-webrtc/react-native-webrtc/pull/642/commits/d5bf74d0379c48446987cd1bd90653f774f5b83b

saghul:
Not in the current shape.

sebamarucci:
Hi all,
I updated react-native-webrtc lib to implemented these two functions for iOS and Android:
`sendDtmfTone(dtmfTone, duration, interToneGap)`
`canInsertDtmf()`

I tried on a personal project and they seem to be working.

@saghul should I update current PR or create a new one?

saghul:
Closing. Now that we have unified plan this would need to be implemented following the w3c API.

PR #657, opened 2019-07-05
M75

saghul:


saghul:
This PR doesn't work with RN 60 since we haven't updated to AndroidX yet. I suggest you stay in 59. The update for 60 won't happen before this PR is done.

saghul:
That would be way too many things :-P

wilau2:
Thank you for opening this PR, I'll have some time next week to play with M75 with our VOIP IOs app. 🎉

naineshsolanki-lmi:
RTCStatisticsReport header is not published with the WebRTC framework. Is that intentional?

saghul:
@naineshsolanki-lmi No. When updating the framework I always delete it and copy over the new one, so what you see is how the framework straight out of the compiler, I make no changes to the umbrella header.

naineshsolanki-lmi:
RTCStatisticsReport is for the v2 statistics. Looks like it's ignored due to a bug in WebRTC build script. This change fixes it:
```--- a/sdk/BUILD.gn
+++ b/sdk/BUILD.gn
@@ -1282,6 +1282,7 @@ if (is_ios || is_mac) {
           "objc/api/peerconnection/RTCDtmfSender.h",
           "objc/api/peerconnection/RTCSSLAdapter.h",
           "objc/api/peerconnection/RTCSessionDescription.h",
+          "objc/api/peerconnection/RTCStatisticsReport.h",
           "objc/api/peerconnection/RTCTracing.h",
           "objc/api/peerconnection/RTCCertificate.h",
           "objc/api/peerconnection/RTCCryptoOptions.h",
@@ -1401,6 +1402,7 @@ if (is_ios || is_mac) {
           "objc/api/peerconnection/RTCRtpTransceiver.h",
           "objc/api/peerconnection/RTCSSLAdapter.h",
           "objc/api/peerconnection/RTCSessionDescription.h",
+          "objc/api/peerconnection/RTCStatisticsReport.h",
           "objc/api/peerconnection/RTCTracing.h",
           "objc/api/peerconnection/RTCVideoSource.h",
           "objc/api/peerconnection/RTCVideoTrack.h",```

saghul:
@naineshsolanki-lmi Is this reprorted upstream?

naineshsolanki-lmi:
Not yet. Will do.

naineshsolanki-lmi:
Here'e the bug report, incase anyone is interested to follow: https://bugs.chromium.org/p/webrtc/issues/detail?id=10816

sambegin:
@saghul  We've tested with the lastest commit in master branch and it seems that the "WebRTC-Audio-iOS-Holding" for the fix you've made for callkit audio issue isn't working. We are still experiencing audio failure when we hold a call. Do we miss something to activate that flag ? Is it enabled by default ? 

Could you provide any example on how you use it? 

Thanks a lot !

saghul:
@sambegin That’s unfortunate :-/ It’s enabled by default, so there is nothing special you need to do.

TBH I’m not sure what else to suggest here... :-(

sambegin:
@saghul  Does that mean that your webrtc fix was more of a "try" rather than a confirmation it was fixing this problem? Where you able to test it and see that it was fixing the issue ?

saghul:
> Does that mean that your webrtc fix was more of a "try" rather than a confirmation it was fixing this problem?

Correct, that's what the WebRTC team themselves said about it.

> Where you able to test it and see that it was fixing the issue ?

No. The use case I have doesn't involve hold so I've never ran into the problem yet.

PR #664, opened 2019-07-18
tools: change android_sdk env path

zxcpoiu:
According to webrtc source `src/build/android/envsetup.sh`, the android sdk path should be in `third_party/android_sdk/public/`, and we are using `third_party/android_tools/sdk/` currently.

Although they are almost the same, but I noticed there are some different files and different file size of the same file. 

Not sure if this is necessary

### DIFF
--- a/tools/build-webrtc.py
+++ b/tools/build-webrtc.py
@@ -126,7 +126,7 @@ def sync(target_dir, platform):
     path_parts = [env['PATH'], depot_tools_dir]
     if platform == 'android':
         # Same as . build/android/envsetup.sh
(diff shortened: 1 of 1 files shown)


saghul:
Have you ran into problems? I haven't, when I built M75.

zxcpoiu:
Nope, both build succeed (M75), just noticed the differences and report here as a memo.

saghul:
Ok. Let's keep it open and monitor it.

zxcpoiu:
`third_party/android_tools/sdk/` is removed on newer webrtc build.
Now we should use `third_party/android_sdk/public/`

zxcpoiu:
rebased

saghul (approved):

PR #665, opened 2019-07-18
Modern constraints

saghul:
This PR implements modern style constraints and the ability to customize the resolution and framerate. All feedback is welcome!

### DIFF
--- a/RTCPeerConnection.js
+++ b/RTCPeerConnection.js
@@ -2,7 +2,6 @@
 
 import EventTarget from 'event-target-shim';
 import {DeviceEventEmitter, NativeModules} from 'react-native';
-import * as RTCUtil from './RTCUtil';
 
 import MediaStream from './MediaStream';
 import MediaStreamEvent from './MediaStreamEvent';
@@ -14,6 +13,7 @@ import RTCSessionDescription from './RTCSessionDescription';
 import RTCIceCandidate from './RTCIceCandidate';
 import RTCIceCandidateEvent from './RTCIceCandidateEvent';
 import RTCEvent from './RTCEvent';
+import * as RTCUtil from './RTCUtil';
 
 const {WebRTCModule} = NativeModules;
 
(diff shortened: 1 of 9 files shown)


iZaL:
waiting for this PR to get merged

iZaL:
@danjenkins yes, I will do that and get back. thanks

vakrolme:
Does this mean that config.js is now working on mobile Jitsi Meet client?

saghul:
@vakrolme Can you reprase? config.js was always working, if you mean if the config values get taken into consideration, then yes, that should be the case now. Note that the `constraints` value is not yet used, but resolution`. We are working on this at the moment.

vakrolme:
@saghul yes, I'm confused about the process of setting custom constraints for mobile Jitsi. 

In short, I want to prioritize resolution over fps, so that resolution always stays as high as possible while fps may drop to as low as 13. There's an [older topic](https://community.jitsi.org/t/jitsi-dev-how-to-change-resolution-and-frame-rate-fps-of-jitsi/13883/15?u=froschgesicht) about constraints on Jitsi community forum, but it seems outdated because of this pull request.

So the question is where can I set `min` and `ideal` for both resolution & fps, and how?

P.S. somewhat off-topic: is there a way to set the RtpEncodingParameter `degradationPreference` to `maintain-resolution`?

saghul:
> So the question is where can I set `min` and `ideal` for both resolution & fps, and how?

In the `constraints` directive. Note, as I said earlier, that we don't use that yet, it's going to happen very very soon.

> P.S. somewhat off-topic: is there a way to set the RtpEncodingParameter `degradationPreference` to `maintain-resolution`?

Not at the moment. We'd have to implement support for that in this plugin to begin with.

vakrolme:
Thanks for a swift reply!

> it's going to happen very very soon.

Is there an estimate on the release?

saghul:
> Is there an estimate on the release timeframe?

I'm working on it this week. It may slip to the next.

vakrolme:
> > Is there an estimate on the release timeframe?
> 
> I'm working on it this week. It may slip to the next.

@saghul can you please give some update, whether the issue was de-prioritized?

saghul:
Yep it got pushed back due to unforeseen issues that need fixing beforehand, sorry about that.

vakrolme:
> Yep it got pushed back due to unforeseen issues that need fixing beforehand, sorry about that.

Are these issues also responsible for `maxFps` being ignored on mobile? https://github.com/jitsi/jitsi-meet/issues/4946

I've just found out about `maxFps` and `minFps` and tried using them as a substitute for constraints, but they don't seem to be taken into consideration.

My `config.js`:
```
    resolution: 1080,
    minFps: 13,
    maxFps: 25,
```

saghul:
No, all of those are ignored.

PR #669, opened 2019-07-30
api: support legacy constraints in create{Offer,Answer}

saghul:


saghul:
ping @danjenkins

saghul:
How does your constraints object look like exactly?

PR #677, opened 2019-08-21
Add note about upgrading the `platform` field in Podfile

jacklj:
The default react native project (created with `react-native init`) has in the generated Podfile: `platform :ios, '9.0'.
When doing `pod install` during `react-native-webrtc` installation, this gives the error:
```
[!] CocoaPods could not find compatible versions for pod "react-native-webrtc":
  In Podfile:
    react-native-webrtc (from `../node_modules/react-native-webrtc`)

Specs satisfying the `react-native-webrtc (from `../node_modules/react-native-webrtc`)` dependency were found, but they required a higher minimum deployment target.
```
Bumping the iOS version number to `10.0` fixes this error.

### DIFF
--- a/Documentation/iOSInstallation.md
+++ b/Documentation/iOSInstallation.md
@@ -26,6 +26,12 @@ Include in a Podfile in your react-native ios directory:
 pod 'react-native-webrtc', :path => '../node_modules/react-native-webrtc'
 ```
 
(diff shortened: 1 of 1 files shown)


mitjnextt:
For me it resolve by changing platform to 11.0, I think now it requires `platform :ios, '11.0'`

Emetrop:
I can confirm that platform is needed now to set up to `11.0`

PR #688, opened 2019-09-05
iOS fix (random) black video

saghul:


ios/RCTWebRTC/VideoCaptureController.m
```diff
@@ -71,15 +71,31 @@ -(void)startCapture {
...
+        if (err) {
+            RCTLogError(@"[VideoCaptureController] Error starting capture: %@", err);
+        } else {
+            RCTLog(@"[VideoCaptureController] Capture started");
+        }
+        dispatch_semaphore_signal(semaphore);
+    }];
 
-    RCTLog(@"[VideoCaptureController] Capture started");
+    dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
```
paweldomas: Does it run on the WebRTC thread and will block it forever if callback does not happen? I'd say it should give up after some time and log an error.

ios/RCTWebRTC/VideoCaptureController.m
```diff
@@ -71,15 +71,31 @@ -(void)startCapture {
...
+    // Stopping the capture happens on another thread. Wait for it.
+    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
+
+    [_capturer stopCaptureWithCompletionHandler:^{
+        RCTLog(@"[VideoCaptureController] Capture stopped");
+        dispatch_semaphore_signal(semaphore);
+    }];
 
-    RCTLog(@"[VideoCaptureController] Capture stopped");
+    dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
```
paweldomas: same here

ios/RCTWebRTC/VideoCaptureController.m
paweldomas: I'm thinking 10 seconds

paweldomas: Also: this wait does not happen when permissions are being requested? It's just async stuff?

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -215,7 +215,7 @@ - (RTCVideoTrack *)createVideoTrack:(NSDictionary *)constraints {
 RCT_EXPORT_METHOD(mediaStreamTrackSetEnabled:(nonnull NSString *)trackID : (BOOL)enabled)
 {
   RTCMediaStreamTrack *track = [self trackForId:trackID];
-  if (track && track.isEnabled != enabled) {
```
paweldomas: why is the check unneeded? Is that checked on another level? Otherwise it would be starting the capturer twice?

saghul: It's checked in JS already. Since we are now waiting for the capture to start, there is no way for 2 starts to happen in parallel. Calling start twice is ok, for example to change the resolution.

ios/RCTWebRTC/VideoCaptureController.m
saghul: Permissions happen out of here, so no, this won't wait for that.

I don't want to add an arbitrary timeout. The handler will always be called. If it's not, it means we have crashed os the OS is in a bad state, in which case there is nothing we can do to help anyway.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
paweldomas: I think it would be god to put that explanation in the commit message.

ios/RCTWebRTC/VideoCaptureController.m
paweldomas: ok

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
saghul: Can do!

saghul:
@paweldomas Updated, PTAL!

PR #694, opened 2019-09-13
Stop an extra setConfig call

TylerWhitman:


TylerWhitman:
Sorry created this on accident

PR #704, opened 2019-09-26
Feature/datachannel with video

Seachaos:


saghul:
?

Seachaos:
it's a mistake, sorry.
we fork it to do some customized change.

saghul:
Have you ever considered discussing upstreaming sons of your changes? If we’d do the same you’d have to write everything by yourself. Be a good Open Source citizen.

Seachaos:
Yes, but we are fork from `1.67.1` and tweak something for bundle mode support.
since WebRTC library change a lot we can't catch up newest version.
that's why we keep old version and maintenance our self ...

When we have time will update it : )
Thanks for you suggestion.

Seachaos:
here is our project.
https://github.com/deepsentinel/react-native-webrtc

Doing experiment and keep old version WebRTC library there.
so it's can not merge anymore. ( that's why have mistake PR...)
But if you want still can look it.

PR #711, opened 2019-10-17
Expose registerGlobals()

### LINKED ISSUE #685: Add registerGlobals() function to pollute global namespace with RTC entities
As discussed [here](https://github.com/versatica/mediasoup-client/issues/5#issuecomment-526581786), there are some WebRTC based JS libraries that depend on `RTCPeerConnection` and so on being available in global namespace.

`cordova-plugin-iosrtc` exposes [iosrtc.registerGlobals()](https://github.com/cordova-rtc/cordova-plugin-iosrtc/blob/master/docs/iosrtc.md#iosrtcregisterglobals) for that.

So the only requirement is that `react-native-webrtc` is loaded first, then the app calls to `registerGlobals()` and then the 3rd party WebRTC library can be load.

ibc:
#### registerGlobals()

By calling this method the JavaScript global namespace gets "polluted" with the following additions:

* `navigator.getUserMedia()` (deprecated W3 API)
* `navigator.mediaDevices.getUserMedia()`
* `navigator.mediaDevices.enumerateDevices()`
* `window.RTCPeerConnection`
* `window.RTCIceCandidate`
* `window.RTCSessionDescription`
* `window.MediaStream`
* `window.MediaStreamTrack`

This is useful to make existing WebRTC JavaScript libraries (that expect those globals to exist) work with react-native-webrtc.

It fixes #685

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,8 @@ import {
   RTCView,
   MediaStream,
   MediaStreamTrack,
-  mediaDevices
+  mediaDevices,
+  registerGlobals
 } from 'react-native-webrtc';
 ```
 Anything about using RTCPeerConnection, RTCSessionDescription and RTCIceCandidate is like browser.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -116,6 +117,22 @@ Rendering RTCView.
 
 ### Custom APIs
 
+#### registerGlobals()
+
+By calling this method the JavaScript global namespace gets "polluted" with the following additions:
+
+* `navigator.getUserMedia()` (deprecated W3 API)
```
saghul: Let's not add this one, since we don't support the callback API and some shims may think we do because we support thie deprecated method.

ibc: Done.

saghul:
Cheers!

Issue #679, opened 2019-08-26
Can I get the RTCView's start paly callback like html <video> onPlay=function()?


jf4444:
#### Platform information

* **React Native version**:
* **Plugin version**:
* **OS**: 
* **OS version**:

saghul:
I think it is possible to implement, but not high on the priority list. PRs are welcome, of course.

sesm:
We have similar feature in our fork: `onFirstFrame` callback that fires when first video frame is received. We use it to stop `loading` animation on top of video view. This is similar to what @jf4444 requested.
Note, however, that audio stream usually starts playing earlier, so this is not exactly the same as `<video> onPlay`
I've created a PR just in case. Feel free to comment, i.e. a different callback name etc

PR #715, opened 2019-10-28
Add onFirstFrame callback

sesm:
This PR fixes https://github.com/react-native-webrtc/react-native-webrtc/issues/679 by adding `onFirstFrame` callback, which fires when first video frame is received. 
Note, however, that audio stream can start playing earlier.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
@@ -4,6 +4,7 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import android.graphics.Color;
 import android.graphics.Point;
+import android.graphics.Color;
 import android.support.v4.view.ViewCompat;
 import android.view.View;
 import android.view.ViewGroup;
@@ -23,6 +24,12 @@ import org.webrtc.RendererCommon.RendererEvents;
 import org.webrtc.RendererCommon.ScalingType;
 import org.webrtc.SurfaceViewRenderer;
 import org.webrtc.VideoTrack;
+
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.bridge.Callback;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.uimanager.events.RCTEventEmitter;
 
 public class WebRTCView extends ViewGroup {
     /**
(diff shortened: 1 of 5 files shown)


ios/RCTWebRTC/RTCVideoViewManager.m
```diff
@@ -263,11 +267,41 @@ - (void)setVideoTrack:(RTCVideoTrack *)videoTrack {
...
+
+/**
+ * Renders a specific video frame. Delegates to the subview of this instance
+ * which implements the actual {@link RTCVideoRenderer}.
+ *
+ * @param frame The video frame to render.
+ */
+- (void)renderFrame:(RTCVideoFrame *)frame {
+    id<RTCVideoRenderer> videoRenderer = self.videoView;
+    if (!firstFrameRendered) {
```
zetorama: In case if I do not provide callback, the App crushes, throwing `EXC_BAD_ACCESS` in Xcode.

I'm not an expert in Objective-C, but this seems to be enough as a quick fix:
```suggestion
    if (self.onFirstFrame && !firstFrameRendered) {
```

Also, note, that I haven't tested/checked Android version.

sesm:
@zetorama applied your changes, thank you! Also fixed a merge artefact for Android. Docs incoming.

sesm:
Updated docs.

android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
```diff
@@ -4,6 +4,7 @@
 import android.content.Context;
 import android.graphics.Color;
 import android.graphics.Point;
+import android.graphics.Color;
```
saghul: duplicated import

android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
```diff
@@ -286,6 +293,13 @@ protected void onDetachedFromWindow() {
     private void onFirstFrameRendered() {
         post(() -> {
             Log.d(TAG, "First frame rendered.");
+            WritableMap event = Arguments.createMap();
+            ReactContext reactContext = (ReactContext) getContext();
+            reactContext.getJSModule(RCTEventEmitter.class).receiveEvent(
+                    getId(),
+                    "onFirstFrame",
+                    event
```
saghul: can we skip the empty object here and pass null?

ios/RCTWebRTC/RTCVideoViewManager.m
```diff
@@ -263,11 +267,41 @@ - (void)setVideoTrack:(RTCVideoTrack *)videoTrack {
...
 }
+#pragma mark - RTCVideoRenderer methods
+
+/**
+ * Renders a specific video frame. Delegates to the subview of this instance
+ * which implements the actual {@link RTCVideoRenderer}.
+ *
+ * @param frame The video frame to render.
+ */
+- (void)renderFrame:(RTCVideoFrame *)frame {
```
saghul: I'd like to see this logic in the VideoTrackAdapter instead, since it already counts video frames.

android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
sesm: Fixed

android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
sesm: Yes we can, fixed

ios/RCTWebRTC/RTCVideoViewManager.m
sesm: I tried to do this, and I couldn't find how to connect `onFirstFrame` in `RTCVideoViewManager` with `VideoTrackAdapter`. Can you please give me a hint on how I can do it? I have no idea how those 2 classes are connected.

sesm:
@saghul sorry for later response, I've applied review comments for Android, but couldn't do it for iOS. Can you help me a little bit? I can't find a way to call `onFirstFrame` of `RTCVideoViewManager` in `VideoTrackAdapter`.

wvddrss:
any update on this?

PR #724, opened 2019-11-19
Add video frame layout (size and rotation) event

anion155:
I needed the way to get video size from RTCView. I made this fork.

### DIFF
--- a/RTCView.js
+++ b/RTCView.js
@@ -65,6 +65,7 @@ const View = requireNativeComponent('RTCVideoView', RTCView, {nativeOnly: {
   accessibilityLiveRegion: true,
   importantForAccessibility: true,
   onLayout: true,
+  onFrameLayout: true,
   nativeID: true,
 }});
 
--- a/android/src/main/java/com/oney/WebRTCModule/RTCVideoViewManager.java
+++ b/android/src/main/java/com/oney/WebRTCModule/RTCVideoViewManager.java
@@ -1,8 +1,13 @@
 package com.oney.WebRTCModule;
 
+import android.support.annotation.Nullable;
+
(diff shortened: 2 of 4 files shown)


saghul:
Thanks for the PR! Code looks good. I'm a bit on the fence about it because it's non-standard, but the same can be said about mirroring I guess... Also IIRC this information is missing from new stats.

anion155:
Standart has `MediaStreamTrack#getSettings` method for this kind of information. https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings https://w3c.github.io/mediacapture-main/#dom-mediastreamtrack-getsettings

saghul:
True that. I guess we could emit an event and capture the values internally so they can be returned when getSettings is called on the track.

sesm:
@anion155 sorry, if I'm missing something, but you can get video width and height from `getStats()` in `googFrameHeightReceived` and `googFrameWidthReceived`

anion155:
@sesm https://developer.mozilla.org/en-US/docs/Web/API/RTCStatsReport

saghul:
@sesm That's what I suggested, yeah. But if I'm not mistaken those are gone in the new stats (that we are not yet using)

anion155:
> @sesm https://developer.mozilla.org/en-US/docs/Web/API/RTCStatsReport

I mean, that there is no such stats in standart.

saghul:
See page 14 here: https://docs.google.com/document/d/1z-D4SngG36WPiMuRvWeTMN7mWQXrf1XKZwVl3Nf1BIE It's alreeady part of the new stats. You should be able to use the legacy one right now, since it's what this plugin uses, still.

PR #726, opened 2019-11-22
fix: ios data channel

iScorpio:
dataChannel.channelId is always equal to -1 while connecting. 
If create data channel with id 0 then it is not work on ios

### DIFF
--- a/RTCPeerConnection.js
+++ b/RTCPeerConnection.js
@@ -386,7 +386,7 @@ export default class RTCPeerConnection extends EventTarget(PEER_CONNECTION_EVENT
       // is reserved due to SCTP INIT and INIT-ACK chunks only allowing a
       // maximum of 65535 streams to be negotiated (as defined by the WebRTC
       // Data Channel Establishment Protocol).
-      for (id = 0; id < 65535 && dataChannelIds.has(id); ++id);
+      for (id = 1; id < 65535 && dataChannelIds.has(id); ++id);
       // TODO Throw an error if no unused id is available.
       dataChannelDict = Object.assign({id}, dataChannelDict);
     }
--- a/ios/RCTWebRTC/WebRTCModule+RTCDataChannel.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCDataChannel.m
@@ -29,12 +29,10 @@ RCT_EXPORT_METHOD(createDataChannel:(nonnull NSNumber *)peerConnectionId
 {
   RTCPeerConnection *peerConnection = self.peerConnections[peerConnectionId];
   RTCDataChannel *dataChannel = [peerConnection dataChannelForLabel:label configuration:config];
-  // XXX RTP data channels are not defined by the WebRTC standard, have been
-  // deprecated in Chromium, and Google have decided (in 2015) to no longer
-  // support them (in the face of multiple reported issues of breakages).
-  if (-1 != dataChannel.channelId) {
+  if (dataChannel != nil && (dataChannel.readyState == RTCDataChannelStateConnecting
+      || dataChannel.readyState == RTCDataChannelStateOpen)) {
     dataChannel.peerConnectionId = peerConnectionId;
-    NSNumber *dataChannelId = [NSNumber numberWithInteger:dataChannel.channelId];
+    NSNumber *dataChannelId = [NSNumber numberWithInteger:config.channelId];
     peerConnection.dataChannels[dataChannelId] = dataChannel;
     dataChannel.delegate = self;
   }


sesm:
@iScorpio could you please describe the problem you are trying to fix and explain how the provided changes fix it?
Thanks!

iScorpio:
You can read discussion https://react-native-webrtc.discourse.group/t/datachannel-id-is-1-and-no-callbacks-on-ios/323/2

After call RTCDataChannel *dataChannel = [peerConnection dataChannelForLabel:label configuration:config];
I have dataChannel where channelId is always equal to -1 while dataChannel have state connecting.
Then the condition "-1 != dataChannel.channelId" nor when not will work

also if channelId is 0 it is not work on IOS

iScorpio:
It also fixes a possible crash. See webrtc source code: https://cs.chromium.org/chromium/src/third_party/webrtc/sdk/objc/api/peerconnection/RTCPeerConnection%2BDataChannel.mm?g=0&l=28

saghul:
Aare you starting the DataChannel from your RN app or receiving the open?

saghul:
Alsom is Android affected by this too>

iScorpio:
> Aare you starting the DataChannel from your RN app or receiving the open?

I am use jitsi-meet sdk for ios. I have problem with openBridgeChannel via dataChannel

saghul:
> I am use jitsi-meet sdk for ios. I have problem with openBridgeChannel via dataChannel.

LoL, I work on that. Maybe we haven't noticed because we have switched to WS for data transport... I'll double check and get back to you.

Do you have any error in the logs you can paste?

iScorpio:
> > I am use jitsi-meet sdk for ios. I have problem with openBridgeChannel via dataChannel.
> 
> LoL, I work on that. Maybe we haven't noticed because we have switched to WS for data transport... I'll double check and get back to you.
> 
> Do you have any error in the logs you can paste?

No. I used debugger and source code webrtc to detect the problem

saghul:
Good find, thank you for digging into it!

gev:
Hi!
@iScorpio, I'd tested your fork. Works as well, but only for one channel

iScorpio:
any news ?

saghul:
I'm looking into this now, sorry for the delay. I'll hopefully have a final review tomorrow.

saghul:
Verified it works, thanks a lot for the patch! I'll be making a new release.

gev:
still can create only one data channel =)

saghul:
Sorry, I have no time to look into it, but a patch would be most welcome.

gev:
Hi, guys!

Can't understand!

In RTCPeerConnection.js
"dataChannelDict" has field "id":
```
  createDataChannel(label: string, dataChannelDict?: ?RTCDataChannelInit) {
    let id;
    const dataChannelIds = this._dataChannelIds;
    if (dataChannelDict && 'id' in dataChannelDict) {
      id = dataChannelDict.id;
      if (typeof id !== 'number') {
        throw new TypeError('DataChannel id must be a number: ' + id);
      }
      if (dataChannelIds.has(id)) {
        throw new ResourceInUse('DataChannel id already in use: ' + id);
      }
    } else {
      // Allocate a new id.
      // TODO Remembering the last used/allocated id and then incrementing it to
      // generate the next id to use will surely be faster. However, I want to
      // reuse ids (in the future) as the RTCDataChannel.id space is limited to
      // unsigned short by the standard:
      // https://www.w3.org/TR/webrtc/#dom-datachannel-id. Additionally, 65535
      // is reserved due to SCTP INIT and INIT-ACK chunks only allowing a
      // maximum of 65535 streams to be negotiated (as defined by the WebRTC
      // Data Channel Establishment Protocol).
      for (id = 1; id < 65535 && dataChannelIds.has(id); ++id);
      // TODO Throw an error if no unused id is available.
      dataChannelDict = Object.assign({id}, dataChannelDict);
    }
    WebRTCModule.createDataChannel(
        this._peerConnectionId,
        label,
        dataChannelDict);
    dataChannelIds.add(id);
    return new RTCDataChannel(this._peerConnectionId, label, dataChannelDict);
  }

```
but WebRTCModule.createDataChannel() in WebRTCModule+RTCDataChannel.m
use field "channelId" in the "config" object instead "id":
```
RCT_EXPORT_METHOD(createDataChannel:(nonnull NSNumber *)peerConnectionId
                              label:(NSString *)label
                             config:(RTCDataChannelConfiguration *)config
{
  RTCPeerConnection *peerConnection = self.peerConnections[peerConnectionId];
  RTCDataChannel *dataChannel = [peerConnection dataChannelForLabel:label configuration:config];
  if (dataChannel != nil && (dataChannel.readyState == RTCDataChannelStateConnecting
      || dataChannel.readyState == RTCDataChannelStateOpen)) {
    dataChannel.peerConnectionId = peerConnectionId;
    NSNumber *dataChannelId = [NSNumber numberWithInteger:config.channelId];
    peerConnection.dataChannels[dataChannelId] = dataChannel;
    dataChannel.delegate = self;
  }
})

```
how field "id" is  converted to "channelId"?

--
Evgeny

gev:
i've found out strange thing.
When setup odd id for a dada channel like: 
`peer.createDataChannel(ASSET, { id: 3, ordered: true, maxPacketLifeTime: 10000 });`
The data channel has opened as well!

PR #763, opened 2020-03-21
1. implements 'mediaDevices.getDisplayMedia'.

zyCwind:
:-)


ahroaZahedi:
I have problem with this code when I want to create new intent in Android for example for file picker I get error in logcat as null pointing exception and I think it's because that code things it's permission requesting intent how can I fix it

PR #769, opened 2020-04-07
Important Fix: Problem in allocating data channel zero

oyerli:
The latest fix on RTCPeerConnection.js broke the allocation of data channel 0. This fixes the problem.


saghul:
This got refactored and no longer applies.

PR #783, opened 2020-05-09
POC: android: support snapshot through RTCView via SurfaceViewRenderer

zxcpoiu:
## Motivation

There has been several discussions and implementations since 2016, see #176
But it needs to patch the underling webrtc build, and only works for local stream (through camera api)

On android, this PR can take snapshot use `onFrame`, so it support both `local` and `remote` stream. To be accurate, it catches video frame from RTCView, and the api invocation is based on each RTCView as well.

Thanks to all contributors in #176 
Especially @davidperrenoud and @EdwarDu
The `SnapshotUtils.java` are credit to the original author.

**Note: I haven't test in detail for every saveTarget and resize part (`maxSize`)**

This PR is a POC and mainly for discussion and test.

## How this PR works

We can add `onFrame` listener to `SurfaceViewRenderer` of  specific `RTCView`, and the event contains Bitmap, which is what we want.

#### see below critical part of codes:

```java
surfaceViewRenderer.addFrameListener(new EglRenderer.FrameListener() {
    @Override
    public void onFrame(Bitmap bitmap) {
        try {
            // --- default is convert to jpeg and save to cameraRoll
        } finally {
            // --- remove listener immediately
            surfaceViewRenderer.removeFrameListener(this);
            // --- send result to js
            module.sendEvent("WebRTCViewSnapshotResult", params);
        }
    }
}, 1);
```

## API differences

AFAIK, the method used in this PR does not exist on iOS, so the two platform might using different approach.

* on android: catch frame from RTCView ( RTCView only )
* on ios: use `react-native-view-shot` ( full screen, like phone's embedded hardware screenshot )

So on ios, if you want only screenshot for specific RTCView, you might need to hide other components yourself.

## Combined Usage Example For Android / iOS

This example only cares about take snapshot for REMOTE Peer

**Note: you need to handle correspond permissions yourself depends on your `saveTarget`**

```js
class CallView extends React.Component {

    constructor(props) {
        this.state = {
            // --- for android to trigger snapshot on remote RTCView
            remoteVideoViewSnapshotOption: null,
            // --- for ios to control whether should render full screen remote view
            hideAllViewExceptRemoteRTCView: false,
        }
    }

    componentDidMount() {
            DeviceEventEmitter.addListener('WebRTCViewSnapshotResult', this.onWebRTCViewSnapshotResult);
    }

    componentWillUnmount() {
            DeviceEventEmitter.removeAllListeners('WebRTCViewSnapshotResult');
    }

    onWebRTCViewSnapshotResult(data) {
        // --- reset option after we got event. It's okey to not reset, since it will trigger a new screenshot as long as props did change again. (likely use shallow compare.)
        this.setState({ remoteVideoViewSnapshotOption: null });
        if (data.error) {
            // --- failed
        }
        if (data.file) {
            // --- saved file path
        }
    }

    async onPressTakeSnapshot() {
        if (PlatformOS.platform === 'android') {
                // --- check required permission, in the case of cameraRoll, you need: READ/WRITE EXTERNAL STORAGE permission
                let hasEnoughPermission = await checkScreenshotPermission();
                if (!hasEnoughPermission) {
                    return false;
                }
                // --- if snapshotOption != null and changed, the remote RTCView will trigger snapshot once and fire an event for result.
                let snapshotOption = {
                    id: uuid.v4(), // --- use any value you think it's unique for each screenshot
                    saveTarget: 'cameraRoll',
                };
                this.setState({ remoteVideoViewSnapshotOption: snapshotOption });
        } else {
            // IOS
            // --- optional: if you want full screen remote view only, take snapshot after you've adjust your views
            this.setState({ hideAllViewExceptRemoteRTCView: true }, async () => {
                    captureScreen = require("react-native-view-shot").captureScreen;
                    CameraRoll = require("@react-native-community/cameraroll");

                    let uriCache = await captureScreen({ format: 'jpg' });
                    if (!uriCache) {
                        return false;
                    }
                    let uriSaved = CameraRoll.saveToCameraRoll(uriCacheRenamed, 'photo');

                    // --- optional: recover your view setup
                    this.setState({ hideAllViewExceptRemoteRTCView: false });
            }
        }
    }

    render() {
        if (this.state.hideAllViewExceptRemoteRTCView) {
            // --- style is for iOS return full screen remote view
            // --- snapshotOption is for android to trigger snapshot
            return (
                 <RTCView
                      streamURL={streamUrl}
                      style={styles.fullScreen}
                      snapshotOption={(streamUrl ? this.state.snapshotOption : null)}
                 />
            );
        } else {
            // --- return (.....full call view setup)
        }
    }
```

## Discussion

I know it's a bit strange to use props to trigger like a function.
But we can not get RTCView instance from module currently. ( And I am not sure what is the proper way for doing so )

If we can store RTCView ID on native side, we might endup have below more intuitive api, like:

* `WebRTC.takeSnapshot(viewId)`
or
* `videoTrack.takeSnapshot()`, and find the corresponded RTCView id at native side

### DIFF
--- a/RTCView.js
+++ b/RTCView.js
@@ -53,7 +53,8 @@ const RTCView = {
      * remote video(s) which appear in the background, and 1 for the local
      * video(s) which appear above the remote video(s).
      */
-    zOrder: PropTypes.number
+    zOrder: PropTypes.number,
+    snapshotOption: PropTypes.object,
   },
 };
 
--- a/android/src/main/java/com/oney/WebRTCModule/RTCVideoViewManager.java
+++ b/android/src/main/java/com/oney/WebRTCModule/RTCVideoViewManager.java
@@ -1,5 +1,8 @@
 package com.oney.WebRTCModule;
 
+import android.support.annotation.Nullable;
+
+import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.uimanager.annotations.ReactProp;
 import com.facebook.react.uimanager.SimpleViewManager;
 import com.facebook.react.uimanager.ThemedReactContext;
@@ -70,5 +73,18 @@ public class RTCVideoViewManager extends SimpleViewManager<WebRTCView> {
   @ReactProp(name = "zOrder")
   public void setZOrder(WebRTCView view, int zOrder) {
     view.setZOrder(zOrder);
+  }
+
+  /**
+   * Take a snapshot of a specific {@link WebRTCView}
+   *
+   * @param view The {@code WebRTCView} on which the specified {@code option} is
+   * to be set.
+   * @param snapshotOption Represent each unique snapshot action, everytime the option changes,
+   * will triiger one snapshot.
+   */
+  @ReactProp(name = "snapshotOption")
+  public void handleTakeSnapshot(WebRTCView view, @Nullable ReadableMap snapshotOption) {
+    view.handleTakeSnapshot(snapshotOption);
   }
 }
--- a/android/src/main/java/com/oney/WebRTCModule/SnapshotUtils.java
+++ b/android/src/main/java/com/oney/WebRTCModule/SnapshotUtils.java
@@ -1,0 +1,161 @@
+package com.oney.WebRTCModule;
+
+import com.facebook.react.bridge.ReactContext;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.Matrix;
+import android.graphics.Rect;
+import android.media.MediaScannerConnection;
+import android.net.Uri;
+import android.os.Environment;
+import android.util.Log;
+
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+
+import java.util.Base64;
+import java.util.UUID;
+
+class SnapshotUtils {
+    private static final String TAG = WebRTCModule.TAG;
+
+    public static final String RCT_CAMERA_SAVE_TARGET_MEMORY = "memory";
+    public static final String RCT_CAMERA_SAVE_TARGET_DISK = "disk";
+    public static final String RCT_CAMERA_SAVE_TARGET_CAMERA_ROLL = "cameraRoll";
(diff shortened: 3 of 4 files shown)


mickey-ichi (approved):

saghul:
Sorry @zxcpoiu but I think it's time we close this. I really don't want to take on the maintenance burden for such a niche feature. If WebRTC itself gains some simpler API to do it we can revisit.

zxcpoiu:
Sure thing! I didn't want it to be merged either, unless there are much more solid api provided officially for both ios and android. Just sharing it for people may need it while tracking if someone knows whether there are upcoming api I had missed.

unnfajar:
> Sure thing! I didn't want it to be merged either, unless there are much more solid api provided officially for both ios and android. Just sharing it for people may need it while tracking if someone knows whether there are upcoming api I had missed.

hi @zxcpoiu, I tried to implement your POC on version 1.87. I modified 4 files, RTCView.js, RTCVideoViewManager.java, SnapshotUtils.java, WebRTCView.java, and used your example above to try to take a snapshot for REMOTE Peer. But  WebRTCViewSnapshotResult was not triggered at all. Did I miss something?

unnfajar:
> > Sure thing! I didn't want it to be merged either, unless there are much more solid api provided officially for both ios and android. Just sharing it for people may need it while tracking if someone knows whether there are upcoming api I had missed.
> 
> hi @zxcpoiu, I tried to implement your POC on version 1.87. I modified 4 files, RTCView.js, RTCVideoViewManager.java, SnapshotUtils.java, WebRTCView.java, and used your example above to try to take a snapshot for REMOTE Peer. But WebRTCViewSnapshotResult was not triggered at all. Did I miss something?

I figured the issue : 

  "snapshotOption={(streamUrl ? this.state.snapshotOption : null)}"

these props require value from snapshotOption state, but snapshotOption was saved on remoteVideoViewSnapshotOption state. so change it to remoteVideoViewSnapshotOption fixed it for me.

  "snapshotOption={(streamUrl ? this.state.snapshotOption : null)}"

PR #787, opened 2020-05-18
feat: update localDescription

onesmash:
update the localDescription when `peerConnectionGotICECandidate` and 
`peerConnectionIceGatheringChanged` event trigger

### DIFF
--- a/RTCPeerConnection.js
+++ b/RTCPeerConnection.js
@@ -309,6 +309,7 @@ export default class RTCPeerConnection extends EventTarget(PEER_CONNECTION_EVENT
         if (ev.id !== this._peerConnectionId) {
           return;
         }
+        this.localDescription = new RTCSessionDescription(ev.localDescription)
         const candidate = new RTCIceCandidate(ev.candidate);
         const event = new RTCIceCandidateEvent('icecandidate', {candidate});
         this.dispatchEvent(event);
@@ -318,7 +319,7 @@ export default class RTCPeerConnection extends EventTarget(PEER_CONNECTION_EVENT
           return;
         }
         this.iceGatheringState = ev.iceGatheringState;
-
+        this.localDescription = new RTCSessionDescription(ev.localDescription)
         if (this.iceGatheringState === 'complete') {
           this.dispatchEvent(new RTCIceCandidateEvent('icecandidate', null));
         }
--- a/RTCPeerConnection.js
+++ b/RTCPeerConnection.js
@@ -79,6 +79,7 @@ export default class RTCPeerConnection extends EventTarget(PEER_CONNECTION_EVENT
   _localStreams: Array<MediaStream> = [];
   _remoteStreams: Array<MediaStream> = [];
   _subscriptions: Array<any>;
+  _senders: Array<any> = [];
 
   /**
    * The RTCDataChannel.id allocator of this RTCPeerConnection.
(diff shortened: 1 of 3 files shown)


saghul:
Why is this necessary?

onesmash:
> Why is this necessary?

localDescription is updated when peerConnectionGotICECandidate and
peerConnectionIceGatheringChanged event trigger

saghul:
> localDescription is updated when peerConnectionGotICECandidate and
> peerConnectionIceGatheringChanged event trigger

I see that's what the does, but I still don't know why. Do browsers behave this way or is a call to getLocalDescription necessary to get the updated SDP?

onesmash:
> > localDescription is updated when peerConnectionGotICECandidate and
> > peerConnectionIceGatheringChanged event trigger
> 
> I see that's what the code does, but I still don't know why you want to do it. Do browsers behave this way or is a call to getLocalDescription necessary to get the updated SDP?

Browsers behave this way，the ice gathering task will update the SDP, https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/iceGatheringState

saghul:
> Browsers behave this way，the ice gathering task will update the SDP, https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/iceGatheringState

That website says nothing about updating the local SDP. Can you find some normative text (or jsfiddle to run in a browser) which validates this?

Prussich:
Guys, @onesmash is right, I've been using my own patch for this for about 3 years. If the local ICE candidate has been updated, then the localDescription SDP should be updated at the same moment. I am very surprised that this has not been fixed at all until now. Even if it's not documented, at least this is how it works in all browsers. You can check it out. You can also read the documentation about it https://developer.mozilla.org/ru/docs/Web/API/WebRTC_API/Signaling_and_video_calling 
"ICE candidate exchange process". The image contains the phrase: "Generate an ICE candidate represented by an SDP string".

saghul:
So you've patched the issue yourself and not shared the patch for 3 years?! Way to go.

Anyway. I realized this applies to other operations as well, like setting the local / remote SDPs. Alternatively we could consider making the better for the SDPs a sync method but that may have performance implications.

Prussich:
> So you've patched the issue yourself and not shared the patch for 3 years?! Way to go.
> 
> Anyway. I realized this applies to other operations as well, like setting the local / remote SDPs. Alternatively we could consider making the better for the SDPs a sync method but that may have performance implications.

Sorry, but if I'm not mistaken, this issue was raised earlier, but you also did not want to fix it. 
The point is that the native module is already doing update localDescription after new candidate was added. It's enough just to ask updated SDP from native module after event with the new candidate appears. Something like this:

```js
EventEmitter.addListener('peerConnectionGotICECandidate', ev => {
        if (ev.id !== this._peerConnectionId) {
          return;
        }
        const candidate = new RTCIceCandidate(ev.candidate);
        const event = new RTCIceCandidateEvent('icecandidate', {candidate});

       WebRTCModule.peerConnectionGetLocalDescription(
            this._peerConnectionId,
            (successful, data) => {
                if (successful) {
                    this.localDescription = new RTCSessionDescription(data);
                    this.dispatchEvent(event);
                }
            });
}),
```

Also you need update localDescription on peerConnectionIceGatheringChanged event too.
This fix does not affect performance in any way. I do not think that additional actions are required for the remote SDP. Everything else works perfectly. May be this is more about the application itself and the signaling protocol for synchronizing these SDPs.

saghul:
The performance implication I mentioned is if we decide to implement the getter as a sync method.

PRs are welcome, but we'll need iOS as well.

PR #804, opened 2020-06-12
Use NativeEventEmitter instead of DeviceEventEmitter

r0b0t3d:
Since `DeviceEventEmitter` is [deprecated](https://github.com/facebook/react-native/blob/1490ab12ef156bf3201882eeabfcac18a1210352/Libraries/EventEmitter/RCTNativeAppEventEmitter.js#L15), this PR switch to use `NativeEventEmitter`

Changes:
1. JS side
- replace `DeviceEventEmitter` with `NativeEventEmitter`
2. iOS native
- extends `RCTEventEmitter`
- use `sendEventWithName` from `RCTEventEmitter` to send event to JS.

### DIFF
--- a/EventEmitter.js
+++ b/EventEmitter.js
@@ -1,0 +1,7 @@
+import {NativeModules, NativeEventEmitter} from 'react-native';
+
+const { WebRTCModule } = NativeModules;
+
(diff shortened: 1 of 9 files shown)


saghul:
Cheers!

PR #855, opened 2020-08-19
macos support

amanforindia:


sesm:
Is it supposed to be used with https://github.com/microsoft/react-native-macos ?
If so, is it a general practice to reuse existing react-native libraries in `react-native-macos` (as opposed to forking)? Can you point to existing libraries that have native dependencies and are usable with both `react-native` and `react-native-macos`?

saghul:
Before this lands we should probably switch to XCFrameworks so we can bundle all architectures.

saghul:
I think there might be some confusion here. There is no forking going on. This PR bundles another WebRTC framework, which we don't want to do. We should build an XCFramework which can accommodate all Apple platforms in a single bundle. Then there are a couple of ifdef adjustments to make this plugin work on react-native-macos, that's all.

saghul:
Merged into master with some slight variations. We now use an XCFramework with 3 slices: iOS (ARM64), iOS (x86_64) and macOS (x86_64).

PR #872, opened 2020-09-14
Update code style to look similar to another class

suraneti:
Update code style to look similar to another class, they don't have an empty line in classes.


saghul:
Not a fan of changes which just fix style, it makes "git archaeology" harder. We'll fix it when working in that file.

PR #883, opened 2020-10-17
Fix Build webrtc using xcode 12

hthetiot:
Since XCode 12 update, it look like by default warnings are treated as errors using clang.

To fix that I confirmed that adding "treat_warnings_as_errors=false" to GN_COMMON_ARGS  does fix the issue to build WebRTC from M69 to M87 for ios and android.

I also upated BuildingWebRTC.md doc to provide reference link to the WebRTC branch names.

Reference:
- https://stackoverflow.com/questions/64111205/enabling-bitcode-for-webrtc-ios
- https://bugs.chromium.org/p/webrtc/issues/detail?id=11729

### DIFF
--- a/Documentation/BuildingWebRTC.md
+++ b/Documentation/BuildingWebRTC.md
@@ -46,6 +46,9 @@ cd
 
 Now the code is ready for building!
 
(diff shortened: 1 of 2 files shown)


saghul:
Thank you!

PR #886, opened 2020-10-22
Feature/add get display media

tmoldovan8x8:


MediaStreamTrack.js
```diff
@@ -87,7 +87,6 @@ class MediaStreamTrack extends EventTarget(MEDIA_STREAM_TRACK_EVENTS) {
   }
 
   applyConstraints() {
-    throw new Error('Not implemented.');
```
saghul, line 90: Please revert this change.

android/src/main/java/com/oney/WebRTCModule/AbstractVideoCaptureController.java
```diff
@@ -0,0 +1,57 @@
+package com.oney.WebRTCModule;
+
+import org.webrtc.VideoCapturer;
+
+public abstract class AbstractVideoCaptureController {
```
saghul, line 5: Very nice!

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -59,6 +77,24 @@
...
+
+        reactContext.addActivityEventListener(new BaseActivityEventListener() {
+            @RequiresApi(api = Build.VERSION_CODES.N)
+            @Override
+            public void onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
+                super.onActivityResult(activity, requestCode, resultCode, data);
+                if (requestCode == PERMISSION_REQUEST_CODE) {
+                    if (resultCode != Activity.RESULT_OK) {
+                        promise.reject("DOMException", "NotAllowedError");
+                        promise = null;
```
saghul, line 89: Is it necessary to nullify the promise here?

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -159,10 +159,11 @@ MediaStreamTrack getTrack(String id) {
      * if audio permission was not granted, there will be no "audio" key in
      * the constraints map.
      */
+    @RequiresApi(api = Build.VERSION_CODES.N)
```
saghul, line 162: Why is this necessary? We currently support (and want to keep doing so) M.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -241,9 +227,137 @@ void disposeTrack(String id) {
...
+            return;
+        }
+
+        Activity currentActivity = this.reactContext.getCurrentActivity();
+        if (currentActivity == null) {
+            promise.reject(new RuntimeException("No current Activity."));
+            return;
+        }
+
+        this.promise = promise;
```
saghul, line 248: Can we call this something less generic than "promise" so we know it's related to getDisplayMedia please?

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -241,9 +227,137 @@ void disposeTrack(String id) {
...
+
+        MediaProjectionManager mediaProjectionManager =
+            (MediaProjectionManager) currentActivity.getApplication().getSystemService(
+                Context.MEDIA_PROJECTION_SERVICE);
+        currentActivity.startActivityForResult(
+            mediaProjectionManager.createScreenCaptureIntent(), PERMISSION_REQUEST_CODE);
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.N)
```
saghul, line 257: What about M? O even L? We should return an error at least.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -241,9 +227,137 @@ void disposeTrack(String id) {
...
+
+        successCallback.accept(streamId, tracksInfo);
+    }
+
+    private VideoTrack createScreenTrack() {
+        DisplayMetrics displayMetrics = getDisplayMetrics();
+        int width = displayMetrics.widthPixels;
+        int height = displayMetrics.heightPixels;
+        int fps = 30;
+        ScreenVideoCaptureController videoCaptureController = new ScreenVideoCaptureController(width, height, fps, mediaProjectionPermissionResultData);
```
saghul, line 316: ScreenCaptureController?

android/src/main/java/com/oney/WebRTCModule/ScreenVideoCaptureController.java
```diff
@@ -0,0 +1,42 @@
+package com.oney.WebRTCModule;
+
+import android.content.Intent;
+import android.media.projection.MediaProjection;
+import android.util.Log;
+
+import org.webrtc.ScreenCapturerAndroid;
+import org.webrtc.VideoCapturer;
+
+public class ScreenVideoCaptureController extends AbstractVideoCaptureController {
```
saghul, line 10: ScreenCaptureController?

android/src/main/java/com/oney/WebRTCModule/ScreenVideoCaptureController.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    protected VideoCapturer createVideoCapturer() {
+        VideoCapturer videoCapturer = new ScreenCapturerAndroid(
+            mediaProjectionPermissionResultData,
+            new MediaProjection.Callback() {
+                @Override
+                public void onStop() {
+                    Log.w(TAG, "User revoked permission to capture the screen.");
```
saghul, line 35: How can we better handle this error?

getDisplayMedia.js
```diff
@@ -0,0 +1,37 @@
...
+import { Platform, NativeModules } from 'react-native';
+
+import MediaStream from './MediaStream';
+import MediaStreamError from './MediaStreamError';
+
+const { WebRTCModule } = NativeModules;
+
+export default function getDisplayMedia(constraints) {
+    if (Platform.OS !== 'android') {
+        return Promise.reject(new TypeError());
```
saghul, line 12:
```suggestion
        return Promise.reject(new Error('Unsupported platform'));
```

saghul (changes requested):
Left some comments, PTAL!

android/src/main/java/com/oney/WebRTCModule/ScreenVideoCaptureController.java
tmoldovan8x8: Actually, this is called when the MediaProjection is stopped, I've changed the logging message. From what I can tell, this can only be stopped by manually calling the stop method on it.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
tmoldovan8x8: It's necessary for the logic in getDisplayMedia(Promise promise) to work, where we verify that there isn't another task in progress.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -241,11 +221,137 @@ void disposeTrack(String id) {
...
         TrackPrivate track = tracks.get(trackId);
-        if (track != null && track.videoCaptureController != null) {
-            track.videoCaptureController.switchCamera();
+        if (track != null && track.videoCaptureController instanceof CameraCaptureController) {
+            CameraCaptureController cameraCaptureController = (CameraCaptureController) track.videoCaptureController;
+            cameraCaptureController.switchCamera();
         }
     }
 
+    public void getDisplayMedia(Promise promise) {
```
saghul, line 230:
```suggestion
    void getDisplayMedia(Promise promise) {
```

Issue #797, opened 2020-06-02
RTCSessionDescription constructor inconsistent behavior


sarumjanuch:
Calling RTCSessionDescription constructor fails, if called with no argument, which is not consistent with current browser behavior. 

#### Expected behavior
new RTCSessionDescription()
RTCSessionDescription {type: null, sdp: null}

#### Observerd behavior
new RTCSessionDescription()
TypeError: Cannot read property 'sdp' of undefined

#### Steps to reproduce the problem
Call new RTCSessionDescription()

#### Platform information
It is in .js root shims.

PR #887, opened 2020-10-22
Added an empty object as the default parameter for RTCSessionDescription

igor-peregudov:
Calling RTCSessionDescription constructor fails, if called with no argument, which is not consistent with current browser behavior.

https://github.com/react-native-webrtc/react-native-webrtc/issues/797

### DIFF
--- a/RTCSessionDescription.js
+++ b/RTCSessionDescription.js
@@ -4,7 +4,7 @@ export default class RTCSessionDescription {
   sdp: string;
   type: string;
 
-  constructor(info) {
+  constructor(info = {}) {
     this.sdp = info.sdp;
     this.type = info.type;
   }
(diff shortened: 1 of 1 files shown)


saghul:
On Firefox:

```
var sdp = new RTCSessionDescription();
Uncaught TypeError: Missing required 'type' member of RTCSessionDescriptionInit
    <anonymous> debugger eval code:1
```

saghul:
On Chrome `type` is `null` and `sdp` is `""`. I'd be ok having that behavior, but the current proposal will have both undefined, which no browser implements.

igor-peregudov:
> On Chrome `type` is `null` and `sdp` is `""`. I'd be ok having that behavior, but the current proposal will have both undefined, which no browser implements.

I corrected the commit

saghul (approved):

Issue #917, opened 2020-12-11
We don't handle `channelId` correctly


xilin:
#### Expected behavior
```
RTCPeerConnection pc = xxx;
const channel1 = pc.createDataChannel("label1", {maxRetransmits: 0, ordered: false});
channel1.onopen = xxx; // should be called
const channel2 = pc.createDataChannel("label2", {maxRetransmits: 0, ordered: false});
channel2.onopen = xxx; // should be called
```

#### Observerd behavior
Only first channel can receive state change event.

#### Steps to reproduce the problem
Just create multiple data channels in iOS, with react-native-webrtc >= 1.75.3

#### Platform information

* **React Native version**: 0.63.3
* **Plugin version**: 1.75.3
* **OS**: iOS
* **OS version**: 10

There is already a PR try to resolve this issue: https://github.com/react-native-webrtc/react-native-webrtc/pull/726 .
However it didn't fix the problem completely.
Based on the [discussion](https://react-native-webrtc.discourse.group/t/datachannel-id-is-1-and-no-callbacks-on-ios/323/4), I found that in the [PR](https://webrtc-review.googlesource.com/c/src/+/131381), aka commit [f3736ed](https://github.com/jitsi/webrtc/commit/f3736ed), those guys introduced checking for `negotiated`. If `negotiated  == false`, id will be reset to -1. And in the following method [AllocateSid](https://github.com/jitsi/webrtc/blob/f3736ed3d8d4e6660ad8b82a8deff7aca8389e5c/pc/data_channel.cc#L58),  it will try to run `potential_sid += 2;`.
So when `- (void)dataChannelDidChangeState:(RTCDataChannel*)channel` is called, `channel.channelId` had been changed, and different from what we recorded in `peerConnection.dataChannels` which is done in `createDataChannel:(nonnull NSNumber *)peerConnectionId
                              label:(NSString *)label
                             config:(RTCDataChannelConfiguration *)config`.
This `negotiated` logic still exists in the newest master branch of webrtc.

Previously, our team used `react-native-webrtc@1.69.2`, which rely on `[M69](https://github.com/jitsi/webrtc/tree/cb536cf7a368e77ec29a6779de7fbebf2c300b70)`,  checking the `datachannel.cc` in that [commit](https://github.com/jitsi/webrtc/blob/cb536cf7a368e77ec29a6779de7fbebf2c300b70/pc/datachannel.cc), there is not `negotiated` logic, so everything is happy back on that time.

[...]

PR #919, opened 2020-12-21
Fix the logic of `dataChannelId` in iOS

xilin:
See discussions here: https://github.com/react-native-webrtc/react-native-webrtc/issues/917
TLDR. webRTC may change passed `dataChannelId` into `potential_sid += 2`, so that in delegate methods `channelId` will be different from JS side.

I cannot compile the `ios/RCTWebRTC.xcodeproj` project. Just tested it in the other project.

### DIFF
--- a/ios/RCTWebRTC/WebRTCModule+RTCDataChannel.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCDataChannel.m
@@ -19,6 +19,22 @@
   objc_setAssociatedObject(self, @selector(peerConnectionId), peerConnectionId, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
 }
 
+/**
+ * webrtc will update channelId internally:
+ * https://github.com/jitsi/webrtc/blob/f3736ed3d8d4e6660ad8b82a8deff7aca8389e5c/pc/data_channel.cc#L39
+ * https://github.com/jitsi/webrtc/blob/f3736ed3d8d4e6660ad8b82a8deff7aca8389e5c/pc/data_channel.cc#L58
+ * so with this property we can keep channelId passed from JS side
+ */
+- (NSNumber *)originDataChannelId
+{
+  return objc_getAssociatedObject(self, _cmd);
+}
+
+- (void)setOriginDataChannelId:(NSNumber *)originDataChannelId
+{
+  objc_setAssociatedObject(self, @selector(originDataChannelId), originDataChannelId, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+}
+
 @end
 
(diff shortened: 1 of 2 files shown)


saghul:
LGTM, maybe you can link to the code inside WebRTC which allocates the ID so we can easily reference it?

xilin:
Here is where the ID is reset
https://github.com/jitsi/webrtc/blob/f3736ed3d8d4e6660ad8b82a8deff7aca8389e5c/pc/data_channel.cc#L39

Here is where WebRTC generates the new id
https://github.com/jitsi/webrtc/blob/f3736ed3d8d4e6660ad8b82a8deff7aca8389e5c/pc/data_channel.cc#L58

saghul:
Let's put it in the code please, so it can be seen in context.

xilin:
Updated.

ios/RCTWebRTC/WebRTCModule+RTCDataChannel.m
```diff
@@ -33,6 +49,7 @@ @implementation WebRTCModule (RTCDataChannel)
       || dataChannel.readyState == RTCDataChannelStateOpen)) {
     dataChannel.peerConnectionId = peerConnectionId;
     NSNumber *dataChannelId = [NSNumber numberWithInteger:config.channelId];
+    dataChannel.originDataChannelId = dataChannelId;
```
saghul, line 52: What happens if no channelId was passed when creating the data channel? Won't this be null?

saghul (commented):
Left a question.

ios/RCTWebRTC/WebRTCModule+RTCDataChannel.m
xilin: See [here](https://github.com/react-native-webrtc/react-native-webrtc/blob/master/RTCPeerConnection.js#L377), we'll always make sure passing a config with id.
At the same time, `config.channelId` is in type `int`, which always has value instead of a Class type.

xilin: > What happens if no channelId was passed when creating the data channel? Won't this be null?

See [here](https://github.com/react-native-webrtc/react-native-webrtc/blob/master/RTCPeerConnection.js#L377), we'll always make sure passing a config with id.
At the same time, `config.channelId` is in type `int`, which always has value instead of a Class type.

xilin: @saghul Any comment on this?

saghul (approved):

saghul:
Sorry for the delay @xilin , landed!

PR #949, opened 2021-03-02
android: handle null values in mandatory constraints

aj-65:
My team sends empty/null values in audio constraints to test device capabilities. This causes a `NullPointerException` on Android but no problems on iOS. 

This PR is checks for null values in `mandatory` audio constraints.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
+++ b/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
@@ -122,7 +122,7 @@ class GetUserMediaImpl {
             if (constraint.getValue() != null) {
                 valid.add(constraint);
             } else {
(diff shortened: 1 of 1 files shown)


saghul:
Why don't we just filter those values instead?

aj-65:
> Why don't we just filter those values instead?

I don't understand what you mean by filtering. 

Do you mean not sending empty audio constraints? If so, my team's library does that to test audio. Also, it works without crashing on iOS. So I wanted Android to be able to handle the same use case.

saghul:
What I mean is that instead of failing to create an audio track we could just ignore the bogus values and create the audio track with the valid ones. Postel's principle and all :-)

aj-65:
> What I mean is that instead of failing to create an audio track we could just ignore the bogus values and create the audio track with the valid ones. Postel's principle and all :-)

@saghul Updated the PR to filter nulls

saghul (approved):

PR #970, opened 2021-04-16
ios: apply image orientation on getDisplayMedia implementation

saghul:


ios/RCTWebRTC/ScreenCapturer.m
```diff
@@ -186,20 +188,38 @@ - (void)readBytesFromStream:(NSInputStream *)stream {
     }
 }
 
-- (void)didCaptureVideoFrame:(CVPixelBufferRef)pixelBuffer {
+- (void)didCaptureVideoFrame:(Message *)message {
```
alexbumbu, line 191: I would add the orientation as a method param. It's more objc style this way.

saghul: Pass the pixel buffer and orientation as 2 parameters? Sure, can do!

alexbumbu (approved):

PR #972, opened 2021-04-21
Fix crash when originDataChannelId is nil

### LINKED ISSUE #962: [iOS] dataChannelDidChangeState crash
Hello, after this PR https://github.com/react-native-webrtc/react-native-webrtc/pull/919 (issue https://github.com/react-native-webrtc/react-native-webrtc/issues/917) I have a crash when a connection is made. Specifically the crash occurs in `dataChannelDidChangeState` in `RTCDataChannel.m` : a `NSDictionnary` is created with the channel id but `channel.originDataChannelId` can be `null`.
Note that it doesn't always happen so it looks like a race condition.

#### Expected behavior

Connecting to a peer should not make the app crash randomly on iOS.

#### Platform information

* **React Native version**: 0.63.4
* **Plugin version**: 1.89.1
* **OS**: iOS
* **OS version**: 14

gerhardberger:
fixes #962

This PR aims to aid the issue that due to some race condition most likely, in some cases (mostly in simulators) datachannel creation can fail because `originDataChannelId` is set to nil and we want to access it, causing a crash. To solve this, in cases like this we just default the id to 1.

### DIFF
--- a/ios/RCTWebRTC/WebRTCModule+RTCDataChannel.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCDataChannel.m
@@ -95,8 +95,10 @@ RCT_EXPORT_METHOD(dataChannelSend:(nonnull NSNumber *)peerConnectionId
 // Called when the data channel state has changed.
 - (void)dataChannelDidChangeState:(RTCDataChannel*)channel
 {
+  NSNumber *id = channel.originDataChannelId ?: @1;
+
   // use originDataChannelId instead of channelId to keep the consistency with JS side
-  NSDictionary *event = @{@"id": channel.originDataChannelId,
+  NSDictionary *event = @{@"id": id,
                           @"peerConnectionId": channel.peerConnectionId,
                           @"state": [self stringForDataChannelState:channel.readyState]};
   [self sendEventWithName:kEventDataChannelStateChanged body:event];
@@ -120,8 +122,11 @@ RCT_EXPORT_METHOD(dataChannelSend:(nonnull NSNumber *)peerConnectionId
     data = [[NSString alloc] initWithData:buffer.data
                                  encoding:NSUTF8StringEncoding];
   }
+
+  NSNumber *id = channel.originDataChannelId ?: @1;
+
   // use originDataChannelId instead of channelId to keep the consistency with JS side
-  NSDictionary *event = @{@"id": channel.originDataChannelId,
+  NSDictionary *event = @{@"id": id,
                           @"peerConnectionId": channel.peerConnectionId,
                           @"type": type,
                           // XXX NSDictionary will crash the process upon


saghul:
I don't like fixing things like this, without clearly understanding why it's nil sometimes.

ibc:
While I agree is good to really understand the issue, I'd say that the PR that produced this bug is pure sorcery: https://github.com/react-native-webrtc/react-native-webrtc/pull/919/files. Does anyone know what it's supposed to do?

So if I understand the core problem properly:

1. We call `const dc = pc.createDataChannel('foo', options)` in JS.
2. Such a method is synchronous so we must return a `RTCDataChannel` JS instance here, and it must contain a proper `id`.
3. And here the problem since, if `options.id` is not given, then native libwebrtc side will choose an available one (this will happen **eventually**, see below), so if we try to read  `dc.id` in JS after creating `dc`, it will return `undefined` or `null`.

Some concerns about this.

It's **NOT** true that native libwebrtc will always set an `id` (if not given in `options`) immediately. See what the [spec says about it](https://www.w3.org/TR/webrtc/#dom-rtcdatachannel-id):

> `id` of type unsigned short, readonly, nullable
>
> The id attribute returns the ID for this RTCDataChannel. The value is initially null, which is what will be returned if the ID was not provided at channel creation time, and the DTLS role of the SCTP transport has not yet been negotiated. Otherwise, it will return the ID that was either selected by the script or generated by the user agent according to [RFC8832]. After the ID is set to a non-null value, it will not change. On getting, the attribute MUST return the value of the [[DataChannelId]] slot.

Notice it says "initially null [...] if DTLS role has not yet been negotiated", which is what happens in this [sample code](https://github.com/react-native-webrtc/react-native-webrtc/issues/962#issuecomment-824004721) that reproduces the crash. Just wondering if this may be related to the issue, the fact that native `id` is **still** null for a while until other things happen. However, my understanding is that libwebrtc never sets it to null.

However we need `id` in both JS and native to hold the references of the DataChannel objects in a map indexed by those `ids`. From my experience, if `A` tells `B` to create a `C` object, and `A` also creates a representation of such a `C` object, it's `A` the one who must provide the `id` to `B` and not the other way around.

So my proposal is:

- Make `pc.createDataChannel()` pass a random/uuid/whatever `internalId` field to native.
- Hold created DataChannels objects in maps indexed by such an `internalId` in both JS and native.
- In native we may need a second map to match native `RTCDataChannel` instances (i.e. in `dataChannelDidChangeState` callback) to its associated `internalId`, so we can fill the event properly by including `internalId` into it.

If this is ok, the only missing piece is: how to tell JS about the real `dataChannel.id` (when `options.id` was not given)? Well, once the native `RTCDataChannel` is created in native side (and assuming that it's `id` is not null as the spec may suggest), native side may just send a custom `gotDataChannelId` event to JS. Such an event would contain both `internalId` (so JS side can obtain the target instance) and the real `id`, and then JS side updates its `id`.

saghul:
> However, my understanding is that libwebrtc never sets it to null.

The default value in WebRTC (the config object at least) is -1, aka unset.

Note that in JS we cheat and always pass an ID: https://github.com/react-native-webrtc/react-native-webrtc/blob/master/RTCPeerConnection.js#L377-L408

ibc:
Ok, so you know that `this. _dataChannelIds` is **never** never being cleared, right? I mean, when a DataChannel is closed its `id` should be removed from `this._dataChannelIds`. It's not right now.

Second,  for (id = 1; id < 65535 && dataChannelIds.has(id); ++id);` will produce `id = 65535` if `ids` are exhausted. Also, libwebrtc just accepts `ids` in 1..1024 range, no matter what the spec says.

Anyway, even with these errors, it looks like `createDataChannel()` is never sending `id: undefined` to native side. If that happened the crash would happen, right?

And especially: if JS is always passing an `id`, why does native side need any `originalId` sorcery at all? It does not make any sense IMHO.

ibc:
> Ok, so you know that `this. _dataChannelIds` is **never** never being cleared, right? I mean, when a DataChannel is closed its `id` should be removed from `this._dataChannelIds`. It's not right now.

The JS `RTCDataChannel` instance should emit a private 'closed' event to its `RTCPeerConnection` parent and the latter should remove its `id`.

xilin:
> Ok, so you know that `this. _dataChannelIds` is **never** never being cleared, right? I mean, when a DataChannel is closed its `id` should be removed from `this._dataChannelIds`. It's not right now.
> 
> Second, `for (id = 1; id < 65535 && dataChannelIds.has(id); ++id);` will produce `id = 65535` if `ids` are exhausted. Also, libwebrtc just accepts `ids` in 1..1024 range, no matter what the spec says.
> 
> Anyway, even with these errors, it looks like `createDataChannel()` is never sending `id: undefined` to native side. If that happened the crash would happen, right?
> 
> And especially: if JS is always passing an `id`, why does native side need any `originalId` sorcery at all? It does not make any sense IMHO.

The reason is [here](https://github.com/react-native-webrtc/react-native-webrtc/issues/917). `id` from JS side will be override in native side.

saghul:
Oops, forgot to close this one. Fixed in https://github.com/react-native-webrtc/react-native-webrtc/commit/1fbe7e14bf540e1651c98ee11fc11f690f61f622

PR #982, opened 2021-04-27
adds dockerFile for android build

tmoldovan8x8:


tools/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM debian:stable-slim
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+#This is required fo bypassing the tzdata installation
+ENV TZ=Europe/Bucharest
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
```
danjenkins, line 7: Any reason you did all these RUN's as separate lines? its better to have them all on one line if possible purely from a docker image size 

And you might want to remove the apt-repo list once youve updated

tools/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN apt-get install -y lsb-release
+RUN apt-get install -y vim
+RUN apt-get install -y ninja-build
+RUN apt-get install -y pkg-config
+RUN apt-get install -y gcc-arm-linux-gnueabihf
+RUN apt-get install -y g++-8-arm-linux-gnueabihf
+RUN apt-get install -y ca-certificates 
+
+#Install JAVA
+RUN mkdir -p /usr/share/man/man1
```
danjenkins, line 23: I'd change the workdir and then do all the run commands together

tools/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+WORKDIR /opt
+RUN curl -L -o openjdk.tar.gz https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz
+RUN mkdir jdk
+RUN tar zxf openjdk.tar.gz -C jdk --strip-components=1
+RUN ln -sf /opt/jdk/bin/* /usr/local/bin/
+
+# clone react-native-webrtc repository
+WORKDIR /
+RUN git clone https://github.com/react-native-webrtc/react-native-webrtc.git
```
danjenkins, line 32: This will only allow you to grab master.... I'd make the org/repo and then the SHA1 arguments that get substituted in so that someone can easily override it in a fork/testing on a different branch... if the fork/branch isnt public thats a problem they would have to overcome.

I get it, you're not aiming for automated builds of libwebrtc with this docker image.... you're using it as an image you can exec into and then run the commands you want.... I still think this would be nice, and would inherently allow for automated builds...

I'd probably go and clone down jitsi/webrtc (configurable) so you can point your tools command to that directory

saghul: This is Titus' first ever Dockerfile, go easy on him ;-)

Currently the plan is to just setup the environment, run the setup phase and add the jitsi/webrtc origin.

Then building the library is a couple of commands away.

Let's see how this plays out, I'll be happy to see it extended further after we lay down the first stone.

danjenkins: I didn't mean to come across harshly :)  sorry if it did!

tools/Dockerfile
tmoldovan8x8: Makes sense, I was not aware about this image size thing.  Now I've tried to group them a bit, so that the code still remains readable.

tools/Dockerfile
tmoldovan8x8: Not sure if this is possible here, since all the following commands need to run in a different workdir

tools/Dockerfile
tmoldovan8x8: :) It's fine. I actually thought about passing the branch for react-native-webrtc, but not sure if that is worth it, since that one is used only for pulling the scripts, on which not a lot of further work is required. For the jitsi/webrtc repo, that would definitely be the case, but so far I was not able to find a solution to add those commands to the docker file. If I try to do any of the setup steps here, crazy permissions and other errors are thrown.

tools/Dockerfile
danjenkins: You could move this into the above command if you wanted to remove a layer.
```suggestion
RUN apt-get update \
	&& apt-get install -y sudo \
				git \
				python \
				curl \
				lsb-release \
				vim \
				ninja-build \
				pkg-config \
				gcc-arm-linux-gnueabihf \
				g++-8-arm-linux-gnueabihf \
				ca-certificates \
	&& apt-get clean && mkdir -p /usr/share/man/man1
```

tools/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+				ninja-build \
+				pkg-config \
+				gcc-arm-linux-gnueabihf \
+				g++-8-arm-linux-gnueabihf \
+				ca-certificates \
+	&& apt-get clean
+
+#Install JAVA
+RUN mkdir -p /usr/share/man/man1
+WORKDIR /opt
```
danjenkins, line 26: `WORKDIR` is only really useful when you've got multiple dockerfile lines doing something.... I'd recommend just removing your `WORKDIR /opt` and changing your RUN to do a `cd`
```suggestion
RUN cd /opt && curl -L -o openjdk.tar.gz https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz \
	&& mkdir jdk \
	&& tar zxf openjdk.tar.gz -C jdk --strip-components=1 \
	&& ln -sf /opt/jdk/bin/* /usr/local/bin/
```

tools/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN curl -L -o openjdk.tar.gz https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz \
+	&& mkdir jdk \
+	&& tar zxf openjdk.tar.gz -C jdk --strip-components=1 \
+	&& ln -sf /opt/jdk/bin/* /usr/local/bin/
+
+# clone react-native-webrtc repository
+WORKDIR /
```
danjenkins, line 33: Same for this one.... I'd just move it into the git clone... 
```suggestion
RUN cd / && git clone https://github.com/react-native-webrtc/react-native-webrtc.git
```

danjenkins (changes requested):

PR #997, opened 2021-05-09
setup `originDataChannelId` for dataChannel created in RTCPeerConnection

xilin:
When a new dataChannel is created from this call stack:

image

https://github.com/jitsi/webrtc/blob/master/pc/data_channel_controller.cc#L259
https://github.com/jitsi/webrtc/blob/master/pc/data_channel_controller.cc#L327
https://github.com/jitsi/webrtc/blob/master/sdk/objc/api/peerconnection/RTCPeerConnection.mm#L161
https://github.com/react-native-webrtc/react-native-webrtc/blob/master/ios/RCTWebRTC/WebRTCModule%2BRTCPeerConnection.m#L543

We need to also setup `originDataChannelId` for it.

Same as we have done in android:
image

More details are [here](https://github.com/react-native-webrtc/react-native-webrtc/issues/962#issuecomment-835842895).

### DIFF
--- a/ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
@@ -542,6 +542,7 @@ RCT_EXPORT_METHOD(peerConnectionGetStats:(nonnull NSNumber *) objectID
 
   NSNumber *dataChannelId = [NSNumber numberWithInteger:dataChannel.channelId];
   dataChannel.peerConnectionId = peerConnection.reactTag;
+  dataChannel.originDataChannelId = dataChannelId;
   peerConnection.dataChannels[dataChannelId] = dataChannel;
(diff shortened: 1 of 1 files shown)


saghul:
As I mentioned in the original issue the current entire approach is flawed really. The id must remain unset until negotiated, if the negotiated flag is not specified.

ibc:
> @ibc could you try this fix to see if your case is fixed?

I'm afraid I don't have a test environment to check it.

ibc:
> This problem is solved, too, thanks.
> 
> 
> 
> Users of the 'patch-package' module can use this file until they update it
> 
> 
> 
> [react-native-webrtc+1.89.1.patch.txt](https://github.com/react-native-webrtc/react-native-webrtc/files/6451842/react-native-webrtc%2B1.89.1.patch.txt)
> 
> 

Please read the comment above made by @saghul

PR #1013, opened 2021-05-26
Created Expo config plugin

EvanBacon:
# Why

> **Note** I've updated the package to support Expo SDK 46 and React Native 69. This no longer requires `react-native-webrtc` to depend on `@expo/config-plugins` directly. Users can continue to use the [out-of-tree plugin](https://www.npmjs.com/package/@config-plugins/react-native-webrtc) while we wait for this PR to be merged.

With Expo SDK 41 we've rolled out an interface called [config plugins](https://docs.expo.io/guides/config-plugins/) which lets users add native modules that aren't in the Expo Go app to their native cloud builds and locally when [prebuilding](https://docs.expo.dev/workflow/prebuild).

This is a [highly requested package](https://expo.canny.io/feature-requests/p/webrtc) so I've created the plugin personally.

# How

- Added a `app.plugin.js` as the main entry point to the plugin. Plugins must run in at node LTS environments (currently that means 14 and greater).

# Test Plan

- Run `npm pack` in the package
- Install the package in a managed Expo project `yarn add react-native-webrtc@../react-native-webrtc/react-native-webrtc-1.89.1.tgz`
- Then add `react-native-webrtc` to the plugins array and build the native app locally with `expo prebuild` and `yarn ios`, `yarn android`. We plan to further automate these steps with `expo install` and `expo run` commands.
- App built correctly without any errors.
  - Running `expo config --type introspect` showed all of the desired config changes in AndroidManifest, Info.plist, and entitlements files.

Modified some values and in the config plugin and updated with `expo prebuild` and `yarn ios`, `yarn android`.


saghul:
Looking good! I’m on leave ATM but will review this once I’m back!

EvanBacon:
People can use this out-of-tree solution in the meantime [`@config-plugins/react-native-webrtc`](https://github.com/EvanBacon/prototype-config-plugins/tree/master/packages/react-native-webrtc). It's not versioned with the code so YMMV over time, but for now it works great! We also have one for callkeep.

package.json
```diff
@@ -17,16 +17,24 @@
...
+    "expo-module-scripts": "^2.0.0"
+  },
   "peerDependencies": {
     "react-native": ">=0.60.0"
   },
   "scripts": {
+    "build:plugin": "tsc --build plugin",
+    "clean:plugin": "expo-module clean plugin",
+    "lint:plugin": "eslint plugin/src/*",
+    "prepare": "yarn clean:plugin && yarn build:plugin",
```
saghul, line 37: Please don't use `yarn` here, not everyone is using it and there is no reason to force users to do so.

saghul (commented):
Sorry for the delay, I finally managed to give this a try. I left a single comment, but after giving this a try I'm honestly not very inclined to merge it.

After this change, we have a whopping 2800 packages as dependencies 😱 (check with `npm ls | wc -l) vs the 12 the project has today.

Now, I do want to facilitate the lives of those using Expo, and AFAIS they just need to use the package you created (https://github.com/Expo/config-plugins/tree/master/packages/react-native-webrtc) right? What problems do you foresee with that approach? The manifest items the package takes care of have not changed in years, so I'm not concerned about hypothetical incompatibilities.

Thoughts?

EvanBacon:
@saghul the dependency you're concerned about has many dependencies in common with a standard React Native project so it's just accessing the code that's already in the node modules folder. 

We can stick with the `@config-plugins/react-native-webrtc` package for a while, but it would be good to have it versioned somehow, and we should also land the markdown changes so users have clear documentation on how to get things working.

Further, I've noticed that the config plugin doesn't fully work with package due to a couple of issues. The main one being that Bitcode is not available on iOS, causing the binary to not be fully compilable in Release mode: 

```
❌  ld: '/Users/expo/Library/Developer/Xcode/DerivedData/CustomGo-ghgjpyndamtfsoeycnvxcexmmzsi/Build/Intermediates.noindex/ArchiveIntermediates/CustomGo/BuildProductsPath/Release-iphoneos/XCFrameworkIntermediates/WebRTC/WebRTC.framework/WebRTC' does not contain bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE), obtain an updated library from the vendor, or disable bitcode for this target. file '/Users/expo/Library/Developer/Xcode/DerivedData/CustomGo-ghgjpyndamtfsoeycnvxcexmmzsi/Build/Intermediates.noindex/ArchiveIntermediates/CustomGo/BuildProductsPath/Release-iphoneos/XCFrameworkIntermediates/WebRTC/WebRTC.framework/WebRTC' for architecture arm64
```

I guess running [this script](https://github.com/react-native-webrtc/react-native-webrtc/blob/master/tools/downloadBitcode.sh) fixes that, but we've spec'd config plugins to avoid running fetch requests. Generally all of the native code is shipped in the node module. I'd love to get this fixed, so let me know if you have any ideas for when this should run.

Another issue is that this module often causes the yarn.lock/Podfile.lock to get corrupted, requiring the lock files to be nuked.

saghul:
> @saghul the dependency you're concerned about has many dependencies in common with a standard React Native project so it's just accessing the code that's already in the node modules folder.

Good point, I'll check it out again.

> The main one being that Bitcode is not available on iOS, causing the binary to not be fully compilable in Release mode:

Does Expo default to using bitcode? Why? It's not mandatory and the advantages are not a clean cut IMHO (Jitsi Meet is built with bitcode, but have no other choice because we have an Apple Watch target). The WebRTC framework with bitcode is ~1GB uncompressed...

> I guess running [this script](https://github.com/react-native-webrtc/react-native-webrtc/blob/master/tools/downloadBitcode.sh) fixes that, but we've spec'd config plugins to avoid running fetch requests. Generally all of the native code is shipped in the node module. I'd love to get this fixed, so let me know if you have any ideas for when this should run.

As I mentioned, the framework is just huge, I don't see a way around it TBH. Well yes, defaulting to no bitcode :-)



> Another issue is that this module often causes the yarn.lock/Podfile.lock to get corrupted, requiring the lock files to be nuked.

I have never run into this myself, can you show an example? Podfile thrashing is usually due to a mismatched CocoaPods version. Can't speak about yarn since I don't use it.

package.json
giautm: This script will auto run when you publish packages. So, just change it to `npm run` to fit your use-case. User not need to run this/depends on what the user use.

jesse-savary:
@saghul is there any way that the bitcode could be shipped as a separate NPM package? If it's not possible to ship an optional/addon package perhaps a `react-native-webrtc-heavy` could be created as an alternative, similar to what `react-native-v8 does: https://github.com/Kudo/react-native-v8/#v8-variants

8BallBomBom:
@EvanBacon Looking good 👍🏻
Thanks for your contribution.

We will be looking to move most of the documentation over to [here](https://react-native-webrtc.github.io/handbook/) eventually.
Can i include the documentation changes you've made once your pr has been merged?

8BallBomBom:
Needs updating to reflect the bitcode removal but maybe include in the next stable update?

8BallBomBom:
Never really used Expo but i can definitely test the pr and ensure things are up to scratch for inclusion 👍🏻

8BallBomBom:
Reopened to be checked at a later period 👍🏻

PR #1032, opened 2021-07-05
feat: implement clone on MediaStreamTrack and MediaStream

### LINKED ISSUE #1030: feat: implement clone on MediaStreamTrack and MediaStream
Hey @saghul 
Would it be possible to have clone implemented on the MediaStream and MediaStreamTrack?

## current situation
current implementation is to throw an error that it isn't implemented. 

Referencing MDN for context. 
https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/clone
https://developer.mozilla.org/en-US/docs/Web/API/MediaStream/clone

See also
https://w3c.github.io/mediacapture-main/#mediastreamtrack
https://w3c.github.io/mediacapture-main/#dom-mediastream-clone

### MediaStream
- [ ] 1. Let streamClone be a newly constructed MediaStream object.
- [ ] 2. Initialize streamClone.MediaStream.id to a newly generated value.
- [ ] 3. Clone each track in this MediaStream object and add the result to streamClone's track set.-
- [ ] 4. Return streamClone.

### MediaStreamTrack
- [ ] 1. Let track be the MediaStreamTrack object to be cloned.
- [ ] 2. Let trackClone be a newly constructed MediaStreamTrack object.
- [ ] 3. Initialize trackClone.id to a newly generated value.
- [ ] 4. Initialize trackClone's kind, label, readyState, and enabled attributes by copying the corresponding values from track.
- [ ] 5. Let trackClone's underlying source be the source of track.
- [ ] 6. Set trackClone's constraints to the active constrains of track.
- [ ] 7. Return trackClone.

## Proposed Solution
I think we can solve this is in two passes. If we relax requirement MediaStreamTrack.3 - which i believe is already the case because AFAIK the track.Id seems stable to me when copied across MediaStreams. We should be able to use the constructor methods to the bulk of the cloning. 

MediaStream.js
``` js
  clone() {
    return new MediaStream(this);
  }
```

MediaStreamTrack.js 
``` js
  clone() {
    return new MediaStreamTrack({
      id: this.id,
      kind: this.kind,
      label: this.label,
      muted: this.muted,
      remote: this.remote,
      constraints: this._constraints,
      enabled: this._enabled,
      settings: this._settings
    });
  }
```

and in the second pass implement the fact that track should act more like MediaStream in that the track.ids are unique but can be pointing to the same source. AFAIK, the track.id is the same id that's used to look up the PrivateTrack so making that unique would cause the id to lose the video source.

I wasn't sure how to tackle that one, because there are some things that are implemented up stream, so I wasn't sure the effect. implementing this in two parts should be low risk if anybody is assuming the track.id's don't change.  And should unblock us in the short term.

### The Why?
Twilio-video.js assumes that clone is implemented. So we need this feature and would be willing to work on it.

jahead:
Closes #1030

### DIFF
--- a/MediaStream.js
+++ b/MediaStream.js
@@ -118,7 +118,7 @@ export default class MediaStream extends EventTarget(MEDIA_STREAM_EVENTS) {
   }
 
   clone() {
-    throw new Error('Not implemented.');
+    return new MediaStream(this)
   }
 
   toURL() {
--- a/MediaStreamTrack.js
+++ b/MediaStreamTrack.js
@@ -50,8 +50,7 @@ class MediaStreamTrack extends EventTarget(MEDIA_STREAM_TRACK_EVENTS) {
     this.remote = info.remote;
 
     const _readyState = info.readyState.toLowerCase();
-    this.readyState = (_readyState === "initializing"
-      || _readyState === "live") ? "live" : "ended";
+    this.readyState = (_readyState === "initializing" || _readyState === "live") ? "live" : "ended";
   }
 
   get enabled(): boolean {
--- a/MediaStreamTrack.js
+++ b/MediaStreamTrack.js
@@ -51,7 +51,7 @@ class MediaStreamTrack extends EventTarget(MEDIA_STREAM_TRACK_EVENTS) {
 
     const _readyState = info.readyState.toLowerCase();
     this.readyState = (_readyState === "initializing"
-                    || _readyState === "live") ? "live" : "ended";
+      || _readyState === "live") ? "live" : "ended";
   }
 
   get enabled(): boolean {
@@ -95,7 +95,17 @@ class MediaStreamTrack extends EventTarget(MEDIA_STREAM_TRACK_EVENTS) {
   }
 
   clone() {
-    throw new Error('Not implemented.');
+    const info = {
+      constraints: this._constraints,
+      enabled: this._enabled,
+      settings: this._settings,
+      id: this.id = info.id,
+      kind: this.kind = info.kind,
+      label: this.label = info.label,
+      muted: this.muted = false,
+      remote: this.remote = info.remote,
+    }
+    return new MediaStreamTrack(info);
   }
 
   getCapabilities() {


saghul:
Hey there! I meant to reply to the original issue you reported, but I'm on leave and forgot.

I don't think this works. The underlying track (the native part) is not duplicated, so if a track is released, all copies are invalidated and a crash will occur.

Why exactly do you need this? I'd need to check, but I'm afraid the ObjC / Java APIs make implementing this very hard because a track and its source are not detachable.

jahead:
Hey, this is just a first pass, Thanks for the feedback

we use twilio-video.js which just calls the underlining webrtc apis in chrome, but we would like to have this work in an react-native environment. Allowing us to standardize our code base. 
Basically our upstream library uses this member property to do some of it's work. 

Is this only any issue for the MediaStreamTrack class?

It seems like 'copying' is supported by the mediaStream following this bit of code

``` js 
      } else if (arg instanceof MediaStream) {
          WebRTCModule.mediaStreamCreate(this.id);
          for (const track of arg.getTracks()) {
              this.addTrack(track);
          }
      }
```
Or maybe I'm reading that wrong. 
So for this to work we would need to only release on the last MediaStreamTrack copy?

jahead:
@saghul without spamming your inbox too much

Following the spec.

> When a MediaStreamTrack object ends for any reason (e.g., because the user rescinds the permission for the page to use the local camera, or because the application invoked the stop() method on the MediaStreamTrack object, or because the User Agent has instructed the track to end for any reason) it is said to be ended.
> 
> When a MediaStreamTrack track ends for any reason other than the stop() method being invoked, the User Agent MUST queue a task that runs the following steps:
> 
>1) If track.readyState has the value "ended" already, then abort these steps.
> 
>2) Set track.readyState to "ended".
> 
> 3) Notify track's source that track is ended so that the source may be stopped, unless other MediaStreamTrack objects depend on it.
> 
> 4) Fire an event named ended at the object.

There is already a construct in the (at least in the java side) for a PrivateTrack, it seems to me all we need is a to have one layer more of indirection, (n:1)  PublicTrack -> Private Track and realistically, that could be done at the java-script level,  Much like how the current MediaStream AFAIK works.

imo, if a PrivateTrack is ended then all the public tracks pointing to the same source should be ended too no?

saghul:
Sorry for the delay.

> we use twilio-video.js which just calls the underlining webrtc apis in chrome, but we would like to have this work in an react-native environment. Allowing us to standardize our code base.

Right. My question is more about twilio-video then. What is it they need that for? I have never needed it. That doesn't mean there is no good reason to have it, but I'm curious if there is a better way to implement equivalent functionality since as I'll mention below, I think this won't be trivial.

> Is this only any issue for the MediaStreamTrack class?

Yes.



> It seems like 'copying' is supported by the mediaStream following this bit of code
> 
> ```js
>       } else if (arg instanceof MediaStream) {
>           WebRTCModule.mediaStreamCreate(this.id);
>           for (const track of arg.getTracks()) {
>               this.addTrack(track);
>           }
>       }
> ```
> 
> Or maybe I'm reading that wrong.
> So for this to work we would need to only release on the last MediaStreamTrack copy?

Yep, this would work. You'd end up with a new stream what has **the same** tracks as the other one. I'm not sure if MediaStream.clone() is supposed to clone the tracks, in which case we cannot do that.



> There is already a construct in the (at least in the java side) for a PrivateTrack, it seems to me all we need is a to have one layer more of indirection, (n:1) PublicTrack -> Private Track and realistically, that could be done at the java-script level, Much like how the current MediaStream AFAIK works.
> 
> imo, if a PrivateTrack is ended then all the public tracks pointing to the same source should be ended too no?

Not quite. PrivateTrack holds some metadata we need, and things like the video capturer. Now, adding an extra level of indirection won't magically give you a new actual track piped to the same source, because in Java / ObjcC there is no API that I'm aware of to access a track's source. This becomes clearer if you think about remote tracks.

Maybe an interesting thing to do is dig into the Chromium source and check how they implement clone, and take it from there.

jahead:
Hey @saghul Sorry for the late reply on this, I've been pulled into doing some work native work, I'll respond with my findings on the weekend.

julien9999:
did you get Twilio video to work with this module?

PR #1045, opened 2021-07-27
Handle Android IllegalStateException in track.kind()

aj-65:
I have been getting this error while testing. It has been difficult to handle in the client app because `mediaStreamAddTrack` and `mediaStreamRemoveTrack` are executed on the `ExecutorService` thread. So, I think we should handle it here.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -533,7 +533,13 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
             return;
         }
 
-        String kind = track.kind();
+        String kind = null;
+        try{
+            kind =  track.kind();
+        }catch (IllegalStateException ex){
+            Log.e(TAG, "mediaStreamAddTrack() error", ex);
+        }
+
         if ("audio".equals(kind)) {
             stream.addTrack((AudioTrack)track);
         } else if ("video".equals(kind)) {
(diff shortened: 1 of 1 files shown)


saghul:
Hey @aj-65! I went through our Crahlytics and haven't seen any such crashes. That doesn't mean they don't happen, but I'm curious. Do you have a repro case? Any chance you are calling addTrack with a track which was already released?

aj-65:
Hi @saghul, hope you're doing well.

My team is using `react-native-webrtc` to support virtual classrooms. I am rendering the peers via:

```
val rtcView = WebRTCView(reactInstanceManager.currentReactContext)
val video = RTCVideoViewManager()
video.setStreamURL(rtcView, stream.streamId)
container.addView(rtcView)
```

If there is a better way, let me know please. But when a peer on a call disconnects, the app crashes due to an uncaught `IllegalStateException` in a non-main thread. This may be due the rendering after a track has been released (like you said). I can create a sample app that reproduces the error if needed.

Screen Shot 2021-07-27 at 1 00 06 PM


I have tested the code change in this PR with my local `react-native-webrtc` and it works as expected. That being said, let me know if I need to change my app design versus handling exceptions in `WebRTCModule`

artushin:
@saghul Hi there, just wanted to check if we can answer any questions about this change. Seems like this fix shouldn't have any side effects and it is blocking a release for us. Would you mind expediting a merge please?

saghul:
> val rtcView = WebRTCView(reactInstanceManager.currentReactContext)
> val video = RTCVideoViewManager()
> video.setStreamURL(rtcView, stream.streamId)
> container.addView(rtcView)

That is not JS, how / where are you running that code? Can you show more context? The plugin is not meant to be used that way, you have to call things from the right thread (the JS thread) or things will just blow up into pieces.

I don't want to land this PR as is because it encourages bad usage. You are dancing in undefined behavior territory and this is just the tip of the iceberg.

aj-65:
@saghul That code is Kotlin running on our native Android app. You made a good point about the plugin being designed for JS use. 
Closing this PR, I'll figure out another way to handle this error. Thanks!

PR #1055, opened 2021-08-24
getDisplayMedia api improvements

asprecic:
This PR removes the restriction that getDisplayMedia has to send a constraints object, in the web api it is completely optional, and in this codebase it is not used anywhere. I believe removing it will make it easier not to trip over these implementation details.

Likewise, getDisplayMedia is also added in registerGlobals under navigator.mediaDevices.getDisplayMedia, as it is in the web api. No reason to register getUserMedia but not getDisplayMedia.

Readme is also updated to state that getDisplayMedia gets exposed when registerGlobals is called.

### DIFF
--- a/getDisplayMedia.js
+++ b/getDisplayMedia.js
@@ -7,11 +7,7 @@ import MediaStreamError from './MediaStreamError';
 
 const { WebRTCModule } = NativeModules;
 
-export default function getDisplayMedia(constraints) {
-    if (!constraints || !constraints.video) {
-        return Promise.reject(new TypeError());
-    }
-
+export default function getDisplayMedia() {
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


getDisplayMedia.js
```diff
@@ -8,10 +8,6 @@ import MediaStreamError from './MediaStreamError';
 const { WebRTCModule } = NativeModules;
 
 export default function getDisplayMedia(constraints) {
```
saghul, line 10: Can you remove the (now unused) constraints argument too?

saghul (commented):
LGTM with a minor comment. Thanks for doing this!

getDisplayMedia.js
asprecic: Done.

saghul (approved):
LGTM, great job!

PR #1058, opened 2021-08-31
android: fix thread-safety issue in getStreamForReactTag

saghul:


android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
```diff
@@ -360,7 +339,7 @@ private void removeRendererFromVideoTrack() {
             // (e.g. addSink, removeSink) on videoTrack. It is OK to
             // skip the removeSink invocation in such a case because
             // VideoTrack#dispose() has performed it already.
-            VideoTrack videoTrack = getVideoTrack();
+            VideoTrack videoTrack = this.videoTrack;
 
             if (videoTrack != null) {
```
tmoldovan8x8, line 344: Actually here do we still need the extra variable and this if?

android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
```diff
@@ -550,13 +529,10 @@ public void setZOrder(int zOrder) {
      * all preconditions for the start of rendering are met.
      */
     private void tryAddRendererToVideoTrack() {
-        VideoTrack videoTrack;
+        VideoTrack videoTrack = this.videoTrack;
```
tmoldovan8x8, line 532: Same here

tmoldovan8x8 (approved):

Issue #1070, opened 2021-09-15
Android: Warnings after upgrading to RN 0.65.*


valentinmakov:
#### Expected behavior
No warnings fired when new RTCPeerConnection object is created

#### Observerd behavior
The following warnings appear after new RTCPeerConnection object is created:

`new NativeEventEmitter()` was called with a non-null argument without the required `addListener` method. 
App@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.webrtcwarning&modulesOnly=false&runModule=true:108258:76
RCTView
View
RCTView
View
AppContainer@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.webrtcwarning&modulesOnly=false&runModule=true:75604:36
WebRTCWarning(RootComponent)@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.webrtcwarning&modulesOnly=false&runModule=true:82460:28

`new NativeEventEmitter()` was called with a non-null argument without the required `removeListeners` method. 
App@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.webrtcwarning&modulesOnly=false&runModule=true:108258:76
RCTView
View
RCTView
View
AppContainer@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.webrtcwarning&modulesOnly=false&runModule=true:75604:36
WebRTCWarning(RootComponent)@http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.webrtcwarning&modulesOnly=false&runModule=true:82460:28

#### Steps to reproduce the problem
Create an empty project which creates the following object:

const testPC = new RTCPeerConnection({iceServers: []});

#### Platform information

* **React Native version**: 0.65.1
* **Plugin version**: 1.92.1
* **OS**: Android
* **OS version**: 11 RP1A.200720.011

PR #1072, opened 2021-09-15
Add required EventEmitter calls.

8BallBomBom:
Reference for the fix over [here](https://github.com/react-native-netinfo/react-native-netinfo/issues/486).
This should fix the recent issue #1070.


saghul:
This looks very odd since those methods are empty. Is there any documentation resource abut this?

8BallBomBom:
Bit of a weird one but it is related to [this](https://github.com/facebook/react-native/commit/114be1d2170bae2d29da749c07b45acf931e51e2).
Something that was changed in RN 0.65.0

As far as i'm aware react-native-reanimated and react-native-firebase have implemented the same fix without issues.
Can't say i'm sure why the change was made though.
Possibly to keep track of events in each module or allow intercepting event listeners.

saghul:
Thanks for digging, I'll take a look.

PR #1076, opened 2021-09-23
ios: handle av system pressure level updates

alexbumbu:
Handle AV system pressure level updates received when using the app in multitasking on iPad devices

### DIFF
--- a/ios/RCTWebRTC/VideoCaptureController.m
+++ b/ios/RCTWebRTC/VideoCaptureController.m
@@ -4,32 +4,36 @@
 #import <React/RCTLog.h>
 
 
-@implementation VideoCaptureController {
-    RTCCameraVideoCapturer *_capturer;
-    NSString *_deviceId;
-    BOOL _running;
-    BOOL _usingFrontCamera;
-    int _width;
-    int _height;
-    int _fps;
-}
+@interface VideoCaptureController ()
 
-@dynamic frameRate;
+@property (nonatomic, strong) RTCCameraVideoCapturer *capturer;
+@property (nonatomic, strong) AVCaptureDeviceFormat *selectedFormat;
+@property (nonatomic, strong) AVCaptureDevice *device;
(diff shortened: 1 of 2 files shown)


ios/RCTWebRTC/VideoCaptureController.m
```diff
@@ -45,88 +48,123 @@ -(instancetype)initWithCapturer:(RTCCameraVideoCapturer *)capturer
...
+        = [self selectFormatForDevice:self.device
+                      withTargetWidth:self.width
+                     withTargetHeight:self.height];
+    if (!format) {
+        RCTLogWarn(@"[VideoCaptureController] No valid formats for device %@", self.device);
 
         return;
     }
 
-    _selectedFormat = format;
```
alexbumbu, line 84: will fix this

ios/RCTWebRTC/VideoCaptureController.m
```diff
@@ -165,4 +203,19 @@ - (AVCaptureDeviceFormat *)selectFormatForDevice:(AVCaptureDevice *)device
...
+- (void)throttleFrameRateForDevice:(AVCaptureDevice *)device {
+    NSError *error = nil;
+    
+    [device lockForConfiguration:&error];
+    if (error) {
+        RCTLog(@"[VideoCaptureController] Could not lock device for configuration: %@", error);
+        return;
+    }
+    
+    device.activeVideoMinFrameDuration = CMTimeMake(1, 20);
```
saghul, line 215: Are these in the right order? We are setting this to 15fps min and 20 fps max, right?

saghul: Never mind: "A device’s minimum frame duration is the reciprocal of its maximum frame rate. "

alexbumbu: yeah, found this a bit confusing too

saghul (approved):
LGTM excellent work!

PR #1082, opened 2021-10-01
api: add RTCDataChannel.bufferedAmount

saghul:
Also implement the bufferedamountlow event.


saghul:
Not sure this is useful at all... @fippo maybe you know?

8BallBomBom:
Might be worth cleaning up to round off some of the Data Channel functionality.
Would be useful for sharing files via channels but most probably a very rare use case.

saghul:
Superseded by https://github.com/react-native-webrtc/react-native-webrtc/pull/1247

PR #1091, opened 2021-10-28
M94

saghul:
This update brings in WebRTC M94 plus some important changes to this plugin in itself:

- Support for iOS arm64 simulator (aka Apple M1 support)
- Binaries are now completely out of this repo, they'll be hosted here going forward: https://github.com/jitsi/webrtc/releases -- as you can see the package.json file contains links to them, this makes it easier to keep the WebRTC version alone while making plugin improvements, while keeping the repo size in check.
-Supported iOS versions are now >= 12 to match upstream WebRTC

All feedback is welcome!


nitrique:
Hi,

Works like a charm on my side :) (both debug an release configuration)
Tested on Macbook pro m1 13 inch + iPhone 12 Pro
 
Also fix a bug on my side : when a call is done, the next call crash the app (I think related to a camera/mic release mistake)

Regards

saghul:
Thanks! Let's continue in master then :-)

PR #1099, opened 2021-12-07
Bugfix/fix stats parsing

tmoldovan8x8:


android/src/main/java/com/oney/WebRTCModule/StringUtils.java
```diff
@@ -74,6 +74,26 @@ private static void appendValue(StringBuilder builder, Object value) {
...
+                    if (firstKey) {
+                        firstKey = false;
+                    } else {
+                        builder.append(",");
+                    }
+                    builder.append("\"").append(entry.getKey()).append("\":");
+                    appendValue(builder, entry.getValue());
+                }
+                builder.append("}");
+            } catch (ClassCastException e) {
```
saghul, line 94: What line can cause this? Looks like just `Map<String, Object> mapValue = (Map) value;`  Then let's keep it confined to that only line, and not append the value in that case, it'll likely be broken.

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -402,33 +402,56 @@ - (NSString *)statsToJSON:(RTCStatisticsReport *)report
...
+    [s appendString:@"}]"];
+  } 
+
+  [s appendString:@"]"];
+
+  return s;
+}
+
+- (NSString *)appendValue:(NSObject*) statisticsValue {
+    NSMutableString *s = [NSMutableString stringWithCapacity:16 * 1024];
```
saghul, line 417: Why create a new one? Let's pass around the original one when calling this function instead.

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -402,33 +402,56 @@ - (NSString *)statsToJSON:(RTCStatisticsReport *)report
...
+      }
+    
+      [s appendString:@"]"];
+    } else if ([statisticsValue isKindOfClass:[NSDictionary class]]) {
+        NSError *error;
+        NSData *jsonData = [NSJSONSerialization dataWithJSONObject:statisticsValue
+                                                           options:NSJSONWritingPrettyPrinted
+                                                           error:&error];
+
+        if (! jsonData) {
```
saghul, line 441: style: extraneous space after `!`

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -402,33 +402,56 @@ - (NSString *)statsToJSON:(RTCStatisticsReport *)report
...
+            [s appendString:@"\""];
+            [s appendString:[NSString stringWithFormat:@"%@", element]];
+            [s appendString:@"\""];
+      }
+    
+      [s appendString:@"]"];
+    } else if ([statisticsValue isKindOfClass:[NSDictionary class]]) {
+        NSError *error;
+        NSData *jsonData = [NSJSONSerialization dataWithJSONObject:statisticsValue
+                                                           options:NSJSONWritingPrettyPrinted
```
saghul, line 438: Don't pretty print it please, it wastes space.

saghul (changes requested):
Left some comments, PTAL!

android/src/main/java/com/oney/WebRTCModule/StringUtils.java
tmoldovan8x8: Indeed that line might cause it. I have removed any handling of the value in case an exception is thrown, but confining the try catch to just the declaration of the variable would mean adding some extra logic that I think would look weird.

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -413,9 +413,7 @@ - (NSString *)statsToJSON:(RTCStatisticsReport *)report
   return s;
 }
 
-- (NSString *)appendValue:(NSObject*) statisticsValue {
-    NSMutableString *s = [NSMutableString stringWithCapacity:16 * 1024];
-
+- (void)appendValue:(NSObject*) statisticsValue :(NSMutableString*) s {
```
saghul, line 416: give it a label, like `toString` otherwise it reads weird :-)

saghul (approved):

PR #1104, opened 2021-12-16
fix "Failed to create EGL context" exception

poorvasingh:
Problem: react-native-webrtc library crashes with "Failed to create EGL context" exception even with single context because EGL context is not released properly on dispose. 

Investigation: I explored chromium library and found this in there documented code at VideoCapturer.java
/*The caller still has ownership of {@code
* surfaceTextureHelper} and is responsible for making sure surfaceTextureHelper.dispose() is
* called. This also means that the caller can reuse the SurfaceTextureHelper to initialize a new
* VideoCapturer once the previous VideoCapturer has been disposed. */

It is further documented in SurfaceTextureHelper.java:
/**
* Helper class for using a SurfaceTexture to create WebRTC VideoFrames. In order to create WebRTC
* VideoFrames, render onto the SurfaceTexture. The frames will be delivered to the listener. Only
* one texture frame can be in flight at once, so the frame must be released in order to receive a
* new frame. Call stopListening() to stop receiveing new frames. Call dispose to release all
* resources once the texture frame is released.
*/

So, this Pull request aims to dispose surfaceTextureHelper properly when track are disposed.
Another reference to the problem can be found here: https://stackoverflow.com/questions/45392309/failed-to-create-egl-context

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
+++ b/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
@@ -107,9 +107,11 @@ class GetUserMediaImpl {
 
         AudioSource audioSource = pcFactory.createAudioSource(peerConstraints);
         AudioTrack track = pcFactory.createAudioTrack(id, audioSource);
+        
+        // surfaceTextureHelper is initialized for videoTrack only, so its null here.
         tracks.put(
             id,
-            new TrackPrivate(track, audioSource, /* videoCapturer */ null, null));
+            new TrackPrivate(track, audioSource, /* videoCapturer */ null, /* surfaceTextureHelper */ null));
 
         return track;
     }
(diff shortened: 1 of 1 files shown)


saghul (commented):
This looks very promising, thank you! @tmoldovan8x8 can you PTAL?

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -109,7 +109,7 @@ private AudioTrack createAudioTrack(ReadableMap constraints) {
         AudioTrack track = pcFactory.createAudioTrack(id, audioSource);
         tracks.put(
             id,
-            new TrackPrivate(track, audioSource, /* videoCapturer */ null));
+            new TrackPrivate(track, audioSource, /* videoCapturer */ null, null));
```
tmoldovan8x8, line 112: Could you please add a comment here to explain why it's fine to add null as surfaceTextureHelper

tmoldovan8x8:
Looks good to me I've also tested it and it seems to be working fine. Just a small comment.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
poorvasingh: done

tmoldovan8x8 (approved):

PR #1110, opened 2022-01-04
android: update uuid to get rid of the warning

8BallBomBom:
Started getting this warning recently.

warning react-native-webrtc > uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.

Should be a simple non-disruptive upgrade.

### DIFF
--- a/src/RTCUtil.js
+++ b/src/RTCUtil.js
@@ -112,6 +112,14 @@ function normalizeMediaConstraints(constraints, mediaType) {
     }
 }
 
+function chr4() {
+    return Math.random().toString(16).slice(-4);
+}
+
+export function uniqueID() {
+    return `${chr4()}${chr4()}-${chr4()}-${chr4()}-${chr4()}-${chr4()}${chr4()}${chr4()}`;
+}
+
 /**
  * Utility for deep cloning an object. Object.assign() only does a shallow copy.
  *
--- a/src/RTCUtil.js
+++ b/src/RTCUtil.js
@@ -113,8 +113,8 @@ function normalizeMediaConstraints(constraints, mediaType) {
 }
 
 /**
- * Utility for creating random float point values.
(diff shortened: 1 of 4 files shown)


saghul:
> Should be a simple non-disruptive upgrade.

On the contrary. This requires `crypto.getRandomValues` to exist, which requires the app to use a polyfill: https://github.com/uuidjs/uuid#getrandomvalues-not-supported

Since we don't use this for cryptographic purposes, but just as a simple ID to map objects I think a better solution would be to inline a function that generated a "unsafe" UUID v4 with Math.random.

8BallBomBom:
Ahhhhh absolutely didn't notice that, hmmmmm...

8BallBomBom:
@saghul Alrighty, nothing fancy but replicates the UUID v4 format 😃

src/RTCUtil.js
```diff
@@ -112,6 +112,14 @@ function normalizeMediaConstraints(constraints, mediaType) {
     }
 }
 
+function chr4() {
+    return Math.random().toString(16).slice(-4);
+}
+
+export function uniqueID() {
```
saghul, line 119: Can you please add some docs here explaining why we are doing it this way?

saghul (commented):
LGTM, please add some docs and we're good!

8BallBomBom:
I'll get to it, would you also prefer that i add the string together in a different way?

src/RTCUtil.js
8BallBomBom: Hopefully good enough 👍🏻

saghul (approved):

saghul:
Cheers!

PR #1015, opened 2021-05-27
[Draft] Openland PR rebased on Master with Updates

jamesdo:
I currently need the functionality from this [pull request](https://github.com/react-native-webrtc/react-native-webrtc/pull/773). The original PR didn't seem to work for me as it was. I (unnecessarily) squashed all changes from `openland`, and rebased on master and reconciled all changes.

This currently compiles for me, this is a very early version of this PR and is generally untested.

Disclaimer: This currently isn't working for me via `react-native run-ios --device`. I need to run `react-native run-ios`, and then once Metro is running, I can start the app via the XCode workspace.

PR #1014, opened 2021-05-27
Transceiver API Rebase on Master

jamesdo:
Rebasing https://github.com/openland/react-native-webrtc on master to bring it up to date.

This rebase is not currently working. I am working on a few strategies right now to get this branch working.

PR #1123, opened 2022-02-14
WIP for using transceivers

danjenkins:
android not compiling yet... something messed up in the apply of the patch.

The reason we're not discussing things in other Pull requests is a) they're closed and b) the structure of the project changed in v94 and it took quite a bit fo work to match those changes up to the pull requests. The resulting Merge will include the git usernames of the other parties who made it happen so that the right attributions are in place.

thanks to https://github.com/react-native-webrtc/react-native-webrtc/pull/1015 and https://github.com/react-native-webrtc/react-native-webrtc/pull/1014 

Some issues that need looking at include:

* addTransceiver is async but thats against spec
* android doesnt compile yet
* random uuids given to video tracks on ios and so trying to access them can be difficult

This is not a definitive list though

This is meant to be more for discussion around issues; I expect we'll end up with a branch elsewhere

 **important - Branch says 94.1 but it's now based off 98**

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -1,5 +1,6 @@
 package com.oney.WebRTCModule;
 
+import java.lang.ref.SoftReference;
 import android.util.Base64;
 import android.util.Log;
 import android.util.SparseArray;
@@ -20,6 +21,9 @@ import org.webrtc.MediaStreamTrack;
 import org.webrtc.PeerConnection;
 import org.webrtc.RtpReceiver;
 import org.webrtc.SessionDescription;
+import org.webrtc.RtpTransceiver;
+import org.webrtc.StatsObserver;
+import org.webrtc.StatsReport;
 import org.webrtc.VideoTrack;
 
 import java.io.UnsupportedEncodingException;
@@ -41,16 +45,27 @@ class PeerConnectionObserver implements PeerConnection.Observer {
     final List<MediaStream> localStreams;
     final Map<String, MediaStream> remoteStreams;
     final Map<String, MediaStreamTrack> remoteTracks;
-    private final VideoTrackAdapter videoTrackAdapters;
+    final boolean isUnifiedPlan;
+    final VideoTrackAdapter videoTrackAdapters;
     private final WebRTCModule webRTCModule;
 
-    PeerConnectionObserver(WebRTCModule webRTCModule, int id) {
+    /**
+     * The <tt>StringBuilder</tt> cache utilized by {@link #statsToJSON} in
+     * order to minimize the number of allocations of <tt>StringBuilder</tt>
+     * instances and, more importantly, the allocations of its <tt>char</tt>
+     * buffer in an attempt to improve performance.
+     */
+    private SoftReference<StringBuilder> statsToJSONStringBuilder
+        = new SoftReference<>(null);
+
+    PeerConnectionObserver(WebRTCModule webRTCModule, int id, boolean isUnifiedPlan) {
         this.webRTCModule = webRTCModule;
         this.id = id;
         this.dataChannels = new HashMap<>();
-        this.localStreams = new ArrayList<>();
-        this.remoteStreams = new HashMap<>();
-        this.remoteTracks = new HashMap<>();
+        this.isUnifiedPlan = isUnifiedPlan;
+        this.localStreams = new ArrayList<MediaStream>();
+        this.remoteStreams = new HashMap<String, MediaStream>();
+        this.remoteTracks = new HashMap<String, MediaStreamTrack>();
         this.videoTrackAdapters = new VideoTrackAdapter(webRTCModule, id);
     }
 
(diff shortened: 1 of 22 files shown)


saghul:
Hoping to get to this one very soon. Some extra context: in our testing anything < 96 is very slow to negotiate with transceivers, so I updated WebRTC to M98 (https://github.com/react-native-webrtc/react-native-webrtc/pull/1126) so we can eliminate that from the equation.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -82,6 +86,10 @@ public WebRTCModule(ReactApplicationContext reactContext, Options options) {
         mPeerConnectionObservers = new SparseArray<>();
         localStreams = new HashMap<>();
 
+        PeerConnectionFactory.initialize(
+                PeerConnectionFactory.InitializationOptions.builder(reactContext)
+                        .createInitializationOptions());
```
davidliu, line 91: PeerConnectionFactory is initialized below, so this seems unneeded.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -400,6 +424,8 @@ public void peerConnectionInit(ReadableMap configuration, int id) {
         PeerConnection.RTCConfiguration rtcConfiguration = parseRTCConfiguration(configuration);
 
         try {
+            ThreadUtils.runOnExecutor(() -> peerConnectionInitAsync(rtcConfiguration, id));
```
davidliu, line 427: This seems to duplicate the same code as below, but isn't run synchronously. Seems unneeded.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -410,6 +436,17 @@ public void peerConnectionInit(ReadableMap configuration, int id) {
             e.printStackTrace();
             throw new RuntimeException(e);
         }
+
```
davidliu, line 439: Missing closing brace for `peerConnectionInit()`

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -777,6 +839,7 @@ public void peerConnectionSetRemoteDescription(ReadableMap sdpMap,
                                                    int id,
                                                    Callback callback) {
         ThreadUtils.runOnExecutor(() -> {
+            peerConnectionAddICECandidateAsync(candidateMap, id, callback);
```
davidliu, line 842: Unrelated code? Also causes a compile error here, should be removed.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -746,6 +807,7 @@ public void onSetSuccess() {
                     WritableMap newSdpMap = Arguments.createMap();
                     newSdpMap.putString("type", newSdp.type.canonicalForm());
                     newSdpMap.putString("sdp", newSdp.description);
+                    newSdpMap.putMap("state", serializeState(id));
```
davidliu, line 810:
```suggestion
                    newSdpMap.putMap("state", serializeState(pcId));
```
Compile error

src/RTCPeerConnection.js
```diff
@@ -94,13 +97,38 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+
     createOffer(options) {
         return new Promise((resolve, reject) => {
             WebRTCModule.peerConnectionCreateOffer(
                 this._peerConnectionId,
                 RTCUtil.normalizeOfferAnswerOptions(options),
                 (successful, data) => {
                     if (successful) {
+                        this._mergeState(data.state);
                         resolve(data);
```
davidliu, line 132:
```suggestion
                        resolve(data.session);
```

Native is now returning a dictionary containing `state` and `session` (the actual sdp). Need to return the `session` value directly to conform to web api spec which returns the `RTCSessionDescription` directly.

src/RTCPeerConnection.js
```diff
@@ -117,6 +145,7 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
```
davidliu, line 149:
```suggestion
                        resolve(data.session);
```

davidliu:
With the above fixes, I can confirm both iOS/Android working with unified plan and `addTransceivers`.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -455,6 +500,15 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
         Log.d(TAG, "onAddTrack");
+        if(isUnifiedPlan){
+            MediaStreamTrack track = receiver.track();
+            if(track != null){
+                if(track.kind().equals(MediaStreamTrack.VIDEO_TRACK_KIND)){
+                    videoTrackAdapters.addAdapter(UUID.randomUUID().toString(), (VideoTrack) track);
+                }
+                remoteTracks.put(track.id(), track);
+            }
+        }
```
murillo128, line 511: Shouldn't we fire an event here?

saghul:
After spending a couple of days tinkering with this PR I am going to abandon ship. The RtpSender / receiver implementation makes it alas unusable. There are also a bunch of unnecessary and / or incorrect changes or weird merges.

Here is what I'm currently working on, PR coming soon:

- Drop plan B altogether
- Add senders / receivers / addTrack / ontrack
- Add transceivers

I think the only way forward it to pull a Safari and switch plans abruptly.

With this plan (pun intended) I would be able to test the core with Jitsi Meet at the middle point, and add transceivers just after.

murillo128:
> I'll open a new one shortly and we can see how to join forces there, WDYT?

Perfect!

PR #1140, opened 2022-03-31
M100

saghul:
I'm going to try and see if Apple approves this one...


gerhardberger:
hey!

I'm wondering, what was the reasoning behind the [`force plan B`](https://github.com/react-native-webrtc/react-native-webrtc/pull/1140/commits/188f0934871aef8f478c7d3fa403deb6ed5648bf) commit for the M100 upgrade?

8BallBomBom:
Probably due to Unified Plan support not being complete.

gerhardberger:
yea, I saw that but Unified Plan support haven't been complete with previous MXX versions as well and we didn't have this forcing in place, so I'm wondering why was it needed to force it with this release.

8BallBomBom:
Good point, most likely done to force the use of what is supported to prevent issues.
But that is entirely assumption.

@saghul Can you elaborate please?

saghul:
In M100 the default changed to Unified, so if you want Plan B you have to explicitly select it.

PR #1151, opened 2022-05-13
ios: cleanup existing connection on screen capture start

alexbumbu:


ios/RCTWebRTC/ScreenCapturer.m
```diff
@@ -141,6 +141,13 @@ - (instancetype)initWithDelegate:(__weak id<RTCVideoCapturerDelegate>)delegate {
     return self;
 }
 
+- (void)setConnection:(SocketConnection *)connection {
```
saghul, line 144: Would it make sense to simplify `stopCapture` to just set connection to `nil` now?

saghul (approved):
LGTM with a question.

ios/RCTWebRTC/ScreenCapturer.m
alexbumbu: yeah, makes sense. Will do another commit with the fix.

PR #1155, opened 2022-05-23
Cleanup Type Issues

SE2Dev:
The current version of the repo has numerous type issues that make `react-native-webrtc` rather difficult to use in a TypeScript project. I've added explicit return types to numerous functions (as they were resolving to `unknown` or `any` on any external project that used `react-native-webrtc`), as well cleaned up several other minor aspects of the code base.

There are other changes that need to be made, but this pull request primarily focuses on the return types and a few other miscellaneous changes.

### DIFF
--- a/src/RTCDataChannel.ts
+++ b/src/RTCDataChannel.ts
@@ -79,7 +79,7 @@ export default class RTCDataChannel extends defineCustomEventTarget(...DATA_CHAN
         return this._readyState;
     }
 
-    send(data: string | ArrayBuffer | ArrayBufferView) {
+    send(data: string | ArrayBuffer | ArrayBufferView): void {
         if (typeof data === 'string') {
             WebRTCModule.dataChannelSend(this._peerConnectionId, this._reactTag, data, 'text');
             return;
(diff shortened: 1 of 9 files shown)


src/RTCDataChannel.ts
```diff
@@ -79,7 +79,10 @@ export default class RTCDataChannel extends defineCustomEventTarget(...DATA_CHAN
         return this._readyState;
     }
 
-    send(data: string | ArrayBuffer | ArrayBufferView) {
+    send(data: string): void;
```
saghul, line 82: Why the 3 lines?

saghul (commented):
This is great, thanks! I left a question, LGTM otherwise!

src/RTCDataChannel.ts
SE2Dev: They're [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads). It won't affect the transpiled JavaScript at all - I mainly did it so that it'd match the overloads found in the actual DOM types seen [here](https://github.com/microsoft/TypeScript/blob/7c6521e8fb8f24a31f28e00e92838ad1b975fbb3/lib/lib.dom.d.ts#L11605).

PR #1157, opened 2022-05-31
examples-etc, random improvements, new example apps

8BallBomBom:
Will be starting on this soon.
As always i'm open to suggestions 😃 

- [ ] Provide a reasonably extensive example app - **Android, iOS, macOS, Web**
- [ ] Provide a simple but extensive example signalling server - **NodeJS**
- [ ] Include documentation for all new example apps
- [ ] Include organisation modules and use cases in documentation
- [ ] Include notification handling via Firebase
- [ ] Include any changes across the board due to module updates
- [ ] Go over all documentation for any final touches
- [ ] Provide a MacOS install guide? **(Since i don't have one this will be interesting...)**


8BallBomBom:
Doesn't sound like a bad idea, could then just link them from here and do a half n half pr, update some of the existing docs here for the other things. 😄

8BallBomBom:
Sounds good, i'll change this pr to just be doc improvements and to cover some of the other things, use case of react-native-incall-manager, other install guides?
Then eventually the call guide could be updated and actually reference/include code samples on how to use the new signalling example server.

8BallBomBom:
Didn't know it was going to close if i renamed the branch haha, ahh well 🤣

PR #1160, opened 2022-06-09
Unified plan support

Yousif-CS:
This PR adds support for the unified plan transceivers API. This mainly adds the following APIs to RTCPeerConnection: 
```
addTransceiver(...)
getTransceivers()
addTrack(...)
removeTrack(...)
``` 
Along with the required classes like `RTCRtpTransceiver`, `RTCRtpSender`, etc...

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -9,6 +9,7 @@ import androidx.annotation.Nullable;
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.WritableArray;
 import com.facebook.react.bridge.WritableMap;
 
@@ -19,6 +20,8 @@ import org.webrtc.MediaStream;
 import org.webrtc.MediaStreamTrack;
 import org.webrtc.PeerConnection;
 import org.webrtc.RtpReceiver;
+import org.webrtc.RtpSender;
+import org.webrtc.RtpTransceiver;
 import org.webrtc.SessionDescription;
 import org.webrtc.VideoTrack;
 
@@ -30,6 +33,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 
 class PeerConnectionObserver implements PeerConnection.Observer {
@@ -37,8 +41,9 @@ class PeerConnectionObserver implements PeerConnection.Observer {
 
     private final Map<String, DataChannelWrapper> dataChannels;
     private final int id;
+    private int transceiverNextId = 0;
+
     private PeerConnection peerConnection;
-    final List<MediaStream> localStreams;
     final Map<String, MediaStream> remoteStreams;
     final Map<String, MediaStreamTrack> remoteTracks;
     private final VideoTrackAdapter videoTrackAdapters;
@@ -48,47 +53,9 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         this.webRTCModule = webRTCModule;
         this.id = id;
         this.dataChannels = new HashMap<>();
-        this.localStreams = new ArrayList<>();
         this.remoteStreams = new HashMap<>();
         this.remoteTracks = new HashMap<>();
         this.videoTrackAdapters = new VideoTrackAdapter(webRTCModule, id);
-    }
-
-    /**
-     * Adds a specific local <tt>MediaStream</tt> to the associated
-     * <tt>PeerConnection</tt>.
-     *
-     * @param localStream the local <tt>MediaStream</tt> to add to the
-     *                    associated <tt>PeerConnection</tt>
-     * @return <tt>true</tt> if the specified <tt>localStream</tt> was added to
-     * the associated <tt>PeerConnection</tt>; otherwise, <tt>false</tt>
-     */
-    boolean addStream(MediaStream localStream) {
-        if (peerConnection != null && peerConnection.addStream(localStream)) {
-            localStreams.add(localStream);
-
-            return true;
-        }
-
-        return false;
-    }
-
-    /**
-     * Removes a specific local <tt>MediaStream</tt> from the associated
-     * <tt>PeerConnection</tt>.
-     *
-     * @param localStream the local <tt>MediaStream</tt> from the associated
-     *                    <tt>PeerConnection</tt>
-     * @return <tt>true</tt> if removing the specified <tt>mediaStream</tt> from
-     * this instance resulted in a modification of its internal list of local
-     * <tt>MediaStream</tt>s; otherwise, <tt>false</tt>
-     */
-    boolean removeStream(MediaStream localStream) {
-        if (peerConnection != null) {
-            peerConnection.removeStream(localStream);
-        }
-
-        return localStreams.remove(localStream);
     }
 
     PeerConnection getPeerConnection() {
@@ -105,19 +72,10 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         // Close the PeerConnection first to stop any events.
         peerConnection.close();
 
-        // PeerConnection.dispose() calls MediaStream.dispose() on all local
-        // MediaStreams added to it and the app may crash if a local MediaStream
-        // is added to multiple PeerConnections. In order to reduce the risks of
-        // an app crash, remove all local MediaStreams from the associated
-        // PeerConnection so that it doesn't attempt to dispose of them.
-        for (MediaStream localStream : new ArrayList<>(localStreams)) {
-            removeStream(localStream);
-        }
-
         // Remove video track adapters
-        for (MediaStream stream : remoteStreams.values()) {
-            for (VideoTrack videoTrack : stream.videoTracks) {
-                videoTrackAdapters.removeAdapter(videoTrack);
+        for (MediaStreamTrack track : this.remoteTracks.values()) {
+            if (track instanceof VideoTrack) {
+                videoTrackAdapters.removeAdapter((VideoTrack) track);
             }
         }
 
@@ -136,6 +94,54 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         remoteStreams.clear();
         remoteTracks.clear();
         dataChannels.clear();
+    }
+
+
+    public synchronized int getNextTransceiverId() {
+        return transceiverNextId++;
+    }
+
+    RtpTransceiver addTransceiver(MediaStreamTrack.MediaType mediaType, RtpTransceiver.RtpTransceiverInit init) {
+        if (peerConnection == null) {
+            return null;
+        }
+
+        return peerConnection.addTransceiver(mediaType, init);
+    }
+
+    RtpTransceiver addTransceiver(MediaStreamTrack track, RtpTransceiver.RtpTransceiverInit init) {
+        if (peerConnection == null) {
+            return null;
+        }
+
+        return peerConnection.addTransceiver(track, init);
+    }
+
+    RtpSender getSender(String id) {
+        if (this.peerConnection == null) {
+            return null;
+        }
+
+        for (RtpSender sender: this.peerConnection.getSenders()) {
+            if (sender.id().equals(id)) {
+                return sender;
+            }
+        }
+
+        return null;
+    }
+
+    RtpTransceiver getTransceiver(String id) {
+        if (this.peerConnection == null) {
+            return null;
+        }
+
+        for (RtpTransceiver transceiver: this.peerConnection.getTransceivers()) {
+            if (transceiver.getSender().id().equals(id)) {
+                return transceiver;
+            }
+        }
+        return null;
     }
 
     WritableMap createDataChannel(String label, ReadableMap config) {
@@ -294,120 +300,6 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         webRTCModule.sendEvent("peerConnectionIceGatheringChanged", params);
     }
 
-    private String getReactTagForStream(MediaStream mediaStream) {
-        for (Iterator<Map.Entry<String, MediaStream>> i
-             = remoteStreams.entrySet().iterator();
-             i.hasNext(); ) {
-            Map.Entry<String, MediaStream> e = i.next();
-            if (e.getValue().equals(mediaStream)) {
-                return e.getKey();
-            }
-        }
-        return null;
-    }
-
-    @Override
-    public void onAddStream(MediaStream mediaStream) {
-        String streamReactTag = null;
-        String streamId = mediaStream.getId();
-        // The native WebRTC implementation has a special concept of a default
-        // MediaStream instance with the label default that the implementation
-        // reuses.
-        if ("default".equals(streamId)) {
-            for (Map.Entry<String, MediaStream> e : remoteStreams.entrySet()) {
-                if (e.getValue().equals(mediaStream)) {
-                    streamReactTag = e.getKey();
-                    break;
-                }
-            }
-        }
-
-        if (streamReactTag == null) {
-            streamReactTag = UUID.randomUUID().toString();
-            remoteStreams.put(streamReactTag, mediaStream);
-        }
-
-        WritableMap params = Arguments.createMap();
-        params.putInt("id", id);
-        params.putString("streamId", streamId);
-        params.putString("streamReactTag", streamReactTag);
-
-        WritableArray tracks = Arguments.createArray();
-
-        for (int i = 0; i < mediaStream.videoTracks.size(); i++) {
-            VideoTrack track = mediaStream.videoTracks.get(i);
-            String trackId = track.id();
-
-            remoteTracks.put(trackId, track);
-
-            WritableMap trackInfo = Arguments.createMap();
-            trackInfo.putString("id", trackId);
-            trackInfo.putString("label", "Video");
-            trackInfo.putString("kind", track.kind());
-            trackInfo.putBoolean("enabled", track.enabled());
-            trackInfo.putString("readyState", track.state().toString());
-            trackInfo.putBoolean("remote", true);
-            tracks.pushMap(trackInfo);
-
-            videoTrackAdapters.addAdapter(track);
-        }
-        for (int i = 0; i < mediaStream.audioTracks.size(); i++) {
-            AudioTrack track = mediaStream.audioTracks.get(i);
-            String trackId = track.id();
-
-            remoteTracks.put(trackId, track);
-
-            WritableMap trackInfo = Arguments.createMap();
-            trackInfo.putString("id", trackId);
-            trackInfo.putString("label", "Audio");
-            trackInfo.putString("kind", track.kind());
-            trackInfo.putBoolean("enabled", track.enabled());
-            trackInfo.putString("readyState", track.state().toString());
-            trackInfo.putBoolean("remote", true);
-            tracks.pushMap(trackInfo);
-        }
-        params.putArray("tracks", tracks);
-
-        SessionDescription newSdp = peerConnection.getRemoteDescription();
-        WritableMap newSdpMap = Arguments.createMap();
-        newSdpMap.putString("type", newSdp.type.canonicalForm());
-        newSdpMap.putString("sdp", newSdp.description);
-        params.putMap("sdp", newSdpMap);
-
-        webRTCModule.sendEvent("peerConnectionAddedStream", params);
-    }
-
-    @Override
-    public void onRemoveStream(MediaStream mediaStream) {
-        String streamReactTag = getReactTagForStream(mediaStream);
-        if (streamReactTag == null) {
-            Log.w(TAG, "onRemoveStream - no remote stream for id: " + mediaStream.getId());
-            return;
-        }
-
-        for (VideoTrack track : mediaStream.videoTracks) {
-            this.videoTrackAdapters.removeAdapter(track);
-            this.remoteTracks.remove(track.id());
-        }
-        for (AudioTrack track : mediaStream.audioTracks) {
-            this.remoteTracks.remove(track.id());
-        }
-
-        this.remoteStreams.remove(streamReactTag);
-
-        WritableMap params = Arguments.createMap();
-        params.putInt("id", id);
-        params.putString("streamId", streamReactTag);
-
-        SessionDescription newSdp = peerConnection.getRemoteDescription();
-        WritableMap newSdpMap = Arguments.createMap();
-        newSdpMap.putString("type", newSdp.type.canonicalForm());
-        newSdpMap.putString("sdp", newSdp.description);
-        params.putMap("sdp", newSdpMap);
-
-        webRTCModule.sendEvent("peerConnectionRemovedStream", params);
-    }
-
     @Override
     public void onDataChannel(DataChannel dataChannel) {
         final String reactTag  = UUID.randomUUID().toString();
@@ -452,9 +344,111 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         webRTCModule.sendEvent("peerConnectionSignalingStateChanged", params);
     }
 
+    /**
+     * Triggered when a new track is signaled by the remote peer, as a result of
+     * setRemoteDescription.
+     */
     @Override
     public void onAddTrack(final RtpReceiver receiver, final MediaStream[] mediaStreams) {
         Log.d(TAG, "onAddTrack");
+
+        ThreadUtils.runOnExecutor(() -> {
+            RtpTransceiver transceiver = null;
+            for(RtpTransceiver t: this.peerConnection.getTransceivers()) {
+                if (Objects.equals(t.getReceiver().id(), receiver.id())) {
+                    transceiver = t;
+                    break;
+                }
+            }
+
+            if (transceiver == null) {
+                return;
+            }
+
+            final MediaStreamTrack track = receiver.track();
+
+            if (remoteTracks.containsKey(track.id())) {
+                // Unlike in the WebRTC spec, the libwebrtc native implementation
+                // fires onTrack on every sRD which has an active receiving transceiver.
+                // So, if we are already keeping track of this transceiver, ignore the event.
+                return;
+            }
+
+            if (track.kind().equals(MediaStreamTrack.VIDEO_TRACK_KIND)){
+                videoTrackAdapters.addAdapter((VideoTrack) track);
+            }
+            remoteTracks.put(track.id(), track);
+
+            WritableMap params = Arguments.createMap();
+            WritableArray streams = Arguments.createArray();
+
+            for (MediaStream stream : mediaStreams) {
+                // Getting the streamReactTag
+                String streamReactTag = null;
+                for (Map.Entry<String, MediaStream> e : remoteStreams.entrySet()) {
+                    if (e.getValue().equals(stream)) {
+                        streamReactTag = e.getKey();
+                        break;
+                    }
+                }
+                if (streamReactTag == null) {
+                    streamReactTag = UUID.randomUUID().toString();
+                    remoteStreams.put(streamReactTag, stream);
+                }
+                streams.pushMap(SerializeUtils.serializeStream(id, streamReactTag, stream));
+            }
+
+            params.putArray("streams", streams);
+            params.putMap("receiver", SerializeUtils.serializeReceiver(id, receiver));
+            params.putInt("transceiverOrder", getNextTransceiverId());
+            params.putMap("transceiver", SerializeUtils.serializeTransceiver(id, transceiver));
+
+            params.putInt("id", this.id);
+
+            webRTCModule.sendEvent("peerConnectionOnTrack", params);
+        });
+    }
+
+    /**
+     * Triggered when the signaling from SetRemoteDescription indicates that a transceiver
+     * will be receiving media from a remote endpoint. This is only called if UNIFIED_PLAN
+     * semantics are specified. The transceiver will be disposed automatically.
+     */
+    @Override
+    public void onTrack(final RtpTransceiver transceiver) {
+    }
+
+    /*
+     * Triggered when a previously added remote track is removed by the remote
+     * peer, as a result of setRemoteDescription.
+     */
+    @Override
+    public void onRemoveTrack(RtpReceiver receiver){
+        // According to the W3C spec, we need to send out
+        // the track Id so that we can remove it from the MediaStream objects stored
+        // at the JS layer, which are the same stream objects passed down to
+        // the `track` event.
+        ThreadUtils.runOnExecutor(() -> {
+            MediaStreamTrack track = receiver.track();
+            WritableMap params = Arguments.createMap();
+            params.putInt("id", this.id);
+            params.putString("trackId", track.id());
+
+            webRTCModule.sendEvent("peerConnectionOnRemoveTrack", params);
+
+        });
+    };
+
+    // This is only added to compile. Plan B is not supported anymore.
+    @Override
+    public void onRemoveStream(MediaStream stream) {
+
+    }
+
+    // This is only added to compile. Plan B is not supported anymore.
+    @Override
+    public void onAddStream(MediaStream stream) {
+
     }
 
     @Nullable
@@ -528,4 +522,5 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         }
         return null;
     }
+
 }
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -17,6 +17,7 @@ import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.ReadableMapKeySetIterator;
 import com.facebook.react.bridge.ReadableType;
 import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.bridge.WritableArray;
 import com.facebook.react.module.annotations.ReactModule;
 import com.facebook.react.modules.core.DeviceEventManagerModule;
 
@@ -37,6 +38,10 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
     static final String TAG = WebRTCModule.class.getCanonicalName();
 
     PeerConnectionFactory mFactory;
+    VideoEncoderFactory mVideoEncoderFactory;
+    VideoDecoderFactory mVideoDecoderFactory;
+
+    // Need to expose the peer connection codec factories here to get capabilities
     private final SparseArray<PeerConnectionObserver> mPeerConnectionObservers;
     final Map<String, MediaStream> localStreams;
 
@@ -132,6 +137,10 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
                 .setVideoDecoderFactory(decoderFactory)
                 .createPeerConnectionFactory();
 
+        // Saving the encoder and decoder factories to get codec info later when needed
+        mVideoEncoderFactory = encoderFactory;
+        mVideoDecoderFactory = decoderFactory;
+
         getUserMediaImpl = new GetUserMediaImpl(this, reactContext);
     }
 
@@ -146,10 +155,20 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
         return (pco == null) ? null : pco.getPeerConnection();
     }
 
-    void sendEvent(String eventName, @Nullable WritableMap params) {
+    void sendEvent(String eventName, @Nullable ReadableMap params) {
         getReactApplicationContext()
             .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
             .emit(eventName, params);
+    }
+
+    void sendError(String eventName, String funcName, @Nullable String message, @Nullable ReadableMap info) {
+        WritableMap errorInfo = Arguments.createMap();
+        errorInfo.putString("func", funcName);
+        if (info != null)
+            errorInfo.putMap("info", info);
+        if (message != null)
+            errorInfo.putString("message", message);
+        sendEvent(eventName, errorInfo);
     }
 
     private PeerConnection.IceServer createIceServer(String url) {
@@ -206,8 +225,7 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
         // Required for perfect negotiation.
         conf.enableImplicitRollback = true;
 
-        // Plan B, just a little longer.
-        conf.sdpSemantics = PeerConnection.SdpSemantics.PLAN_B;
+        conf.sdpSemantics = PeerConnection.SdpSemantics.UNIFIED_PLAN;
 
         if (map == null) {
             return conf;
@@ -438,7 +456,7 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
         }
     }
 
-    private MediaStreamTrack getTrack(String trackId) {
+    public MediaStreamTrack getTrack(String trackId) {
         MediaStreamTrack track = getLocalTrack(trackId);
 
         if (track == null) {
@@ -479,6 +497,255 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
         }
 
         return mediaConstraints;
+    }
+
+
+
+    @ReactMethod(isBlockingSynchronousMethod = true)
+    public WritableMap peerConnectionAddTransceiver(int id, ReadableMap options) {
+        try {
+            return (WritableMap) ThreadUtils.submitToExecutor((Callable<Object>) () -> {
+                PeerConnectionObserver pco = mPeerConnectionObservers.get(id);
+                if (pco == null) {
+                    Log.d(TAG, "peerConnectionAddTransceiver() peerConnection is null");
+                    return null;
+                }
+
+                RtpTransceiver transceiver = null;
+                if (options.hasKey("type")) {
+                    String kind = options.getString("type");
+                    transceiver = pco.addTransceiver(SerializeUtils.parseMediaType(kind),
+                        SerializeUtils.parseTransceiverOptions(options.getMap("init")));
+                } else if (options.hasKey("trackId")) {
+                    String trackId = options.getString("trackId");
+                    MediaStreamTrack track = getTrack(trackId);
+                    transceiver = pco.addTransceiver(track,
+                        SerializeUtils.parseTransceiverOptions(options.getMap("init")));
+
+                } else {
+                    // This should technically never happen as the JS side checks for that.
+                    Log.d(TAG, "peerConnectionAddTransceiver() no type nor trackId provided in options");
+                    return null;
+                }
+
+                if (transceiver == null) {
+                    Log.d(TAG, "peerConnectionAddTransceiver() Error adding transceiver");
+                    return null;
+                }
+                WritableMap params = Arguments.createMap();
+                // We need to get a unique order at which the transceiver was created
+                // to reorder the cached array of transceivers on the JS layer.
+                params.putInt("transceiverOrder", pco.getNextTransceiverId());
+                params.putMap("transceiver", SerializeUtils.serializeTransceiver(id, transceiver));
+                return params;
+            }).get();
+        } catch (InterruptedException | ExecutionException e) {
+            Log.d(TAG, "peerConnectionAddTransceiver() " + e.getMessage());
+            return null;
+        }
+    }
+
+    @ReactMethod(isBlockingSynchronousMethod = true)
+    public WritableMap peerConnectionAddTrack(int id,
+                                              String trackId,
+                                              ReadableMap options) {
+        try {
+            return (WritableMap) ThreadUtils.submitToExecutor((Callable<Object>) () -> {
+                PeerConnectionObserver pco = mPeerConnectionObservers.get(id);
+                if (pco == null) {
+                    Log.d(TAG, "peerConnectionAddTrack() peerConnection is null");
+                    return null;
+                }
+
+                MediaStreamTrack track = getLocalTrack(trackId);
+                if (track == null) {
+                    Log.w(TAG, "peerConnectionAddTrack() couldn't find track " + trackId);
+                    return null;
+                }
+
+                List<String> streamIds = new ArrayList<>();
+                if (options.hasKey("streamIds")) {
+                    ReadableArray rawStreamIds = options.getArray("streamIds");
+                    if (rawStreamIds != null) {
+                        for (int i = 0; i < rawStreamIds.size(); i++) {
+                            streamIds.add(rawStreamIds.getString(i));
+                        }
+                    }
+                }
+                RtpSender sender = pco.getPeerConnection().addTrack(track, streamIds);
+
+                // Need to get the corresponding transceiver as well
+                RtpTransceiver transceiver = pco.getTransceiver(sender.id());
+
+                // We need the transceiver creation order to reorder the transceivers array
+                // in the JS layer.
+                WritableMap params = Arguments.createMap();
+                params.putInt("transceiverOrder", pco.getNextTransceiverId());
+                params.putMap("transceiver", SerializeUtils.serializeTransceiver(id, transceiver));
+                params.putMap("sender", SerializeUtils.serializeSender(id, sender));
+                return params;
+            }).get();
+        } catch (InterruptedException | ExecutionException e) {
+            Log.d(TAG, "peerConnectionAddTrack() " + e.getMessage());
+            return null;
+        }
+    }
+
+    @ReactMethod
+    public void senderSetParameters(int id, String senderId, ReadableMap options, Promise promise) {
+        ThreadUtils.runOnExecutor(() ->{
+            try {
+                PeerConnectionObserver pco = mPeerConnectionObservers.get(id);
+                if (pco == null) {
+                    Log.d(TAG, "senderSetParameters() peerConnectionObserver is null");
+                    promise.reject(new Exception("Peer Connection is not initialized"));
+                    return;
+                }
+
+                RtpSender sender = pco.getSender(senderId);
+                if (sender == null) {
+                    Log.w(TAG, "senderSetParameters() sender is null");
+                    promise.reject(new Exception("Could not get sender"));
+                    return;
+                }
+
+                RtpParameters params = sender.getParameters();
+                params = SerializeUtils.updateRtpParameters(options, params);
+                sender.setParameters(params);
+                promise.resolve(SerializeUtils.serializeRtpParameters(sender.getParameters()));
+            } catch (Exception e) {
+                Log.d(TAG, "senderSetParameters: " + e.getMessage());
+                promise.reject(e);
+            }
+        });
+    }
+
+    @ReactMethod
+    public void peerConnectionRemoveTrack(int id, String senderId) {
+            ThreadUtils.runOnExecutor(() -> {
+                WritableMap identifier = Arguments.createMap();
+                identifier.putInt("peerConnectionId", id);
+                identifier.putString("senderId", senderId);
+                try {
+                    PeerConnectionObserver pco = mPeerConnectionObservers.get(id);
+                    if (pco == null) {
+                        Log.d(TAG, "peerConnectionRemoveTrack() peerConnection is null");
+                        sendError("peerConnectionOnError", "removeTrack", "Peer Connection is not initialized", null);
+                        return;
+                    }
+                    RtpSender sender = pco.getSender(senderId);
+                    if (sender == null) {
+                        Log.w(TAG, "peerConnectionRemoveTrack() sender is null");
+                        sendError("peerConnectionOnError", "removeTrack", "Sender is null", null);
+                        return;
+                    }
+
+                    boolean successful = pco.getPeerConnection().removeTrack(sender);
+                    if (successful) {
+                        sendEvent("peerConnectionOnRemoveTrackSuccessful", identifier);
+                        return;
+                    }
+                    sendError("peerConnectionOnError", "removeTrack", "Internal Error", identifier);
+                } catch (Exception e) {
+                    Log.d(TAG, "peerConnectionRemoveTrack() " + e.getMessage());
+                    sendError("peerConnectionOnError", "removeTrack", e.getMessage(), identifier);
+                }
+            });
+    }
+
+    @ReactMethod
+    public void transceiverStop(int id,
+                                String senderId) {
+        ThreadUtils.runOnExecutor(() -> {
+            WritableMap identifier = Arguments.createMap();
+            identifier.putInt("peerConnectionId", id);
+            identifier.putString("transceiverId", senderId);
+            try {
+                PeerConnectionObserver pco = mPeerConnectionObservers.get(id);
+                if (pco == null) {
+                    Log.d(TAG, "transceiverStop() peerConnectionObserver is null");
+                    // Cannot really report an error specific to RTCRtpTransceiver as the peer connection
+                    // hasn't really been initialized, this call from the web API should technically
+                    // not happen unless both peer connection and transceivers are initialized properly
+                    // That is why constructors or create methods should always be synchronous.
+                    return;
+                }
+                RtpTransceiver transceiver = pco.getTransceiver(senderId);
+
+                transceiver.stopStandard();
+                sendEvent("transceiverStopSuccessful", identifier);
+            } catch (Exception e) {
+                Log.d(TAG, "transceiverStop(): " + e.getMessage());
+                sendError("transceiverOnError", "stopTransceiver", e.getMessage(), identifier);
+            }
+        });
+    }
+
+    @ReactMethod
+    public void senderReplaceTrack(int id,
+                                   String senderId,
+                                   String trackId,
+                                   Promise promise) {
+        ThreadUtils.runOnExecutor(() -> {
+            try {
+                PeerConnectionObserver pco = mPeerConnectionObservers.get(id);
+                if (pco == null) {
+                    Log.d(TAG, "senderReplaceTrack() peerConnectionObserver is null");
+                    promise.reject(new Exception("Peer Connection is not initialized"));
+                    return;
+                }
+
+                RtpSender sender = pco.getSender(senderId);
+                if (sender == null) {
+                    Log.w(TAG, "senderReplaceTrack() sender is null");
+                    promise.reject(new Exception("Could not get sender"));
+                    return;
+                }
+
+                MediaStreamTrack track = getLocalTrack(trackId);
+                sender.setTrack(track, false);
+                promise.resolve(true);
+            } catch (Exception e) {
+                Log.d(TAG, "senderReplaceTrack(): " + e.getMessage());
+                promise.reject(e);
+            }
+        });
+    }
+
+    @ReactMethod
+    public void transceiverSetDirection(int id,
+                                        String senderId,
+                                        String direction) {
+
+        ThreadUtils.runOnExecutor(() -> {
+            WritableMap identifier = Arguments.createMap();
+            WritableMap params = Arguments.createMap();
+            identifier.putInt("peerConnectionId", id);
+            identifier.putString("transceiverId", senderId);
+            try {
+                PeerConnectionObserver pco = mPeerConnectionObservers.get(id);
+                if (pco == null) {
+                    Log.d(TAG, "transceiverSetDirection() peerConnectionObserver is null");
+                    // Cannot really report an error specific to RTCRtpSender as the peer connection
+                    // hasn't really been initialized, this call from the web API should technically
+                    // not happen unless both peer connection and sender are initialized
+                    return;
+                }
+                RtpTransceiver transceiver = pco.getTransceiver(senderId);
+                if (transceiver == null){
+                    Log.d(TAG, "transceiverSetDirection() transceiver is null");
+                    return;
+                }
+
+                RtpTransceiver.RtpTransceiverDirection oldDirection = transceiver.getDirection();
+                params.putString("oldDirection", SerializeUtils.serializeDirection(oldDirection));
+                transceiver.setDirection(SerializeUtils.parseDirection(direction));
+            } catch (Exception e) {
+                Log.d(TAG, "transceiverSetDirection(): " + e.getMessage());
+                params.putMap("identifiers", identifier);
+                sendError("transceiverOnError", "setDirection", e.getMessage(), params);
+            }
+        });
     }
 
     @ReactMethod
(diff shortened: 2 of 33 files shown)


src/RTCRtpReceiver.ts
```diff
@@ -0,0 +1,36 @@
...
+    constructor(info: { peerConnectionId: number, id: string, track: MediaStreamTrack }) {
+        this._id = info.id;
+        this._peerConnectionId = info.peerConnectionId;
+        this._track = info.track;
+    }
+
+    static getCapabilities(kind: "audio" | "video"): RTCRtpCapabilities {
+
+        if (kind == "audio") {
+            throw new Error("Unimplemented capabilities for audio");
```
saghul, line 21: We could hardcode some, since we provide the matching WebRTC build...

src/RTCRtpSender.ts
```diff
@@ -0,0 +1,50 @@
...
+    constructor(info: { peerConnectionId: number, id: string, track?: MediaStreamTrack }) {
+        this._peerConnectionId = info.peerConnectionId;
+        this._id = info.id;
+        if (info.track)
+            this._track = info.track;
+    }
+
+    replaceTrack(track: MediaStreamTrack | null): Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            WebRTCModule.senderReplaceTrack(this._peerConnectionId, this._id, track ? track.id : null, (successful, data) => {
```
saghul, line 22: Make this return a native promise instead. Callbacks in this library are an ancient thing we try to get rid off.

src/RTCRtpTransceiver.ts
```diff
@@ -0,0 +1,127 @@
...
+        receiver: RTCRtpReceiver,
+    }) {
+        super();
+        this._peerConnectionId = args.peerConnectionId;
+        this._id = args.id;
+        this._mid = args.mid ? args.mid : null;
+        this._direction = args.direction;
+        this._currentDirection = args.currentDirection;
+        this._stopped = args.isStopped;
+        this._sender = new RTCRtpSender({ peerConnectionId: args.peerConnectionId, id: args.sender.id, track: args.sender.track? args.sender.track : undefined });
```
saghul, line 42: Isn't this done outside and we pass the sender / receiver object in the constructor? If not, why not?

saghul (commented):
Great progress, left a bunch of comments, PTAL!

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -455,6 +369,125 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
     public void onAddTrack(final RtpReceiver receiver, final MediaStream[] mediaStreams) {
         Log.d(TAG, "onAddTrack");
+
+        // Get The timestamp so that we can reorder the transceivers on the JS layer
+        Long tsLong = System.currentTimeMillis()/1000;
+        String ts = tsLong.toString();
+
+        MediaStreamTrack track = receiver.track();
+
+        if (track == null || sdpSemantics != PeerConnection.SdpSemantics.UNIFIED_PLAN) {
```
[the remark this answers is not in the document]
Yousif-CS:
```  
@Nullable
public MediaStreamTrack track() {
  return cachedTrack;
}
```
Indicates that it could be null, that is why there is a check

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -455,6 +369,125 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+        for( RtpTransceiver transceiver : transceivers ) {
+            if(transceiver.getReceiver() != null && receiver.id().equals(transceiver.getReceiver().id())) {
+                eventInfo.putMap("transceiver", serializeTransceiver(transceiver));
+                break;
+            }
+        }
+        params.putInt("id", this.id);
+        params.putMap("info", eventInfo);
+
+        webRTCModule.sendEvent("peerConnectionOnTrack", params);
```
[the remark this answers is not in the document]
Yousif-CS: The way I understood it is that `onAddTrack` is called when there also MediaStream objects to which the track was added, and `onTrack` is called when it is a streamless track. Other than that, they are virtually the same event.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -455,6 +369,125 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+        // at the JS layer, which are the same stream objects passed down to 
+        // the `track` event.
+
+        WritableMap params = Arguments.createMap();
+        params.putInt("id", this.id);
+        params.putString("trackId", receiver.track().id());
+        webRTCModule.sendEvent("peerConnectionOnRemoveTrack", params);
+        
+        params.putBoolean("mute", true);
+        webRTCModule.sendEvent("mediaStreamTrackOnMuteChanged", params);
```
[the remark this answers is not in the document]
Yousif-CS: https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-removetrack quotes: 
"_When the other peer stops sending a track in this manner, the track is removed from any remote [MediaStream](https://www.w3.org/TR/mediacapture-streams/#dom-mediastream)s that were initially revealed in the [track](https://w3c.github.io/webrtc-pc/#event-track) event, and if the [MediaStreamTrack](https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack) is not already muted, a mute event is fired at the track._"

Indeed, we can emit this event on the JS layer.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -528,4 +561,104 @@ private String signalingStateString(PeerConnection.SignalingState signalingState
...
+    }
+
+    private ReadableMap serializeTrack(MediaStreamTrack track) {
+        WritableMap trackInfo = Arguments.createMap();
+        trackInfo.putString("id", track.id());
+        trackInfo.putInt("peerConnectionId", this.id);
+        if (track.kind().equals("video")) {
+            trackInfo.putString("label", "Video");
+        } else if (track.kind().equals("audio")) {
+            trackInfo.putString("label", "Aideo");
```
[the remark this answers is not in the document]
Yousif-CS: This was based off an older PR, I am not sure a proper label for the track source should be on the native side as there is no way to query that information there. I will leave the label as empty strings

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -528,4 +561,104 @@ private String signalingStateString(PeerConnection.SignalingState signalingState
...
+        } else {
+            throw new Error("Unknown kind: " + track.kind());
+        }
+        trackInfo.putString("kind", track.kind());
+        trackInfo.putBoolean("enabled", track.enabled());
+        trackInfo.putString("readyState", track.state().toString());
```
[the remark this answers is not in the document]
Yousif-CS: Yes, but on the Web APIs it is either one of "live" or "ended", this way converting it toString() is appropriate.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -528,4 +561,104 @@ private String signalingStateString(PeerConnection.SignalingState signalingState
...
+        res.putString("id", transceiver.getSender().id());
+        res.putInt("peerConnectionId", this.id);
+        String mid = transceiver.getMid();
+        if (mid != null) {
+            res.putString("mid", mid);
+        }
+        res.putString("direction", serializeDirection(transceiver.getDirection()));
+        RtpTransceiver.RtpTransceiverDirection currentDirection = transceiver.getCurrentDirection();
+        if (currentDirection != null) {
+            res.putString("currentDirection", serializeDirection(transceiver.getCurrentDirection()));
```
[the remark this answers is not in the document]
Yousif-CS: It currently is initialised to null in JS

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -485,6 +513,243 @@ MediaConstraints constraintsForOptions(ReadableMap options) {
...
+                RtpTransceiver transceiver = null;
+                if (options.hasKey("type")) {
+                    String kind = options.getString("type");
+                    transceiver = pco.addTransceiver(mediaTypeForString(kind), parseTransceiverOptions(options.getMap("init")));
+                } else if (options.hasKey("trackId")) {
+                    String trackId = options.getString("trackId");
+                    MediaStreamTrack track = getTrack(trackId);
+                    transceiver = pco.addTransceiver(track, parseTransceiverOptions(options.getMap("init")));
+
+                }
```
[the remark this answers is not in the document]
Yousif-CS: I made sure that does not happen on the JS side. I will add a check here as well

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -485,6 +513,243 @@ MediaConstraints constraintsForOptions(ReadableMap options) {
...
+                // We need the timestamp to reorder the transceivers array 
+                Long tsLong = System.currentTimeMillis()/1000;
```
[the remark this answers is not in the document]
Yousif-CS: I don't think it is possible due to us needing to persist the same transceivers objects on the JS side instead of replacing them.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -485,6 +513,243 @@ MediaConstraints constraintsForOptions(ReadableMap options) {
...
+                params.putMap("transceiver", pco.serializeTransceiver(transceiver));
+                return params;
+            }).get();
+        } catch (InterruptedException | ExecutionException e) {
+            Log.d(TAG, "peerConnectionAddTransceiver() " + e.getMessage());
+            return null;
+        }
+    }
+
+    @ReactMethod(isBlockingSynchronousMethod = true)
```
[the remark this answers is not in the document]
Yousif-CS: It is not really possible since we need to return an RTCRtpSender object on the web API that could potentially be created with this call.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -611,6 +868,30 @@ public void mediaStreamTrackSwitchCamera(String id) {
         });
     }
 
+    /**
+     * This serializes the transceivers current directions and sends them to the
+     * JS layer for potential update
+     */
+    private void sendTransceiverCurrentDirectionUpdate(int id) {
```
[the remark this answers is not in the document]
Yousif-CS: this method was intended to follow the spec to allow us to update the current direction when create/set offer/answer happens. I will add the mid field as well as I believe it is necessary.

src/EventEmitter.ts
```diff
@@ -1,7 +1,28 @@
...
 
 const EventEmitter = new NativeEventEmitter(WebRTCModule);
 
 export default EventEmitter;
+
+const _subscriptions: Record<any, EmitterSubscription[]> = {}; 
+
+type EventHandler = (event: any) => void;
+
+export function addListener(listener: any, eventName: string, eventHandler: EventHandler): void {
```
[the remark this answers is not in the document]
Yousif-CS: We cannot store EmitterSubscription objects inside listeners as it causes a cyclic dependency (it breaks loggers, for example when dealing with JSON.stringify())

src/MediaStreamTrack.ts
```diff
@@ -15,20 +16,23 @@ class MediaStreamTrack extends defineCustomEventTarget(...MEDIA_STREAM_TRACK_EVE
...
-
+    
     constructor(info) {
         super();
 
         this._constraints = info.constraints || {};
         this._enabled = info.enabled;
         this._settings = info.settings || {};
         this._muted = false;
+        this._peerConnectionId = info.peerConnectionId;
```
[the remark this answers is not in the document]
Yousif-CS: It is, and it is a valid use case, however I don't think it is currently supported in react-native-webrtc? For example, I know that track Ids are not necessarily unique for different peer connections, so if we want to be able to addTrack across different peer connections, we would need a unique identifier like a react tag or something similar.

src/RTCRtpReceiver.ts
Yousif-CS: Could you elaborate more on that? the native side currently does not use AudioEncoderFactory so it wasn't possible to use the same hack for video.

src/RTCRtpTransceiver.ts
Yousif-CS: You are right.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -105,15 +74,6 @@ void close() {
         // Close the PeerConnection first to stop any events.
         peerConnection.close();
 
-        // PeerConnection.dispose() calls MediaStream.dispose() on all local
```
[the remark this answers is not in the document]
Yousif-CS: the implementation of `removeStream` was removed in earlier commits made on this branch, I assumed that in this case the call to `removeStream` was unnecessary (it was failing to compile). I will add it back along with the function implementation.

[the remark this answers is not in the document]
saghul: Please check what PeerConnection.dispose does now, it's possible this is indeed no longer necessary, but we need to check.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -455,6 +369,125 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
     @Override
     public void onAddTrack(final RtpReceiver receiver, final MediaStream[] mediaStreams) {
         Log.d(TAG, "onAddTrack");
+
+        // Get The timestamp so that we can reorder the transceivers on the JS layer
```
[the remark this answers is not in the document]
saghul: Yes but if you keep the JS and native in sync then they have matching indexes in the transceivers array don't they?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
saghul: Right. But is it possible for it to be null inside onAddTrack?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
saghul: Got it.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
saghul: Sounds good!

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
saghul: What about updates? Can it transition to null?

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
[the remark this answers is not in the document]
saghul: Yeah. Let's Theo an exception and leave a comment that JS checks for it.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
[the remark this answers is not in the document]
saghul: What I had in mind here is that you'd then match them and update them as necessary. Not replace them.

src/EventEmitter.ts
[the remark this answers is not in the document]
saghul: I see. You mean if logging a full track or sender for example?

src/MediaStreamTrack.ts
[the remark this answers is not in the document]
saghul: I think it's currently possible.

Loca Tracks are not referenced by PC but by their ID, which is guaranteed to be unique because we generate it (unlike the remote ones, for local ones The id and react tag are the same) using UUID v4.

src/RTCRtpReceiver.ts
saghul: I mean that we could just hard code opus, isac, etc since the WebRTC build which this plugin uses will have those for sure.

Or maybe there is a way to query it, but I don't k on right off the bat.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -138,6 +98,62 @@ void close() {
...
+        }
+
+        if (peerConnection == null) {
+            throw new Error("Cannot Add transceiver: peer connection is null");
+        }
+
+        return peerConnection.addTransceiver(track, init);
+    }
+
+    String resolveTransceiverId(RtpTransceiver transceiver) {
```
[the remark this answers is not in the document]
Yousif-CS: According to the spec, the sender is always created nonetheless. Here is the function definition from native WebRTC: 
```
/**
 * The sender attribute exposes the RtpSender corresponding to the RTP media
 * that may be sent with the transceiver's mid. The sender is always present,
 * regardless of the direction of media.
 * https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-sender
 */
public RtpSender getSender() {
  return cachedSender;
}
```

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
Yousif-CS: Although the enum is converted to an upper case string. I will fix that.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -485,6 +513,243 @@ MediaConstraints constraintsForOptions(ReadableMap options) {
...
+            case "recvonly":
+                return RtpTransceiver.RtpTransceiverDirection.RECV_ONLY;
+            case "inactive":
+                return RtpTransceiver.RtpTransceiverDirection.INACTIVE;
+        }
+        throw new Error("Invalid direction");
+    }
+
+    private RtpTransceiver.RtpTransceiverInit parseTransceiverOptions(ReadableMap map) {
+        RtpTransceiver.RtpTransceiverDirection direction = RtpTransceiver.RtpTransceiverDirection.SEND_RECV;
```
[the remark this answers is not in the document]
Yousif-CS: I will change this to `INACTIVE`

src/MediaStreamTrack.ts
```diff
@@ -97,10 +101,19 @@ class MediaStreamTrack extends defineCustomEventTarget(...MEDIA_STREAM_TRACK_EVE
     getSettings() {
         return deepClone(this._settings);
     }
-
+    
     release(): void {
         WebRTCModule.mediaStreamTrackRelease(this.id);
     }
+
+    _registerEvents(): void {
```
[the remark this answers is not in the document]
Yousif-CS: Inside the constructor of MediaStreamTrack. I will fix that.

src/RTCRtpCapabilities.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (ev.codecs === 'undefined') {
+            throw new Error('Invalid event object passed to senderGetCapabilities: codecs is undefined');
+        }
+        _receiverCapabilities = new RTCRtpCapabilities(ev.codecs);
+    }),
+]
+
+// Initialize capabilities on module import
+if (!_senderCapabilities) {
+    WebRTCModule.senderGetCapabilities();
```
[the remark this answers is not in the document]
Yousif-CS: It is possible.

src/RTCRtpReceiver.ts
Yousif-CS: I couldn't find any method afaik. I'd say we hard code them for now.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
Yousif-CS: Oh I don't think that was the intention. The idea is if addTransceiver(...) returns a transceiver object, that exact object has to be the one receiving updates. IMO it would be better to add transceivers one by one. Otherwise, if we serialize the whole transceivers array, the only thing we would gain from it is their order. I think adding one timestamp to the returned object is probably more efficient than a whole array of transceivers, don't you think?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
Yousif-CS: I guess not, I wrote this from a defensive point of view :) I will update it.

src/EventEmitter.ts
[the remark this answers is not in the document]
Yousif-CS: Yes.

src/MediaStreamTrack.ts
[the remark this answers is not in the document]
Yousif-CS: Thats a valid use case, but I am referring to an SFU use case. For example, adding a remote track from one peer connection to another. This might not be possible in this current setting.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
Yousif-CS: You are correct. I just checked the spec. when stop is called, currentDirection should be set to null. However, that is already handled through the "transceiverOnStopSuccessful" and currentDirection is set to null on the JS side.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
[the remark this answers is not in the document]
Yousif-CS: Right.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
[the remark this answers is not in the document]
Yousif-CS: I don't mind making this change, but my rationale was that its unnecessary overhead to serialise all transceivers every time where we can just send a timestamp and use that to sort their order, don't you think? However, the last call is yours.

src/RTCPeerConnection.ts
```diff
@@ -171,6 +159,101 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+    }
+
+    addTransceiver(source: 'audio' | 'video' | MediaStreamTrack, init): RTCRtpTransceiver {
+        let src = {};
+        if (source === 'audio') {
+            src = { type: 'audio' };
+        } else if (source === 'video') {
+            src = { type: 'video' };
+        } else {
+            src = { trackId: source.id };
```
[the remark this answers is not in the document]
Yousif-CS: The argument is one of 'audio' | 'video' | MediaStreamTrack. The track is never allowed to be null if that is what you are referring to?

src/RTCPeerConnection.ts
```diff
@@ -188,144 +271,180 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+            const channel = new RTCDataChannel(ev.dataChannel);
+            // @ts-ignore
+            this.dispatchEvent(new RTCDataChannelEvent('datachannel', { channel }));
+        });
+
+        // Since the current underlying architecture performs certain actions
+        // Asynchronously when the outer web API expects synchronous behaviour
+        // This is the only way to report error on operations for those who wish
+        // to handle them.
+        addListener(this, 'peerConnectionOnError', ev => {
```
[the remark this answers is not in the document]
Yousif-CS: WebRTCModule.peerConnectionRemoveTrack now does.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
saghul: Gotcha. In that case how about this: each PC will start with a transceiver counter of 0. Every time a new transceiver is crated we use it and increment it. I think the code will be more concise and easier to follow like that.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +362,154 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+        // Getting the transceiver object associated with the receiver for the event
+        List<RtpTransceiver> transceivers = peerConnection.getTransceivers();
+        for( RtpTransceiver transceiver : transceivers ) {
+            if(transceiver.getReceiver() != null && receiver.id().equals(transceiver.getReceiver().id())) {
+                eventInfo.putMap("transceiver", serializeTransceiver(transceiver));
+                break;
+            }
+        }
+        params.putInt("id", this.id);
+        params.putMap("info", eventInfo);
```
saghul, line 420: Does this add anything? Maybe we could flatten the structure and send all the fields in `params` ?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +362,154 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
     public void onAddTrack(final RtpReceiver receiver, final MediaStream[] mediaStreams) {
         Log.d(TAG, "onAddTrack");
+
+        // Get The timestamp so that we can reorder the transceivers on the JS layer
+        Long tsLong = System.currentTimeMillis()/1000;
+        String ts = tsLong.toString();
+
+        MediaStreamTrack track = receiver.track();
+
+        if (track == null) {
```
saghul, line 379: This should be impossible here, since the receiver is live.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +362,154 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+            return;
+        }
+        if(track.kind().equals(MediaStreamTrack.VIDEO_TRACK_KIND)){
+            videoTrackAdapters.addAdapter((VideoTrack) track);
```
saghul, line 383: Remote tracks are disposed when the RtpReceiver is disposed. We need to make sure we have removed the adapter by then or we'll be holding on to a released remote track.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +362,154 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
         webRTCModule.sendEvent("peerConnectionSignalingStateChanged", params);
     }
 
+    /**
+     * Triggered when a new track is signaled by the remote peer, as a result of
+     * setRemoteDescription.
+     */
     @Override
     public void onAddTrack(final RtpReceiver receiver, final MediaStream[] mediaStreams) {
```
saghul, line 370: Checked the C++ code, this is what the comment says for `OnAddTrack`:

```
  // This is called when a receiver and its track are created.
  // TODO(zhihuang): Make this pure virtual when all subclasses implement it.
  // Note: This is called with both Plan B and Unified Plan semantics. Unified
  // Plan users should prefer OnTrack, OnAddTrack is only called as backwards
  // compatibility (and is called in the exact same situations as OnTrack).
```

So I'd say we drop our `onAddTrack` handler and just use `onTrack`.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +362,154 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+
+        // Get The timestamp so that we can reorder the transceivers on the JS layer
+        Long tsLong = System.currentTimeMillis()/1000;
+        String ts = tsLong.toString();
+
+        if (track == null) {
+            return;
+        }
+        if(track.kind().equals(MediaStreamTrack.VIDEO_TRACK_KIND)){
+            videoTrackAdapters.addAdapter((VideoTrack) track);
```
saghul, line 445: Ditto about the adapter.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +362,154 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+        WritableArray streams = Arguments.createArray();
+        
+        eventInfo.putArray("streams", streams);
+        eventInfo.putMap("receiver", serializeReceiver(receiver));
+        eventInfo.putMap("transceiver", serializeTransceiver(transceiver));
+        eventInfo.putString("timestamp", ts);
+
+        // Getting the transceiver object associated with the receiver for the event
+        params.putInt("id", this.id);
+        params.putMap("info", eventInfo);
```
saghul, line 460: Let's flatten it.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +362,154 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+        webRTCModule.sendEvent("peerConnectionOnTrack", params);
+    }
+
+    /**
+     * Triggered when the signaling from SetRemoteDescription indicates that a transceiver
+     * will be receiving media from a remote endpoint. This is only called if UNIFIED_PLAN
+     * semantics are specified. The transceiver will be disposed automatically.
+     */
+    @Override
+    public void onTrack(final RtpTransceiver transceiver) {
```
saghul, line 431: What thread does this run on? It should probably be deferred to the executor so accessing any things on the instance is safe.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +362,154 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+
+        webRTCModule.sendEvent("peerConnectionOnTrack", params);
+    }
+
+    /*
+     * Triggered when a previously added remote track is removed by the remote
+     * peer, as a result of setRemoteDescription.
+     */
+    @Override
+    public void onRemoveTrack(RtpReceiver receiver){
```
saghul, line 470: If the track is killed at this point we need to dispose of the adapter.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +362,154 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+     * Removes a specific local <tt>MediaStream</tt> from the associated
+     * <tt>PeerConnection</tt>.
+     *
+     * @param localStream the local <tt>MediaStream</tt> from the associated
+     *                    <tt>PeerConnection</tt>
+     * @return <tt>true</tt> if removing the specified <tt>mediaStream</tt> from
+     * this instance resulted in a modification of its internal list of local
+     * <tt>MediaStream</tt>s; otherwise, <tt>false</tt>
+     */
+    boolean removeStream(MediaStream localStream) {
```
saghul, line 507: Do we need this?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -528,4 +583,102 @@ private String signalingStateString(PeerConnection.SignalingState signalingState
...
         return null;
     }
+
+    /**
+     * Serialization and Parsing helpers
+     */
+
+    private WritableMap serializeStream(String streamReactTag, MediaStream stream) {
+
+        Log.d(TAG, "streamReactTag = " + streamReactTag);
```
saghul, line 593: Drop these at the end.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -528,4 +583,102 @@ private String signalingStateString(PeerConnection.SignalingState signalingState
...
+            default:
+                throw new Error("Invalid direction");
+        }
+    }
+
+    private ReadableMap serializeTrack(MediaStreamTrack track) {
+        WritableMap trackInfo = Arguments.createMap();
+        trackInfo.putString("id", track.id());
+        trackInfo.putInt("peerConnectionId", this.id);
+        if (track.kind().equals("video")) {
```
saghul, line 632: I'd say we leave this out of native and just set it to the empty string in JS

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -485,6 +497,281 @@ MediaConstraints constraintsForOptions(ReadableMap options) {
...
+                                             ReadableMap options) {
+        try {
+            return (WritableMap) ThreadUtils.submitToExecutor((Callable<Object>) () -> {
+                PeerConnectionObserver pco = mPeerConnectionObservers.get(id);
+                if (pco == null) {
+                    Log.d(TAG, "peerConnectionAddTrack() peerConnection is null");
+                    return null;
+                }
+
+                MediaStreamTrack track = getTrack(trackId);
```
saghul, line 609: use getLocalTrack

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -485,6 +497,281 @@ MediaConstraints constraintsForOptions(ReadableMap options) {
...
+                List<String> streamIds = new ArrayList<>();
+                if (options.hasKey("streamIds")) {
+                    ReadableArray rawStreamIds = options.getArray("streamIds");
+                    if (rawStreamIds != null) {
+                        for (int i = 0; i < rawStreamIds.size(); i++) {
+                            streamIds.add(rawStreamIds.getString(i));
+                        }
+                    }
+                }
+                RtpSender sender = pco.getPeerConnection().addTrack(track, streamIds);
```
saghul, line 619: We have a little problem which might require patching WebRTC. The returned `RtpSender` owns the track and will dispose it when the sender itself goes away.

I don't see how to change this. There is a `setTrack` API in the sender which allows to replace the track, and there is a `takeOwnership` parameter, but that will still dispose the previous track...

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -485,6 +497,281 @@ MediaConstraints constraintsForOptions(ReadableMap options) {
...
+
+                // Need to get the corresponding transceiver as well
+                RtpTransceiver transceiver = pco.getTransceiver(sender.id());
+
+                if (transceiver == null) {
```
saghul, line 624: How could this happen?

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
saghul: Maybe not. Those objects are refcounted so it's just our instance of the track that will be gone. We need a fresh native object that represents the same entity. Not sure how to do that (yet) though.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -560,15 +847,7 @@ public void mediaStreamRelease(String id) {
                 Log.d(TAG, "mediaStreamRelease() stream is null");
                 return;
             }
-
             localStreams.remove(id);
-
-            // MediaStream.dispose() may be called without an exception only if
-            // it's no longer added to any PeerConnection.
-            for (int i = 0, size = mPeerConnectionObservers.size(); i < size; i++) {
-                mPeerConnectionObservers.valueAt(i).removeStream(stream);
```
saghul, line 569: We might need to removeTrack instead here, in order to accomplish the same. Even though I no longer see why MediaStream.dispose() would fail...

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -611,6 +890,33 @@ public void mediaStreamTrackSwitchCamera(String id) {
         });
     }
 
+    /**
+     * This serializes the transceivers current direction and mid and returns them
+     * for update when an sdp negotiation/renegotiation happens 
+     */
+    private ReadableArray getTransceiversSdpInfo(int id) {
```
saghul, line 897: Let's be a bit more generic and call it just info? We can update a number of things.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -818,6 +1106,60 @@ public void onSetFailure(String s) {
...
+        for(VideoCodecInfo codecInfo: videoCodecInfos) {
+            codecs.pushMap(mapForVideoCodecInfo(codecInfo));
+        }
+        params.putArray("codecs", codecs);
+        return params;
+
+    }
+
+    @ReactMethod(isBlockingSynchronousMethod = true)
+    public WritableMap receiverGetCapabilities() {
```
saghul, line 1139: This was only called once, right?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: That seems right. However, there is an issue with regards to fetching the Receiver's associated MediaStreams. It seems that on the C++ side, there is a function to do that referenced here 
```cpp
  // This is called when signaling indicates a transceiver will be receiving
  // media from the remote endpoint. This is fired during a call to
  // SetRemoteDescription. The receiving track can be accessed by:
  // `transceiver->receiver()->track()` and its associated streams by
  // `transceiver->receiver()->streams()`.
  // Note: This will only be called if Unified Plan semantics are specified.
  // This behavior is specified in section 2.2.8.2.5 of the "Set the
  // RTCSessionDescription" algorithm:
  // https://w3c.github.io/webrtc-pc/#set-description
  virtual void OnTrack(
      rtc::scoped_refptr<RtpTransceiverInterface> transceiver) {}

```

which exists inside `rtp_transceiver_interface.h`: 
```cpp

  // The list of streams that `track` is associated with. This is the same as
  // the [[AssociatedRemoteMediaStreams]] internal slot in the spec.
  // https://w3c.github.io/webrtc-pc/#dfn-associatedremotemediastreams
  // TODO(hbos): Make pure virtual as soon as Chromium's mock implements this.
  // TODO(https://crbug.com/webrtc/9480): Remove streams() in favor of
  // stream_ids() as soon as downstream projects are no longer dependent on
  // stream objects.
  virtual std::vector<std::string> stream_ids() const;
  virtual std::vector<rtc::scoped_refptr<MediaStreamInterface>> streams() const;

```
while I couldn't find that on the Android native side. Any ideas?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: It turns out, no. We are not modifying storing local media streams inside the observer anymore.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
Yousif-CS: I am not quite sure this is the case. This function snippet from `RtpSender.java` seems to suggest the track is only disposed when it is owned by the sender (i.e. when setTrack has been used). 
```
  public void dispose() {
    checkRtpSenderExists();
    if (dtmfSender != null) {
      dtmfSender.dispose();
    }
    if (cachedTrack != null && ownsTrack) {
      cachedTrack.dispose();
    }
    JniCommon.nativeReleaseRef(nativeRtpSender);
    nativeRtpSender = 0;
  }
```

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
Yousif-CS: It was written defensively. I will remove it.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
Yousif-CS: Yes, I made it a module constant on the typescript side that only gets called once, so it made sense to have it be blocking.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
Yousif-CS: Internally, `MediaStream.dispose()` calls `removeTrack`. I do not see either why we cannot use `MediaStream.dispose()` here. It currently is being used.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
Yousif-CS: It is not needed to call removeStream as that API is deprecated.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: I made it run on the executor.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
Yousif-CS: Done.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: Done

src/RTCRtpReceiver.ts
Yousif-CS: Done

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: Removed.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
Yousif-CS: Done.

src/RTCPeerConnection.ts
```diff
@@ -171,6 +159,101 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+        if (!successful) throw new Error("Error removing track from sender");
+        existingSender._track = null;
+
+        const [existingTransceiver] = this
+            .getTransceivers()
+            .filter((t) => t.sender.id === existingSender.id);
+        existingTransceiver._direction = existingTransceiver.direction === 'sendrecv' ? 'recvonly' : 'inactive'; 
+    }
+
     getStats() {
```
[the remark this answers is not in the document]
Yousif-CS: Is this referring to finding the existingSender? i.e. changing: 
```ts
const existingSender = this
    .getSenders()
    .find((s) => s === sender);
``` 
to use `filter` instead?

Yousif-CS (commented):
Hey Saghul, have a look at the updates and let me know what you think.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: The spec does not say anything about destroying remote tracks after receiving an `onRemoveTrack` event. It mentions muting them if they are not already. So, I think it is just safest to destroy the adapters at the end of the peer connection session

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
saghul: I see it's indeed missing in both Android and iOS. I'd say we make a patch to WebRTC adding them. Thoughts?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
saghul: Ack.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
saghul: So there will be a bunch of zombie MediaStreamTrack objects? That sounds weird. Specially since transceivers can be reused, right? Can you please double check with a fiddle in Chrome?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
saghul: Please verify what Chrome does.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
saghul: But look at the constructor and the default value for `ownsTrack`:

```
/** Java wrapper for a C++ RtpSenderInterface. */
public class RtpSender {
  private long nativeRtpSender;

  @Nullable private MediaStreamTrack cachedTrack;
  private boolean ownsTrack = true;
  private final @Nullable DtmfSender dtmfSender;

  @CalledByNative
  public RtpSender(long nativeRtpSender) {
    this.nativeRtpSender = nativeRtpSender;
    long nativeTrack = nativeGetTrack(nativeRtpSender);
    cachedTrack = MediaStreamTrack.createMediaStreamTrack(nativeTrack);

    long nativeDtmfSender = nativeGetDtmfSender(nativeRtpSender);
    dtmfSender = (nativeDtmfSender != 0) ? new DtmfSender(nativeDtmfSender) : null;
  }
```

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
saghul: 👍

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
saghul: Wait, this is weird. I checked the API and it doesn't mention any streams property. So AFAIS there is no way to get a reference to the stream from a transceiver object. Am I missing something?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +359,124 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+        // According to the W3C spec, we need to send out 
+        // the track Id so that we can remove it from the MediaStream objects stored
+        // at the JS layer, which are the same stream objects passed down to 
+        // the `track` event.
+        ThreadUtils.runOnExecutor(() -> {
+            MediaStreamTrack track = receiver.track();
+            WritableMap params = Arguments.createMap();
+            params.putInt("id", this.id);
+            params.putString("trackId", track.id());
+            params.putBoolean("mute", true);
```
saghul, line 463: Do this in JS

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -528,4 +550,94 @@ private String signalingStateString(PeerConnection.SignalingState signalingState
         }
         return null;
     }
+
+    /**
+     * Serialization and Parsing helpers
+     */
+
```
saghul, line 557: Drop the extraneous newlines

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -528,4 +550,94 @@ private String signalingStateString(PeerConnection.SignalingState signalingState
...
+    /**
+     * This method is currently missing serializing DtmfSender
+     * and transport.
+     * TODO: Add transport and dtmf fields to the serialized sender to match the web APIs.
+     */
+    public ReadableMap serializeSender(RtpSender sender) {
+        WritableMap res = Arguments.createMap();
+        res.putString("id", sender.id());
+        res.putInt("peerConnectionId", this.id);
+        if (sender.track() != null)
```
saghul, line 614: use braces around the if pl even if it's a single statement

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -452,9 +359,124 @@ public void onSignalingChange(PeerConnection.SignalingState signalingState) {
...
+            WritableMap params = Arguments.createMap();
+            WritableArray streams = Arguments.createArray();
+            
+            params.putArray("streams", streams);
+            params.putMap("receiver", serializeReceiver(receiver));
+            params.putMap("transceiver", serializeTransceiver(transceiver));
+            params.putInt("transceiverOrder", getNextTransceiverId());
+            params.putInt("id", this.id);
+
+            webRTCModule.sendEvent("peerConnectionOnTrack", params);
```
saghul, line 444: Do we also need to update the SDP representations in the JS side? I think so?

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
saghul: So the following test: create a local video track, display it in a view. Create a PC and add it. Destroy the PC. Try to render the track again. Does it work?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
saghul: Ah right. We need it for the event. So weird.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -485,6 +497,275 @@ MediaConstraints constraintsForOptions(ReadableMap options) {
...
+            }).get();
+        } catch (InterruptedException | ExecutionException e) {
+            Log.d(TAG, "peerConnectionAddTrack() " + e.getMessage());
+            return null;
+        }
+    }
+
+    @ReactMethod
+    public void peerConnectionRemoveTrack(int id,
+                                            String senderId) {
```
saghul, line 639: Can you put this on the same line? Use 120 for the (soft) limit.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -485,6 +497,275 @@ MediaConstraints constraintsForOptions(ReadableMap options) {
...
+                }
+
+                RtpSender sender = transceiver.getSender();
+                MediaStreamTrack track = getLocalTrack(trackId);
+                //TODO: Discuss local track ownership.
+                sender.setTrack(track, false);
+                promise.resolve(true);
+            } catch (Exception e) {
+                Log.d(TAG, "senderReplaceTrack(): " + e.getMessage());
+                promise.reject(new Exception(e.getMessage()));
```
saghul, line 728: Why not just reject it with `e` ?

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -676,8 +944,11 @@ public void onCreateFailure(String s) {
                 @Override
                 public void onCreateSuccess(SessionDescription sdp) {
                     WritableMap params = Arguments.createMap();
-                    params.putString("sdp", sdp.description);
-                    params.putString("type", sdp.type.canonicalForm());
+                    WritableMap sdpInfo = Arguments.createMap();
+                    sdpInfo.putString("sdp", sdp.description);
+                    sdpInfo.putString("type", sdp.type.canonicalForm());
+                    params.putArray("transceiverUpdates", getTransceiversInfo(id));
```
saghul, line 950: transceiverInfo? for symmetry with sdpInfo

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -818,6 +1098,60 @@ public void onSetFailure(String s) {
         });
     }
 
+    private WritableMap mapForVideoCodecInfo(VideoCodecInfo info) {
+        WritableMap params = Arguments.createMap();
+        params.putString("mimeType", "video/" + info.name);
+        return params;
+    }
+
+    private WritableMap doReceiverGetCapabilities() {
```
saghul, line 1107: this is only used in a single place, why not inline it?

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -818,6 +1098,60 @@ public void onSetFailure(String s) {
...
+        WritableMap params = Arguments.createMap();
+        WritableArray codecs = Arguments.createArray();
+        for(VideoCodecInfo codecInfo: videoCodecInfos) {
+            codecs.pushMap(mapForVideoCodecInfo(codecInfo));
+        }
+        params.putArray("codecs", codecs);
+        return params;
+    }
+
+    private WritableMap doSenderGetCapabilities() {
```
saghul, line 1118: ditto

android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
```diff
@@ -189,6 +189,9 @@ private VideoTrack getVideoTrackForStreamURL(String streamURL) {
             }
         }
 
+        if (videoTrack == null) {
```
saghul, line 192: IS this necessary?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: Have a look at [this](https://jsfiddle.net/pub2kjmf/). I tested it out on Chrome and it seems that the track still exists after a call to removeTrack. You will see that `onmute` but `onended` is not called. Also, looking at the chromium webrtc source: 
```cpp
// Called when signaling indicates that media will no longer be received on a
// track.
// With Plan B semantics, the given receiver will have been removed from the
// PeerConnection and the track muted.
// With Unified Plan semantics, the receiver will remain but the transceiver
// will have changed direction to either sendonly or inactive.
// https://w3c.github.io/webrtc-pc/#process-remote-track-removal
// TODO(hbos,deadbeef): Make pure virtual when all subclasses implement it.
virtual void OnRemoveTrack(
  rtc::scoped_refptr<RtpReceiverInterface> receiver) {}
```

It does not say anything about destroying the track.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: Yep, patching WebRTC seems to be our only option, although for now, it seems to me that onAddTrack still works, so I am not sure if the patch is urgent.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: Verified in a comment above.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -528,4 +550,94 @@ private String signalingStateString(PeerConnection.SignalingState signalingState
...
+
+    private ReadableMap serializeReceiver(RtpReceiver receiver) {
+        WritableMap res = Arguments.createMap();
+        res.putString("id", receiver.id());
+        res.putInt("peerConnectionId", this.id);
+        res.putMap("track", serializeTrack(receiver.track()));
+        return res;
+    }
+
+    public ReadableMap serializeTransceiver(RtpTransceiver transceiver) {
```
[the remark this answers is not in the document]
Yousif-CS: That would be better, but the reason why it is inside the observer is because it has access to the peerConnectionId, it does not matter that much as we can provide it as an argument to the functions. What do you reckon?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: If you are referring to updating the transceivers, then that is already done as we are serialising the transceivers here as well. On the JS side, the objects themselves are being updated.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -611,6 +884,31 @@ public void mediaStreamTrackSwitchCamera(String id) {
...
+    private ReadableArray getTransceiversInfo(int id) {
+        PeerConnectionObserver pco = mPeerConnectionObservers.get(id);
+        if (pco == null) {
+            return null;
+        }
+        PeerConnection peerConnection = pco.getPeerConnection();
+        WritableArray transceiverUpdates = Arguments.createArray();
+        for(RtpTransceiver transceiver: peerConnection.getTransceivers()) {
+            RtpTransceiver.RtpTransceiverDirection direction = transceiver.getCurrentDirection();
+            if (direction == null) continue;
```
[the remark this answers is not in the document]
Yousif-CS: This is an excerpt from W3C: "_If this transceiver has never been represented in an offer/answer exchange, the value is null. If the transceiver is [stopped](https://www.w3.org/TR/webrtc/#dfn-stopped), the value is "stopped"_". I will however remove the check for the sake of uniformity.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
saghul: One leas thing to worry about I guess!

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
saghul: A parameter sounds good!

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
saghul: No I mean the SDPs for pc.RemoteDrscriptipn.sdp and so on.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -48,49 +52,11 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         this.webRTCModule = webRTCModule;
         this.id = id;
         this.dataChannels = new HashMap<>();
-        this.localStreams = new ArrayList<>();
```
saghul, line 51: I just remembered. Without this, how does MediaStream.release() work?

We need this custom API both for local streams and tracks to dispose the native side since we cannot latch on the JS GC to do it then.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: I believe that would have happened prior to this event being invoked based on the description of OnTrack: 
```
    /**
     * Triggered when the signaling from SetRemoteDescription indicates that a transceiver
     * will be receiving media from a remote endpoint. This is only called if UNIFIED_PLAN
     * semantics are specified. The transceiver will be disposed automatically.
     */
```

saghul: Maybe after addTrack / removeTrack ?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: `WebRTCModule` stores a hashmap of local streams? here: 
```java
    final Map<String, MediaStream> localStreams;
```

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: if negotiation is necessary after those calls, the native side would fire up a negotiation needed event that propagates to the JS side, at which point subsequent calls to createOffer and setRemoteDescription should update the sdps, right?

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
Yousif-CS: `mediaStream` functions inside `WebRTCModule` interact with that hashmap for the lifetime of mediaStreams.

saghul: Awesome I had forgotten we store them there! 😅

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
Yousif-CS: I did the following test on the example app `GumTestApp` as follows: I created a replace button which creates a new peer connection everytime and attaches the same track everytime. It was done on an Android Emulator. The app did not crash or cause any issues:
```js
  const [pc, setPc] = useState(new RTCPeerConnection());
  const replace = () => {
    console.log('replace');
    pc.close();
    setPc(new RTCPeerConnection());
    const track = stream.getTracks()[0];
    pc.addTrack(track, [stream]);
  }

```

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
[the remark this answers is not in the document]
Yousif-CS: Done.

saghul:
This is now ready to test on Android.

I'll continue with the iOS implementation.

fcancela:
Hey, great work done over here! 

I wanted to know if you would be able to come up with an estimate on when this work could be expected to be finished. This could really help us to plan our work for next months.

Thanks in advance!

saghul:
I expect to have a release candidate in a matter of weeks.

ovyan:
@saghul c'mon dude, we are really waiting for you, good luck!

vanGalilea:
Hey @saghul and @Yousif-CS 👋🏻 is there a way of expediting this by any means? We would like to help, if possible. This is a very important building block we need.

saghul:
> Hey @saghul and @Yousif-CS 👋🏻 is there a way of expediting this by any means? We would like to help, if possible. This is a very important building block we need.

Hey there! You can test the current branch and provide feedback!

saghul (approved):

PR #1174, opened 2022-08-07
Type RTCView props

IuryPiva:
Typescript does not like when we use any of the props documented for this component, to solve this I've added them to the exported component.


saghul:
Did you intend to close this?

PR #1178, opened 2022-08-30
Eng 5202 control audio session

filipi87:
Improving the way that we are handling audio in order to work better with apps that need to execute other audios at the same time.

More details can be found on this linear [ticket](https://linear.app/dailyco/issue/ENG-5202/provide-better-control-over-our-audio-sessions-on-react-native) and on this previous [PR](https://github.com/daily-co/react-native-webrtc/pull/27) that is now closed.

Changes on this PR:
- Just emitting the audio focus loss event on Android if we are in a call.


saghul:
Gotta love daily's involvement with upstream.

filipi87:
Hi @saghul,

Yes, that would be nice.

Most of our changes are related to providing better control on the audio devices. So we have:
- Changed the enumerateDevices method, to return all the possible options;
- Creating an automatic mode for audio or video calls. Depending on the mode, we choose a different audio route;
- And add support so the user can check what is the current audio route which he is using and change It.

If you wish to take a look on our changes:
https://github.com/react-native-webrtc/react-native-webrtc/compare/1.94.1...daily-co:react-native-webrtc:main

If you think those changes can be worth it to the main fork, I can try to talk with the team, and maybe make a plan for that.

saghul:
Thanks for getting back!



> * Changed the enumerateDevices method, to return all the possible options;

I'd be curious to know, that sounds like an uncontroversial change which we should be able to land.

> * Creating an automatic mode for audio or video calls. Depending on the mode, we choose a different audio route;
> 
> * And add support so the user can check what is the current audio route which he is using and change It.

Since apps may have different opinions on what they want to do, I went out of my way to do that on the application layer. Our (Jitsi's) audio management is separate from this plugin: https://github.com/jitsi/jitsi-meet/blob/master/ios/sdk/src/AudioMode.m and https://github.com/jitsi/jitsi-meet/blob/master/android/sdk/src/main/java/org/jitsi/meet/sdk/AudioModeModule.java

> * 
> 
> 
> If you wish to take a look on our changes: [1.94.1...daily-co:react-native-webrtc:main](https://github.com/react-native-webrtc/react-native-webrtc/compare/1.94.1...daily-co:react-native-webrtc:main)
> 
> If you think those changes can be worth it to the main fork, I can try to talk with the team, and maybe make a plan for that.

I think it would be in Daily's best interest to use upstream to get all fixes (including security) and features instead of having to deal with merges.

Had I taken the same road Daily has a few years ago this plugin would've been much different, or just be unmaintained.

PR #1187, opened 2022-10-05
Some fixes and sendEncodings for Android

8BallBomBom:
I need to add `sendEncodings` for iOS but i don't have an Apple device to test with currently.
Also replaced the use of `forEach` due to complaints of requiring a minimum Android SDK of 24 but the project minimum is 21.
Then just some random formatting changes.

Including the following on the init array for a sending Transceiver seems to give the correct `SDP offer`.
```
sendEncodings: [
	{
		rid: 'h',
		maxBitrate: 1200 * 1024
	},
	{
		rid: 'm',
		maxBitrate: 600 * 1024,
		scaleResolutionDownBy: 2
	},
	{
		rid: 'l',
		maxBitrate: 300 * 1024,
		scaleResolutionDownBy: 4
	}
]
```

The resulting `SDP` now includes the following.
```
a=rid:h send
a=rid:m send
a=rid:l send
a=simulcast:send h;m;l
```
I haven't gone as far as to actually test if simulcasting actually works or needs extra implementation.


saghul:
Good effort, thanks @8BallBomBom !

> I need to add `sendEncodings` for iOS but i don't have an Apple device to test with currently.

I can help with the testing, making adjustments.

> Also replaced the use of `forEach` due to complaints of requiring a minimum Android SDK of 24 but the project minimum is 21.

Hum, that's weird, the necessary bytecode version is 1.8, even though more modern features can be used. We haven't seen this problem in Jitsi Meet, but our minimum is 23. Can you please share the error?



> Then just some random formatting changes and removed imports/references that are not required.

Can you please split those out of this PR? We could land this and the above while you continue the work on setEncodings.



> Including the following on the init array for a sending Transceiver seems to give the correct `SDP offer`.

Nice!

saghul:
Looks like a Gradle bug: https://stackoverflow.com/questions/68046599/lint-is-giving-error-if-i-use-java-lang-iterableforeach-in-android-code

So not a real compilation bug.

8BallBomBom:
Do you want me to ignore it or include the changes with the new PR?

8BallBomBom:
Was expecting that to keep things open, no worries i'll do a new PR for the simulcast stuff and then we could address formatting/layout issues in a more wide spread PR at some point 👍🏻

PR #1191, opened 2022-10-10
Fix parameter types

saghul:


src/RTCRtcpParameters.ts
```diff
@@ -1,14 +1,23 @@
+export interface RTCRtcpParametersInit {
```
tmoldovan8x8, line 1: typo, extra c

tmoldovan8x8:
LGTM

src/RTCRtcpParameters.ts
saghul: Nope, that's correct because it's the RTCP parameters, not the RTP ones. Confusing, I know!

Issue #1197, opened 2022-10-14
RTCPeerConnection.addTrack signature inconsistent


whitej14-hrc:
#### Expected Behavior

In a browser environment, `RTCPeerConnection.addTrack` can be called with the rest parameter syntax, i.e., `RTCPeerConnection.addTrack(track, stream1, stream2)`

See https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addTrack

#### Observed Behavior

When using react-native-webrtc, an array is expected for the second argument and the rest parameter syntax is not supported.

#### Steps to reproduce the issue

1. Create an RTCPeerConnection
2. Add a track with rest parameter syntax for second argument
3. An error with message `streams.map is undefined` will be thrown

#### Platform Information

* **React Native Version**: 0.70
* **Plugin Version**: 106.0.0-beta.5
* **OS**: iPad OS
* **OS Version**: 15

whitej14-hrc:
Thank you folks, I appreciate the quick turnaround 🙏

8BallBomBom:
Welcome, thanks for contributing 😃

PR #1199, opened 2022-10-14
Fix RTCPeerConnection.addTrack signature

whitej14-hrc:
See issue: https://github.com/react-native-webrtc/react-native-webrtc/issues/1197

This PR updates the signature of RTCPeerConnection.addTrack() to allow a rest parameter for the second argument, accepting one or more `MediaStream`s rather than an array of `MediaStream`s.

https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addTrack

Additionally, this PR updates `registerGlobals()` to expose the `MediaStreamTrackEvent` interface in the global context. This should improve compatibility with libraries like [SIP.js](https://github.com/onsip/SIP.js) which expect this interface to be available globally (example usage [here](https://github.com/onsip/SIP.js/blob/main/src/platform/web/session-description-handler/session-description-handler.ts#L181)).

https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrackEvent

### DIFF
--- a/src/RTCPeerConnection.ts
+++ b/src/RTCPeerConnection.ts
@@ -232,13 +232,13 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
 
     /**
      * @brief Adds a new track to the {@link RTCPeerConnection},
-     * and indicates that it is contained in the specified {@link MediaStream} array.
-     * This method has to be synchronous as the W3C API expects a track to be return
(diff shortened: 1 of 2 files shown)


8BallBomBom (approved):
Simple fix, looks good 👍🏻 
Beat me to it.

saghul (approved):

PR #1201, opened 2022-10-19
ios: ensure property not null to prevent crash

hans00:
Do `addIceCandidate` without `sdpMid` will crash on [this line](https://webrtc.googlesource.com/src/+/refs/heads/main/sdk/objc/api/peerconnection/RTCIceCandidate.mm#65)

In [spec](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/sdpMid) it must not null

### DIFF
--- a/src/RTCIceCandidate.ts
+++ b/src/RTCIceCandidate.ts
@@ -1,9 +1,13 @@
 export default class RTCIceCandidate {
     candidate: string;
-    sdpMLineIndex: number | null;
-    sdpMid: string | null;
+    sdpMLineIndex: number;
+    sdpMid: string;
 
     constructor({ candidate = '', sdpMLineIndex = null, sdpMid = null }) {
+        if (sdpMLineIndex === null || sdpMid === null) {
+            throw new TypeError('`sdpMLineIndex` and `sdpMid` must not null');
+        }
+
         this.candidate = candidate;
         this.sdpMLineIndex = sdpMLineIndex;
         this.sdpMid = sdpMid;
--- a/src/RTCIceCandidate.ts
+++ b/src/RTCIceCandidate.ts
@@ -1,22 +1,12 @@
 export default class RTCIceCandidate {
     candidate: string;
-    sdpMLineIndex: number;
-    sdpMid: string;
+    sdpMLineIndex: number | null;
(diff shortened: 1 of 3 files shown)


saghul:
Can you please add the check in the JS side instead, so it applies to Android as well?

saghul:
Thanks for updating the PR! Those checks are too harsh though. Look at how the `RTCIceCandidate` class should handle parameters: https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/RTCIceCandidate

Also see addIceCandidate: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addIceCandidate

> The method will throw a TypeError exception if both sdpMid and sdpMLineIndex are null.

src/RTCIceCandidate.ts
```diff
@@ -1,22 +1,12 @@
...
-            throw new TypeError('`sdpMLineIndex` must be number');
-        }
-        if (typeof info?.sdpMid !== 'string') {
-            throw new TypeError('`sdpMid` must be string');
-        }
-
-        this.candidate = info.candidate;
-        this.sdpMLineIndex = info.sdpMLineIndex;
-        this.sdpMid = info.sdpMid;
+    constructor({ candidate = '', sdpMLineIndex = null, sdpMid = null }) {
```
saghul, line 6: Can you make this throw in this case, as per the spec? https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/RTCIceCandidate#exceptions

saghul:
One more comment and we're good, nice work!

src/RTCIceCandidate.ts
hans00: added

omerts:
@hans00 @saghul Null is a valid value for `sdpMid` & `sdpMLineIndex`.

See: https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate#instance_properties

image

saghul:
Note that regardless of that the spec says we are bound by what the native WebRTC library accepts. As you can see in the original post, it didn't support that.

It's a year later now, so that might have changed.

PR #1207, opened 2022-10-20
Fix remove track

saghul:


src/RTCPeerConnection.ts
```diff
@@ -552,14 +546,13 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+            // As per the spec, all we need to do is mark the track as muted:
+            // https://w3c.github.io/webrtc-pc/#process-remote-track-removal
             for (const stream of this._remoteStreams.values()) {
-                const trackIdx = stream._tracks.findIndex(t => t.id === ev.trackId);
+                const [ track ] = stream._tracks.filter(t => t.id === ev.trackId);
 
-                if (trackIdx !== -1) {
-                    stream._tracks.splice(trackIdx, 1);
+                if (track) {
+                    track._muted = true;
```
tmoldovan8x8, line 555: Is the track collected anytime?

saghul: According to the spec (linked in the commit) the track should remain there, but be marked as muted.

src/RTCPeerConnection.ts
```diff
@@ -546,14 +546,13 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
 
             log.debug(`${this._pcId} onremovetrack`);
 
-            // Based on the W3C specs https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-removetrack,
-            // we need to remove the track from any media streams
-            // that were previously passed to the `track` event.
+            // As per the spec, all we need to do is mark the track as muted:
+            // https://w3c.github.io/webrtc-pc/#process-remote-track-removal
             for (const stream of this._remoteStreams.values()) {
```
Yousif-CS, line 551: _"When the other peer stops sending a track in this manner, the track is removed from any remote [MediaStream](https://www.w3.org/TR/mediacapture-streams/#dom-mediastream)s that were initially revealed in the [track](https://w3c.github.io/webrtc-pc/#event-track) event, and if the [MediaStreamTrack](https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack) is not already muted, a mute event is fired  at the track."_

I feel this means the track is removed from remote streams as well?

saghul: Hum, interesting. I don't see that mentioned on the other side (processing of the remote SDP) but you're right.

One limitation we do have is that `ontrack` won't fire again if you add the track to more streams later on.

saghul:
Updated, PTAL @Yousif-CS

Yousif-CS (approved):
LGTM

src/RTCPeerConnection.ts
murillo128: The process for updating the remote track/stream association is described here:

https://w3c.github.io/webrtc-pc/#set-associated-remote-streams

I think that the comment about `removetrack` is misleading and we should point to this section instead.

Also note that the association algorithm is not done in chrome based on the `OnRemoveTrack` event, but on the 

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/peerconnection/rtc_peer_connection.cc;l=2318;drc=327a3573ed3d1ee29d433ee215e50947f1b94170;bpv=0;bpt=1

which is called ones an SDP processing (either local/remote) is finished:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/peerconnection/rtc_peer_connection_handler.cc;l=700;drc=327a3573ed3d1ee29d433ee215e50947f1b94170;bpv=1;bpt=1

saghul: Thanks for taking a look Sergio! ❤️

PR #1208, opened 2022-10-20
onAddTrack: Do not skip track if track.id() exists

cpoile:
#### Summary
This PR fixes a problem we've noticed when  `peerconnection.onAddTrack` is fired. Mainly, the library skips adding the track if the `track.id()` already exists in the `remoteTracks` map.

This is causing us a problem because a peer can send a video track with an id (e.g., `screen`), and then stop sending the track, then start sending the track again (with the same id, e.g. `screen`). 

The code has this helpful comment (thank you for that, btw):

```
 // Unlike in the WebRTC spec, the libwebrtc native implementation
 // fires onTrack on every sRD which has an active receiving transceiver.
 // So, if we are already keeping track of this transceiver, ignore the event.
```

We were using a fork of `react-native-webrtc` (https://github.com/openland/react-native-webrtc) which did not do this check, and it was working well for us: e.g. https://github.com/openland/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java#L510-L514

The spec is a little unclear (at least to me) about what should happen when resending a track that was closed earlier. But it seems that at the very least, `track.id()`s are not guaranteed to be unique (https://blog.mozilla.org/webrtc/rtcrtptransceiver-explored/), and "[if you're sending track ids out-of-band today hoping to correlate local and remote tracks, WebRTC 1.0 will break you](https://blog.mozilla.org/webrtc/the-evolution-of-webrtc/)", and the track ids are not even guaranteed to be the same as the id sent from the peer [point 2 from the W3C draft](https://w3c.github.io/webrtc-pc/#rtcrtpreceiver-interface). That leads me to think we shouldn't rely on them for indexing tracks over time.

So we propose to not use the ids to check if the track already exists. Instead, add tracks when `peerconnection.onAddTrack` is called, and then let the application logic decide on how to handle the `onTrack` event.

#### Alternatives:
- We could index remoteTracks using a synthentic id, either a UUID or `track.Id() + transciever.getMid()` which should be unique even when the same `track.Id()`s are used. But then we would have to maintain a mapping between (the non-unique) `track.Id()` -> (the unique) UUID, and determine which unique track to return when `getTrack(String trackId)` is called.
- Some other solution that you can see that I can't?

As an aside: In an ideal world, there would be a clear way to stop sending a video track and we could clean up the `remoteTracks` map, but there doesn't seem to be a consensus (https://blog.mozilla.org/webrtc/the-evolution-of-webrtc/). E.g., we tried removing the track explicitly and renegotiating, then adding the track again and renegotiating, but that only seems to work on iOS.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -367,13 +367,6 @@ class PeerConnectionObserver implements PeerConnection.Observer {
 
             final MediaStreamTrack track = receiver.track();
 
-            if (remoteTracks.containsKey(track.id())) {
-                // Unlike in the WebRTC spec, the libwebrtc native implementation
-                // fires onTrack on every sRD which has an active receiving transceiver.
-                // So, if we are already keeping track of this transceiver, ignore the event.
-                return;
-            }
-
             if (track.kind().equals(MediaStreamTrack.VIDEO_TRACK_KIND)){
                 videoTrackAdapters.addAdapter((VideoTrack) track);
             }
--- a/ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
@@ -801,11 +801,7 @@ RCT_EXPORT_METHOD(peerConnectionRemoveTrack:(nonnull NSNumber *)objectID
         }
         
         RTCMediaStreamTrack *track = rtpReceiver.track;
-        
-        if (peerConnection.remoteTracks[track.trackId]) {
-            return;
-        }
-        
+
         if (track.kind == kRTCMediaStreamTrackKindVideo) {
             RTCVideoTrack *videoTrack = (RTCVideoTrack *)track;
             [peerConnection addVideoTrackAdapter: videoTrack];


saghul:
Hey there! Thanks for the feedback! Some comments.



> This is causing us a problem because a peer can send a video track with an id (e.g., `screen`), and then stop sending the track, then start sending the track again (with the same id, e.g. `screen`).

That sounds more like a label, not an id. The id is defined as:

```
 [[Id]], initialized to a newly generated unique identifier string. See [MediaStream](https://w3c.github.io/mediacapture-main/#dom-mediastream).[id](https://w3c.github.io/mediacapture-main/#dom-mediastream-id) attribute for guidelines on how to generate such an identifier.
 [id](https://w3c.github.io/mediacapture-main/#dom-mediastream-id) of type [DOMString](https://webidl.spec.whatwg.org/#idl-DOMString), readonly
```
```
    The id attribute MUST return the value to which it was initialized when the object was created.

    When a [MediaStream](https://w3c.github.io/mediacapture-main/#dom-mediastream) is created, the User Agent MUST generate an identifier string, and MUST initialize the object's [id](https://w3c.github.io/mediacapture-main/#dom-mediastream-id) attribute to that string, unless the object is created as part of a special purpose algorithm that specifies how the stream id must be initialized. A good practice is to use a UUID [[rfc4122](https://w3c.github.io/mediacapture-main/#bib-rfc4122)], which is 36 characters long in its canonical form. To avoid fingerprinting, implementations SHOULD use the forms in section 4.4 or 4.5 of RFC 4122 when generating UUIDs.

    An example of an algorithm that specifies how the stream id must be initialized is the algorithm to associate an incoming network component with a [MediaStream](https://w3c.github.io/mediacapture-main/#dom-mediastream) object. [[WEBRTC](https://w3c.github.io/mediacapture-main/#bib-webrtc)]
```

See: https://w3c.github.io/mediacapture-main/#dfn-id

> The spec is a little unclear (at least to me) about what should happen when resending a track that was closed earlier. But it seems that at the very least, `track.id()`s are not guaranteed to be unique (https://blog.mozilla.org/webrtc/rtcrtptransceiver-explored/), and "[if you're sending track ids out-of-band today hoping to correlate local and remote tracks, WebRTC 1.0 will break you](https://blog.mozilla.org/webrtc/the-evolution-of-webrtc/)", and the track ids are not even guaranteed to be the same as the id sent from the peer [point 2 from the W3C draft](https://w3c.github.io/webrtc-pc/#rtcrtpreceiver-interface). That leads me to think we shouldn't rely on them for indexing tracks over time.

On the contrary. The spec above makes it pretty clear that the ID is generated by the user agent, and not the user, and that in fact does not correlate, which is fine, we don't correlate them.

Look at how the receiver is built, which is what may trigger `ontrack`:

```
Let track be a new MediaStreamTrack object. The source of track is a remote source provided by receiver. Note that the track.id is generated by the [user agent](https://www.w3.org/TR/webrtc/#dfn-user-agent) and does not map to any track IDs on the remote side.
```

See: https://www.w3.org/TR/webrtc/#rtcrtpreceiver-interface

> So we propose to not use the ids to check if the track already exists. Instead, add tracks when `peerconnection.onAddTrack` is called, and then let the application logic decide on how to handle the `onTrack` event.

I think your analysis of the situation is incorrect, but I could be wrong, because indeed the logic is quite twisted: https://www.w3.org/TR/webrtc/#process-remote-tracks

So I think both cases would be possible: our handling of IDs looks correct to me, but perhaps the `ontrack` event should fire more often.

I think the best way forward would be to see what browsers do with a simple JS fiddle or something similar. Have a couple of transceivers between 2 pcs, then stop sending on one and then resume, and check if the event is fired.

> * We could index remoteTracks using a synthentic id, either a UUID or `track.Id() + transciever.getMid()` which should be unique even when the same `track.Id()`s are used. But then we would have to maintain a mapping between (the non-unique) `track.Id()` -> (the unique) UUID, and determine which unique track to return when `getTrack(String trackId)` is called.

They are already UUIDs, and AFAICT don't change.

> As an aside: In an ideal world, there would be a clear way to stop sending a video track and we could clean up the `remoteTracks` map, but there doesn't seem to be a consensus (https://blog.mozilla.org/webrtc/the-evolution-of-webrtc/). E.g., we tried removing the track explicitly and renegotiating, then adding the track again and renegotiating, but that only seems to work on iOS.

The track should remain there but muted, as per the track removal procedure: https://www.w3.org/TR/webrtc/#process-remote-track-removal (this was wrong and I'm fixing it in https://github.com/react-native-webrtc/react-native-webrtc/pull/1207

cpoile:
Thanks for the comments @saghul, I appreciate all the work you and the others are doing on this library--we depend on it quite a bit, so thank you!

> That sounds more like a label, not an id. The id is defined as: [snip...]
> See: https://w3c.github.io/mediacapture-main/#dfn-id

You're right, this makes sense from the mediacapture/useragent pov -- you can be sure they'll generate UUIDs.  But from what I can tell  https://www.w3.org/TR/webrtc/#rtcrtpreceiver-interface point 2 states: 
```text
Let track be a new [MediaStreamTrack](https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack) 
object [[GETUSERMEDIA](https://www.w3.org/TR/webrtc/#bib-getusermedia)]. The source of track is a remote 
source provided by receiver. Note that the track.id is generated by the 
[user agent](https://www.w3.org/TR/webrtc/#dfn-user-agent) and does not map to any track IDs on the
remote side.
```

And maybe that's what's happening here?  libwebrtc is giving us back the remote's `track.id`, instead of generating a new UUID itself. 

But even if that's not the problem, then we're still left with the consensus that track.id is basically ignored now: e.g., the disagreements (which I don't completely understand, tbh -- still learning this): https://github.com/w3c/webrtc-pc/issues/1562 , https://github.com/rtcweb-wg/jsep/issues/842 , and https://github.com/w3c/webrtc-pc/issues/1718 . This is a good summary: https://github.com/w3c/webrtc-pc/issues/2005#issuecomment-436930893

> I think the best way forward would be to see what browsers do with a simple JS fiddle or something similar. Have a couple of transceivers between 2 pcs, then stop sending on one and then resume, and check if the event is fired.

Yep, here's a fiddle showing `ontrack` firing with the same track id: https://jsfiddle.net/wxk9h5ab/3/
It's not exactly the same situation that led to this PR, but I think the result is the same -- I think we should follow suit and  fire the `ontrack` event, even if we get a `track.id` we've seen before.

> The track should remain there but muted, as per the track removal procedure: https://www.w3.org/TR/webrtc/#process-remote-track-removal (this was wrong and I'm fixing it in https://github.com/react-native-webrtc/react-native-webrtc/pull/1207

Thank you!

saghul:
I did some investigation in Chrome and indeed, `ontrack` fires again if a transceiver switches from `sendrecv` to `recvonly` to `sendrecv` again.

Now, an important thing here is that the track instance is the exact same, they are essentially the same JS object. So the code in `peerConnectionOnTrack` would need to be adjusted in order to not create a new JS instance for a track whuich represents the same native track.

saghul:
I think that's a separate problem (or maybe not even a problem at all).

Chrome fires ontrack for ththe exact same track instance as before.

So keeping our adapter indexed by ID is ok, for now at least.

cpoile:
Ok, I'll defer to you as the expert here. What should the next step be? 

Imo, I think at least we should be firing an `ontrack` even if the `track.id` has been seen before.

saghul:
(Untested), here you go: https://github.com/react-native-webrtc/react-native-webrtc/pull/1212

PR #1212, opened 2022-10-25
pc: fix ontrack not firing multiple times in certain situations

saghul:
The ontrack event should fire for the same track for example when the sender switches the direction from sendrecv to recvonly and then back.

This would cause the recipient's direction to go from sendrecv to sendonly and back. The ontrack event is sent when the direction changes fron NOT *recv* to *recv*.

Closes: https://github.com/react-native-webrtc/react-native-webrtc/pull/1208

### DIFF
--- a/src/RTCPeerConnection.ts
+++ b/src/RTCPeerConnection.ts
@@ -484,7 +484,7 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
 
             log.debug(`${this._pcId} ontrack`);
 
-            const track = new MediaStreamTrack(ev.receiver.track);
+            let track;
             let transceiver;
 
             // Make sure transceivers are stored in timestamp order. Also, we have to make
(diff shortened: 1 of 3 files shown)


src/RTCPeerConnection.ts
```diff
@@ -484,27 +484,31 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+            // when the transceiver direction (on the sending side) switches from
+            // sendrecv to recvonly and then back.
 
-                transceiver = new RTCRtpTransceiver({ ...ev.transceiver, receiver, sender });
-                this._insertTransceiverSorted(ev.transceiverOrder, transceiver);
-            } else {
+            if (oldTransceiver) {
                 transceiver = oldTransceiver;
-                transceiver._receiver._track = track;
+                track = transceiver._receiver._track;
```
saghul, line 501: Note to self: mark the track as unmuted at this point.

Yousif-CS (approved):
LGTM

cpoile:
Thank you @saghul!

PR #1220, opened 2022-11-09
iOS: Fix MediaStreamTrack readyState always ended

### LINKED ISSUE #1219: iOS: received track from PeerConnection onTrack event is always readyState "ended"
#### Expected Behavior
Received active tracks should be in the "initializing" or "live" state.

#### Observed Behavior
Received track is always "ended". Notably, when testing, the audio for an audio track is actually still being received and can be heard through the speakers, so the underlying track at the native layer is still live.

#### Steps to reproduce the issue
1. Negotiate and receive a track from remote on an iOS device.
2. Examine the `readyState` property of the MediaStreamTrack in js.

#### Platform Information

```
    "react": "17.0.2",
    "react-native": "0.67.5",
    "react-native-webrtc": "106.0.0-beta.5"
```
* **OS**: iOS 15.6

davidliu:
Fixes #1219

### DIFF
--- a/ios/RCTWebRTC/SerializeUtils.m
+++ b/ios/RCTWebRTC/SerializeUtils.m
@@ -160,10 +160,10 @@
     NSString *readyState;
     switch (track.readyState) {
         case RTCMediaStreamTrackStateLive:
-            readyState = @"Live";
+            readyState = @"live";
             break;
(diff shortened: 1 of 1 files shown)


saghul:
Good catch! Can you lowercase the constants while you're at it?

saghul (approved):

PR #1239, opened 2022-11-26
RTCRtpReceiver/RTCRtpSender getStats implementation

davidliu:
Looks like native doesn't have direct access to the getStats filtered for receiver/sender, so this is an approximation for `RTCRtpReceiver`/`RTCRtpSender`. Matched against what I saw with chrome web behavior. 

Seems to include these stats (rtpId = receiver/sender id):
* Track with `trackIdentifier === rtpId`
* All `inbound-rtp` (receiver)/`outbound-rtp` (sender) with `trackId === Track.id`
* Any `codec` specified by the inbound/outbound-rtp `codecId` member.
* Any `remote-outbound-rtp` (receiver)/`remote-inbound-rtp` (sender) specified by the inbound/outbound-rtp `ssrc`
* The `media-source` stats with `trackIdentifier === rtpId` (only exists for sender)
* All `certificate` stats
* The `transport` stat
* The `candidate-pair` with `nominated === true`
* The `local-candidate` specified by candidate-pair's `localCandidateId`.
* The `remote-candidate` specified by candidate-pair's `remoteCandidateId`.

I actually didn't see `stream` being returned in any stats on web, whether by PeerConnection or the rtp objects, but I included them in the stats returned here.

----

This could technically be done in JS I think, but there's a warning about performance here, so kept this one native. If this looks acceptable, I'll look into getting an iOS implementation done soon as well.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -19,6 +19,8 @@ import org.webrtc.IceCandidate;
 import org.webrtc.MediaStream;
 import org.webrtc.MediaStreamTrack;
 import org.webrtc.PeerConnection;
+import org.webrtc.RTCStats;
+import org.webrtc.RTCStatsReport;
 import org.webrtc.RtpReceiver;
 import org.webrtc.RtpSender;
 import org.webrtc.RtpTransceiver;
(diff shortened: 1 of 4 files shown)


davidliu:
Looks like iOS has access to the getStats at the native level, so no approximation code needed there.

saghul:
Ohai! Thank you for this! Are you aware that the Android native counterpart landed not that long ago? https://webrtc.googlesource.com/src/+/4dc6e05ac96b25ea38b5bad7d76c44bd550b0dba

I was eyeing doing this, so thanks for the help! What do you think about waiting for the release, and landing this with the right Android implementation, once we bump the WebRTC dependency?

davidliu:
Ah, didn't see that they added that. What's the timeline for WebRTC dependency bumps? Looks like that'd be coming out with m109.

Currently using `getStats` for some internal testing bits so it would be really nice to have sooner than later. Can we work with this filtering method and then replace it with the official version later so that we don't have to wait?

saghul:
Sounds good. I'll give it a review now.

saghul (approved):

saghul:
Cheers!

Issue #805, opened 2020-06-12
Feature request: Custom video source (video generated by code)?


fzyzcjy:
#### Expected behavior
Hi thanks for the wonderful library! I wonder how can I use a custom video source? For simplicity, just consider the custom source as one generated by some code.

Thanks very much!

#### Observerd behavior
Do not have this feature.

#### Steps to reproduce the problem
N/A

#### Platform information
N/A

* **React Native version**: latest
* **Plugin version**: latest
* **OS**:  both
* **OS version**: latest

danjenkins:
I'm not sure what you mean @fzyzcjy ? You want to be able to create a media stream (somehow) and pass it into the lib? Thats what already happens with getting the video / audio streams.....Although I'm not sure if the stream is required to be made using this native module due to the react native bridge. I'll add a label of feature request for now and wait for others to pipe in with the practical info

fzyzcjy:
Yes I want to do that. Thanks!

saghul:
You'd need to implement this in native in the form of a custom capturer.

Desintegrator:
> Yes I want to do that. Thanks!
@fzyzcjy 
Hi! Were you able to implement custom capturer?

fkrauthan:
Is there any documentation about implementing a custom capturer? I am currently looking if it is possible to combine VisonCamera as my camera source (https://mrousavy.com/react-native-vision-camera/docs/guides/frame-processors) with react-native-webrtc to transmit the video stream.

saghul:
Screen sharing is implemented as a custom capturer, maybe that can serve as guidance.

fkrauthan:
But when I was digging into that it seems like screen sharing (at least on Android) is just using the webrtc android library that provides the actual data stream?

fkrauthan:
Ah ok. But it also does seem to be titled integrated into the UserMedia code and there seem to be no easy way to inject an outside video source implementation into the existing system. So I still would have to fork this library and build my own custom version or am I missing something?

saghul:
We don't prove an API for that at the moment.

I just mentioned how it could be accomplished.

fkrauthan:
Ah ok that makes sense. Can we maybe reopen this feature request? Given that there is currently no API it would be great if an API for this use-case could be added to the library.

saghul:
PRs are welcome. I don't want to leave it open giving people the expectation it's going to happen, because nobody is currently working on it.

[...]

PR #1243, opened 2022-12-02
Add ability to create local custom tracks

davidliu:
Addresses #805. Exposes the needed methods to create a video track with a custom capturer.

Note: I was able to get Android working fine. iOS streamed video to remote subscribers fine, but when displaying that same video locally, it only displayed a solid color; not sure where the source of that issue lies.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
+++ b/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
@@ -307,7 +307,7 @@ class GetUserMediaImpl {
         displayMediaPromise = null;
     }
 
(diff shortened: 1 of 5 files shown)


davidliu:
Looks like there's a bug with RTCMTLVideoView, it can't render RGBA CVPixelBuffers (which is what I was using to output frames for testing; remote subscribers always receive I420 which is why they render fine). If I convert the local frames to I420, they render fine. This seems like an issue that can be addressed separately. I've also filed a bug on chromium: https://bugs.chromium.org/p/webrtc/issues/detail?id=14741

saghul:
Thanks for this!

I'm a bit busy this end of year, hoping to get Unified Plan as stable soon, I'll find some time to get to this afterwards.

saghul (approved):

PR #1250, opened 2022-12-11
Bringing iOS transceiver init handling closer to Android

8BallBomBom:
Need to do some testing to ensure everything is working as expected.

### DIFF
--- a/ios/RCTWebRTC/SerializeUtils.m
+++ b/ios/RCTWebRTC/SerializeUtils.m
@@ -234,7 +234,9 @@
     RTCRtpTransceiverInit *transceiverInit = [RTCRtpTransceiverInit new];
     
     NSString *direction = [params objectForKey: @"direction"];
-    [transceiverInit setDirection: [SerializeUtils parseDirection: direction]];
+    if (direction) {
+        [transceiverInit setDirection: [SerializeUtils parseDirection: direction]];
+    }
     
     NSArray *streamIds = [params objectForKey: @"streamIds"];
     if (streamIds) {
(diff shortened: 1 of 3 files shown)


8BallBomBom:
@saghul Seems to be working as expected 👍🏻 
But might need a fix up.

ios/RCTWebRTC/SerializeUtils.m
```diff
@@ -193,21 +193,65 @@ + (NSString *) serializeDirection: (RTCRtpTransceiverDirection) direction {
...
+      [encoding setScaleResolutionDownBy: (NSNumber *) params[@"scaleResolutionDownBy"]];
+    }
+
+    return encoding;
+}
+
++ (RTCRtpTransceiverInit *) parseTransceiverOptions: (NSDictionary *) params {
+    RTCRtpTransceiverInit *transceiverInit = [RTCRtpTransceiverInit new];
+    
+    NSString *direction = [params objectForKey: @"direction"];
```
saghul, line 236: Here we should check if the direction is not nil before proceeding, like we do on Android: https://github.com/react-native-webrtc/react-native-webrtc/blob/f2278cf36ddaff43b505f19cad558240a4997aab/android/src/main/java/com/oney/WebRTCModule/SerializeUtils.java#L292

saghul (changes requested):
One nit, LGTM otherwise, good work!

ios/RCTWebRTC/SerializeUtils.m
8BallBomBom: Just figured as the direction is set as Inactive by default, the same functionality occurs without checking if the direction is nil.

Should i still proceed?

8BallBomBom: Ahhh wait alright, yes my thinking of that was backwards, will address that now 👍🏻

saghul (approved):

saghul:
:rocket:

Issue #1257, opened 2022-12-21
Crash when releasing the local stream on android


qusaieilouti99:
When releasing the local stream the app crashed on android and this is how we release the stream:
```
 releaseLocalStream = () => {
        if (this.localStream) {
            this.localStream.release(true);
            this.localStream = undefined;
        }
    }
```

the crash report:
```
FATAL EXCEPTION: pool-17-thread-1

java.util.ConcurrentModificationException

FATAL EXCEPTION: pool-17-thread-1
Process: com.pingme2022, PID: 20168
java.util.ConcurrentModificationException
	at java.util.IdentityHashMap$IdentityHashMapIterator.nextIndex(IdentityHashMap.java:734)
	at java.util.IdentityHashMap$ValueIterator.next(IdentityHashMap.java:832)
	at org.webrtc.VideoTrack.dispose(VideoTrack.java:59)
	at com.oney.WebRTCModule.GetUserMediaImpl$TrackPrivate.dispose(GetUserMediaImpl.java:496)
	at com.oney.WebRTCModule.GetUserMediaImpl.disposeTrack(GetUserMediaImpl.java:241)
	at com.oney.WebRTCModule.WebRTCModule.lambda$mediaStreamTrackRelease$16$com-oney-WebRTCModule-WebRTCModule(WebRTCModule.java:821)
	at com.oney.WebRTCModule.WebRTCModule$$ExternalSyntheticLambda16.run(Unknown Source:4)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.lang.Thread.run(Thread.java:923)
```

Possible solution:
[Link](https://www.baeldung.com/java-concurrentmodificationexception) 

The line causing the issue is 59 in this file: [org.webrtc](https://chromium.googlesource.com/external/webrtc/+/master/sdk/android/api/org/webrtc/VideoTrack.java)

The issue happened to me only once while doing heavy testing on the video calls.

#### Expected Behavior
it shouldn't crash

#### Observed Behavior
it crashed

#### Steps to reproduce the issue
N/A

#### Platform Information

* **React Native Version**: 0.70.6
* **Plugin Version**: 106.0.1
* **OS**: Android 
* **OS Version**:  11

saghul:
Can you consistently reproduce this?

[...]

PR #1258, opened 2022-12-22
android: fix race condition when disposing tracks

saghul:
Make sure add/removeSink operations run in the WebRTC thread, so it's safe to dispose and not get an error when concurrently modifying the sinks container, which happens inside WebRTC.

Fixes: https://github.com/react-native-webrtc/react-native-webrtc/issues/1257


tmoldovan8x8 (approved):

8BallBomBom:
Testing all good 👍🏻

saghul:
Thank you!

PR #1262, opened 2022-12-27
fix newSdp = null in PeerConnectionObserver.java

qusaieilouti99:
and comment the logger

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -396,8 +396,11 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         params.putMap("candidate", candidateParams);
         SessionDescription newSdp = peerConnection.getLocalDescription();
         WritableMap newSdpMap = Arguments.createMap();
-        newSdpMap.putString("type", newSdp.type.canonicalForm());
-        newSdpMap.putString("sdp", newSdp.description);
+        // Can happen when doing a rollback.
+        if (newSdp != null) {
+            newSdpMap.putString("type", newSdp.type.canonicalForm());
+            newSdpMap.putString("sdp", newSdp.description);
+        }
         params.putMap("sdp", newSdpMap);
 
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -15,7 +15,7 @@ import RTCSessionDescription from './RTCSessionDescription';
 import RTCView from './RTCView';
 import ScreenCapturePickerView from './ScreenCapturePickerView';
 
-Logger.enable('*');
+// Logger.enable('*');
```
saghul, line 18: Please leave this the way it was.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -396,8 +396,11 @@ public void onIceCandidate(final IceCandidate candidate) {
         params.putMap("candidate", candidateParams);
         SessionDescription newSdp = peerConnection.getLocalDescription();
         WritableMap newSdpMap = Arguments.createMap();
-        newSdpMap.putString("type", newSdp.type.canonicalForm());
-        newSdpMap.putString("sdp", newSdp.description);
+        // Can happen when doing a rollback.
+        if (newSdp != null) {
```
saghul, line 400: Don't we need to handle this in the JS layer?

saghul (commented):
Can you please apply the same fix on iOS too? I left a couple of comments too, thank you!

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
qusaieilouti99: It's already handled this way on peerConnectionSetLocalDescription on setSuccess in the native layer.
so I just copied that

qusaieilouti99: I'm not familiar with objective c, so if you're back can you please fix it on IOS.

saghul: Ok

PR #1268, opened 2023-01-01
Fix new sdp, prevent crashing if rollback causes local description to be undefined

8BallBomBom:
Pulled in #1262 and added in the iOS parts.
Seems to be building, currently untested.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -396,8 +396,11 @@ class PeerConnectionObserver implements PeerConnection.Observer {
         params.putMap("candidate", candidateParams);
         SessionDescription newSdp = peerConnection.getLocalDescription();
         WritableMap newSdpMap = Arguments.createMap();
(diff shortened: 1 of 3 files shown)


saghul:
Thanks for picking this up! We also need to set the description to null in RTCPeerConnection.ts in the ICE candidate handler event.

8BallBomBom:
Ahhhhhhh right, will include that now.

So both the `peerConnectionGotICECandidate` and `peerConnectionIceGatheringChanged` events need to check if the received sdp is empty and then set the descriptions as null?

Question is, couldn't we just send that as null from native rather than an empty array?
Should we handle it like that instead?

8BallBomBom:
Didn't need the extra variable but... cosmetics 😅 
Needs testing 👍🏻

saghul:
The problem is that the serialization uses JSON, which doesn't allow for null.

One way would be to not send the new SDP object at all, but I think the native code looks better this way...

8BallBomBom:
Ahhhh right yeah didn't think about that and makes sense 😄 
PR is good to go if tested and working 👍🏻

saghul (approved):
LGTM, nice work!

Issue #1251, opened 2022-12-12
Missing events for MediaStreamTrack


davidliu:
I'm unable to receive the events for `MediaStream` and `MediaStreamTrack` objects.

#### Expected Behavior
An event listener set like so:

````
mediastream.addEventListener('addtrack', (event) => { });

mediastream.onaddtrack = (event) => { };
````

Should fire when the appropriate event happens.

#### Observed Behavior
Events do not fire. Looks like they're still unimplemented at the moment for `MediaStream` and `MediaStreamTrack` objects.

#### Platform Information

* **React Native Version**: 0.69.5
* **Plugin Version**: 106.0.0-beta8
* **OS**: Android
* **OS Version**: 12

davidliu:
@saghul There's still the missing `MediaStreamTrack` events, though couldn't find an easy way to trigger those events.

saghul:
Only `ended` is missing AFAICT, right?

davidliu:
Ah, I see the muted events now. Yeah, the `ended` event is missing.

fukemy:
I am have same problem too and need to get ended event

saghul:
Can you find in the spec when it should fire?

davidliu:
https://www.w3.org/TR/mediacapture-streams/#event-mediastreamtrack-ended

https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/ended_event

I have some work in progress on this; the Android side is simple enough since it exposes when the capturer is stopped, but iOS strangely doesn't have any listeners for this. Can easily handle cases where the user stops the capturer, but cases where the capturer stops for another reason (i.e. system stops it) are a little up in the air.

saghul:
Sounds good, I'll be happy to review the PR when ready!

saghul:
Fixed in https://github.com/react-native-webrtc/react-native-webrtc/pull/1269

Note that we don't intend to support the ended event unless for screen-sharing streams when the use stops the broadcast from the system UI.

PR #1269, opened 2023-01-04
Local MediaStreamTrack ended event

davidliu:
Addresses #1251.

Little unsure on what the spec specifically is, since at least [one browser seems to not fire the `ended` event upon user-initiated stop](https://stackoverflow.com/questions/55953038/why-is-the-ended-event-not-firing-for-this-mediastreamtrack); this implementation fires `ended` whenever the capturer is stopped, regardless of reason.


8BallBomBom:
Testing on Android seems good, will test on iOS soon.
As for the spec, unsure but will do some research 👍🏻

android/src/main/java/com/oney/WebRTCModule/ScreenCaptureController.java
```diff
@@ -57,6 +57,7 @@ protected VideoCapturer createVideoCapturer() {
                 public void onStop() {
                     Log.w(TAG, "Media projection stopped.");
                     orientatationListener.disable();
+                    stopCapture();
```
saghul, line 60: Is this change necessary because of this PR or was it necessary before?

android/src/main/java/com/oney/WebRTCModule/CameraEventsHandler.java
```diff
@@ -13,7 +13,7 @@ class CameraEventsHandler implements CameraVideoCapturer.CameraEventsHandler {
     // Callback invoked when camera closed.
     @Override
     public void onCameraClosed() {
-        Log.d(TAG, "CameraEventsHandler.onFirstFrameAvailable");
+        Log.d(TAG, "CameraEventsHandler.onCameraClosed");
```
saghul, line 16: Thanks for taking care of this!

android/src/main/java/com/oney/WebRTCModule/TrackCapturerObserver.java
```diff
@@ -0,0 +1,52 @@
...
+  /** Notify that the capturer has been stopped. */
+  public void onCapturerStopped() {
+    this.capturerObserver.onCapturerStopped();
+
+    WritableMap params = Arguments.createMap();
+    params.putString("trackId", trackId);
+
+    Log.d(TAG, "ended event trackId: " + trackId);
+
+    webRTCModule.sendEvent(
```
saghul, line 44: Technically a track ID can be duplicated across different peer connections. That's why we always have the PC ID to go with track events.

saghul:
I'm a little torn on this one. According to Jan-Ivar, who works at Mozilla, it should not fire if the user stopped the track themselves: https://stackoverflow.com/a/55960232

That only leaves the abrupt terminations on the table, or more clearly stopping a screen capture with the system stop button.

android/src/main/java/com/oney/WebRTCModule/ScreenCaptureController.java
davidliu: Internally, `ScreenCapturerAndroid` doesn't listen to the media projection `onStop` callback, so the capturer stays "running" until directly stopped. The PR itself relies on stopping the capture to know when to emit the ended event, but aside from that, seems like a good idea to have it stop alongside the media projection stop to release any resources immediately.

android/src/main/java/com/oney/WebRTCModule/TrackCapturerObserver.java
davidliu: Hmm, this one is a little hairy since tracks are created independently from their eventual peer connection. I was figuring it'd be safe enough since we use UUIDs for the tracks. This'll need a little more thinking to insert the PC id into here when they get associated to each other.

saghul: I'll double check, since we have changed things around indeed.

android/src/main/java/com/oney/WebRTCModule/ScreenCaptureController.java
saghul: Can you pl send a PR with this change alone?

davidliu:
@saghul Hmm, if that's the case, I think I can just add a flag to the capturers so as not emit the event when the sdk/user calls stop manually, does that sound reasonable?

saghul:
I'd like to thread lightly here. My understanding is a track cannot be resumed once it gets to "ended".

davidliu:
Hmm, is the concern that the user can somehow resume the capturer even after it's stopped? Not quite understanding the issue here.

saghul:
> Hmm, is the concern that the user can somehow resume the capturer even after it's stopped? Not quite understanding the issue here.

Maybe I misunderstood what you meant, but if we emit ended when the capturer is interrupted, what happens when it resumes?

davidliu:
AFAICT, the only capturer at the moment that resumes itself is iOS's camera capturer (this one still needs to be addressed, or left as a TODO). The other ones don't have any sort of resume functionality and stop indefinitely until `startCapture` is called again.

Normally this would fine since we the web API of MediaStreamTrack doesn't have start function, but did find a snag here; `MediaStreamTrack.setEnabled` additionally starts and stops the capturer, so users could resume a capturer that way.

Checked on Chrome and Safari, they seem to keep the capturer going while the track is disabled, while Firefox stops the capturer while the track is disabled. If we went the way of Chrome, that could solve this issue though might be a big unexpected change for users of the library.

saghul:
> AFAICT, the only capturer at the moment that resumes itself is iOS's camera capturer (this one still needs to be addressed, or left as a TODO). The other ones don't have any sort of resume functionality and stop indefinitely until `startCapture` is called again.

Right.



> Normally this would fine since we the web API of MediaStreamTrack doesn't have start function, but did find a snag here; `MediaStreamTrack.setEnabled` additionally starts and stops the capturer, so users could resume a capturer that way.

That's an internal API and it's not exposed so that's ok.



> Checked on Chrome and Safari, they seem to keep the capturer going while the track is disabled, while Firefox stops the capturer while the track is disabled. If we went the way of Chrome, that could solve this issue though might be a big unexpected change for users of this library.

I implemented the Firefox behavior explicitly :-) Google once told me they'd eventually implement it too but was not high priority. Stopping the capture when enabled is set to false makes video muting way faster because there is no need to destroy and re-create the track, as we have to do on Chrome, for example.

davidliu:
>>Normally this would fine since we the web API of MediaStreamTrack doesn't have start function, but did find a snag here; MediaStreamTrack.setEnabled additionally starts and stops the capturer, so users could resume a capturer that way.

>That's an internal API and it's not exposed so that's ok.

I mention it here because although it's internal behavior, it's something a user might inadvertently do; it would go against spec if we emit an "ended" event due to camera error, but the user flips the enabled event and winds up successfully restarting the capturer.

>I implemented the Firefox behavior explicitly :-) Google once told me they'd eventually implement it too but was not high priority. Stopping the capture when enabled is set to false makes video muting way faster because there is no need to destroy and re-create the track, as we have to do on Chrome, for example.

Hmm, if that's the case, I think I'll abandon trying to do `ended` events for the camera cases since there's no permanent end state that I know of there (aside from permission revocation, but those kill the app on iOS/Android), and just leave the ended events for the external screen capture stop.

saghul:
Sounds good to me!

davidliu:
Cleaned up the PR to only do screen capturer events, take a look when you can.

src/MediaStreamTrack.ts
```diff
@@ -36,6 +39,8 @@ class MediaStreamTrack extends defineCustomEventTarget(...MEDIA_STREAM_TRACK_EVE
         this.id = info.id;
         this.kind = info.kind;
         this.remote = info.remote;
+
+        this._registerEvents();
```
saghul, line 43: We should only do this for local tracks. It's not necessary to release remote tracks, so those will leak.

saghul (changes requested):
One final comment, LGTM otherwise.

src/MediaStreamTrack.ts
davidliu: Added in a check on `this.remote` and tested, seems to work fine there. 

A little worried about the SerializeUtils functions always returning true for remote though:

https://github.com/react-native-webrtc/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/SerializeUtils.java#L76

https://github.com/react-native-webrtc/react-native-webrtc/blob/0651e2b0600435dacd7b1e13d5f47914252e1e35/ios/RCTWebRTC/SerializeUtils.m#L174

I don't think there's a way for this flag to be incorrectly set for local tracks atm though, so should be fine if this isn't misused.

saghul: Hum good point, we should check that.

saghul (approved):

PR #1272, opened 2023-01-06
Clear iOS videoView by rendering a 1x1 blank frame before changing track

### LINKED ISSUE #1233: setVideoTrack clear previous frame
#### Expected Behavior
It should clear the latest frame when I change stream URL.

#### Observed Behavior
The old frame remains until the peer connection is established.

#### Steps to reproduce the issue
When I change stream URL, It doesn't clear the frame and the frame freezes until peer connection established.
Android has `cleanSurfaceViewRenderer` function and It runs in `setVideoTrack` function. There is no cleaner function for iOS. iOS sets _videoSize to zero but It doesn't clear the frame.

#### Platform Information

* **React Native Version**: 0.70.3
* **Plugin Version**: 1.106.1
* **OS**: iOS
* **OS Version**: 15.5

davidliu:
Fixes #1233.

### DIFF
--- a/ios/RCTWebRTC/RTCVideoViewManager.m
+++ b/ios/RCTWebRTC/RTCVideoViewManager.m
@@ -321,15 +321,15 @@ typedef NS_ENUM(NSInteger, RTCVideoViewObjectFit) {
     CVReturn err = CVPixelBufferCreate(NULL, 1, 1, kCVPixelFormatType_32BGRA, NULL, &pixelBuffer);
     if (err == kCVReturnSuccess) {
       const int kBytesPerPixel = 4;
-      CVPixelBufferLockBaseAddress( pixelBuffer, 0 );
-      int bufferWidth = (int)CVPixelBufferGetWidth( pixelBuffer );
-      int bufferHeight = (int)CVPixelBufferGetHeight( pixelBuffer );
-      size_t bytesPerRow = CVPixelBufferGetBytesPerRow( pixelBuffer );
-      uint8_t *baseAddress = CVPixelBufferGetBaseAddress( pixelBuffer );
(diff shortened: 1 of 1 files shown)


ios/RCTWebRTC/RTCVideoViewManager.m
```diff
@@ -314,6 +316,38 @@ - (void)setVideoTrack:(RTCVideoTrack *)videoTrack {
 
     _videoTrack = videoTrack;
 
+    // Clear the videoView by rendering a 1x1 blank frame.
+    CVPixelBufferRef pixelBuffer;
+    CVReturn err = CVPixelBufferCreate(NULL, 1, 1, kCVPixelFormatType_32BGRA, NULL, &pixelBuffer);
+    if (err != kCVReturnSuccess) {
```
saghul, line 322: Is this error handling correct? We are only doing stuff if we don't succeed?

ios/RCTWebRTC/RTCVideoViewManager.m
```diff
@@ -314,6 +316,38 @@ - (void)setVideoTrack:(RTCVideoTrack *)videoTrack {
...
+      const int kBytesPerPixel = 4;
+      CVPixelBufferLockBaseAddress( pixelBuffer, 0 );
```
saghul, line 324: style, here and elsewhere: please don't put spaces around (

ios/RCTWebRTC/RTCVideoViewManager.m
davidliu: Whoops! Thanks for catching this.

saghul (approved):
Nice work! Please squash when merging and adapt the commit message to the usual convention we follow, thank you!

Issue #1087, opened 2021-10-13
Inconsistency in deviceId between enumerateDevices() and getUserMedia()'s constraints on Android


audricschiltknecht:
#### Expected behavior
The deviceId returned by `enumerateDevices()` can be used as a constraint in `getUserMedia()`.

#### Observed behavior
The deviceId returned by `enumerateDevices()` is the index of the device in the list returned from `cameraEnumerator.getDeviceNames();`.
The deviceId used in the constraint parameter of `getUserMedia()` is expected to be the device name.

#### Steps to reproduce the problem

We can see https://github.com/react-native-webrtc/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java#L149 here that the `deviceId` put in the returned array from `enumerateDevices()` is the index of the device in the list returned by `cameraEnumerator.getDeviceNames();`.
However in https://github.com/react-native-webrtc/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java#L134, we compare the `deviceId` from the constraint (ie. an int converted to string) with the device name (which could be whatever string).

In my reading of the WebRTC's specs, these 2 values should match.

#### Platform information

* **React Native version**: 0.64.2
* **Plugin version**: 1.89.1
* **OS**: Android
* **OS version**: 10

Hesowcharov:
Noticed the same issue and would like to provide additional observations (hope it may be helpful for further update/rewriting).

In my personal case, I see the difference in output of `enumerateDevices` between android's Camera1Enumerator and Camera2Enumerator APIs.

Camera1Enumerator (Xiaomi miPad 4, android 8.1):
```
{"deviceId": "0", "facing": "environment", "groupId": "", "kind": "videoinput", "label": "Camera 0, Facing back, Orientation 90"}
{"deviceId": "1", "facing": "front", "groupId": "", "kind": "videoinput", "label": "Camera 1, Facing front, Orientation 270"}
```

Camera2Enumerator (Xiaomi Mi 9 Lite, android 10):
```
{"deviceId": "0", "facing": "environment", "groupId": "", "kind": "videoinput", "label": "0"}
{"deviceId": "1", "facing": "front", "groupId": "", "kind": "videoinput", "label": "1"}
```

Take a look at `deviceId` and `label`. 
[...]

PR #1273, opened 2023-01-06
fix: camera switch issue in devices using Camera1Enumerator

khushal87:
When we try to enumerate devices in the mobile devices using `Camera1Enumerator`, the device response returned when used to switch the camera doesn't change the face of the device; rather it mirrors the front-facing camera. On `Camera2Enumerator` devices, it works fine. To address this issue, we made the `deviceId` the same as the `label` of the device. On `Camera2Enumerator`, this wasn't an issue because the deviceId and the label are the same as the index. On `Camera1Enumerator` devices, we do the same thing and this solves the issue.

Pixel 10, Android 12(Camera2Enumerator)
```
{"deviceId": "0", "facing": "environment", "groupId": "", "kind": "videoinput", "label": "0"}
{"deviceId": "1", "facing": "front", "groupId": "", "kind": "videoinput", "label": "1"}
```

Xiaomi Redmi Y2, Android 9(Camera1Enumerator)
```
{"deviceId": "0", "facing": "environment", "groupId": "", "kind": "videoinput", "label": "Camera 0, Facing back, Orientation 90"}
{"deviceId": "1", "facing": "front", "groupId": "", "kind": "videoinput", "label": "Camera 1, Facing front, Orientation 270"}
```

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
+++ b/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
@@ -152,7 +152,7 @@ class GetUserMediaImpl {
             }
             WritableMap params = Arguments.createMap();
             params.putString("facing", isFrontFacing ? "front" : "environment");
-            params.putString("deviceId", "" + i);
+            params.putString("deviceId", deviceName);
             params.putString("groupId", "");
             params.putString("label", deviceName);
             params.putString("kind", "videoinput");


saghul:
This will break if devices actually have a name, which happens when an external camera is used for example.

Are you sure the bug isn't present elsewhere?

khushal87:
I created a patch in my project, and this change fixed the issue for both old and newer Android versions. For iOS, it isn't an issue. It's hard to say if adding an external camera will cause an issue.

saghul:
It's not clear to me what you are fixing here exactly.

khushal87:
Basically, if you try to `enumerateDevices` and use the response of the device received to change/toggle the camera device from the front to the back camera on old android devices, you will get your image mirrored rather than switching the camera. This happens because the `deviceId` is different from the device's label. On the new version of CameraEnumerator, it isn't an issue because the deviceId and label have the same value. This solution is added to solve this issue on old devices and, thereby, all the devices. Many users never face this issue because they use `_switchCamera`to switch the camera devices. But this solution addresses if someone wants to use the `enumerateDevices` API.

More here - https://github.com/react-native-webrtc/react-native-webrtc/issues/1087

saghul:
I think the root of the problem is here: https://github.com/react-native-webrtc/react-native-webrtc/blob/5c515f02fc9970c621b5730d7a06742115d6362b/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java#L132

We are incorrectly checking for the device ID, checking for the name instead.

The correct solution here is to fix searching for the ID, not turning the ID into a copy of the label, which seems to coincidentally match in some cases.

An ID should be an identifier, and the label a human readable name. This is clearer in iOS, but in Android we had to get a bit creative and use the device array index.

khushal87:
Hey @saghul, thanks for responding. To know do you understand the problem I am referring to here? I will be happy to deep dive to solve it with the correct fix but just needed to know if you, too, find it an issue.

saghul:
I suspect the problem is the wrong camera is being opened because the device match logic is incorrect.

It may work on one of the enumerators, but it's certainly wrong.

santhoshvai:
Hi, @saghul I created a PR based on your suggestion. Please have a look at https://github.com/react-native-webrtc/react-native-webrtc/pull/1277

Issue #1282, opened 2023-01-12
UVC Webcam/Mic Support


cristiano-linvix:
Hello!!

Any planing to add support to UVC?

I found some projects on the internet that use GoPro as an example.

[android_uvc_webrtc](https://github.com/kuanting/android_uvc_webrtc)
[UvcCapturer.java](https://github.com/kuanting/android_uvc_webrtc/blob/main/UVCWebRTC/app/src/main/java/net/aiwings/uvcwebrtc/UvcCapturer.java)

[android-webrtc-usb-camera](https://github.com/sutogan4ik/android-webrtc-usb-camera)
[UsbCapturer.java](https://github.com/sutogan4ik/android-webrtc-usb-camera/blob/master/UsbCapturer.java)

I don't know how to integrate these examples within the current project.

It looks like the api being used to enumerate the devices has UVC support, or not?

I found it here:
> https://source.android.com/docs/core/camera/external-usb-cameras

The Android platform supports the use of plug-and-play USB cameras (that is, webcams) using the standard [Android Camera2 API](https://developer.android.com/reference/android/hardware/camera2/package-summary.html) and the camera [HIDL](https://android.googlesource.com/platform/hardware/interfaces/+/refs/heads/master/camera/provider/2.4/ICameraProvider.hal) interface. Webcams generally support [USB video class (UVC)](https://en.wikipedia.org/wiki/USB_video_device_class) drivers and on Linux, the standard [Video4Linux (V4L)](https://en.wikipedia.org/wiki/Video4Linux) driver is used to control UVC cameras.

cristiano-linvix:
I found [your comment](https://github.com/jitsi/jitsi-meet/issues/4937#issuecomment-565400854) about the [license](https://github.com/sutogan4ik/android-webrtc-usb-camera/issues/2) of [package](https://github.com/sutogan4ik/android-webrtc-usb-camera/blob/master/UsbCapturer.java), and its Apache, can be used, rigth?

cristiano-linvix:
I'd really like to write code for this, but my skills are limited in native code.
If you give me a way, I can actually test get this to work.

saghul:
The [AndroidUSBCamera](https://github.com/jiangdongguo/AndroidUSBCamera) library looks quite nice!

Integrating it here would require that we have some kind of "meta enumerator" that lists all standard and USB cameras, and then use the right capturer depending on the chosen one.

[...]

Issue #828, opened 2020-07-10
Android: External (UVC) camera support


williamtdepaula:
I need to display a video from an external USB OTG camera, but it is not listed in mediaDevices.enumerateDevices (). In other applications with OTG support, displays the video correctly.

#### Expected behavior
Enumerate and display video from external camera

#### Steps to reproduce the problem
Connect a USB OTG camera and try to display video

* **React Native version**: 0.62.2
* **Plugin version**: 1.75.3
* **OS**: Android 
* **OS version**: 9.0

saghul:
WebRTC doesn't support this natively, so a new capturer would need to be added. There are some available but they need to have a compatible license (GPL is alas, not an option).

Previous conversations on the topic: https://github.com/jitsi/jitsi-meet/issues/4937

sohel-khan:
@williamtdepaula  I have publish a npm  if you still looking for solution then here it is : https://www.npmjs.com/package/react-native-webrtc-usb-lib

saghul:
Sorry, but why on earth would you publish a fork of this library instead of sending a proper contribution? People would now have to choose between this library, and yours, which doesn't even maintain Git history, so it's hard to tell how to far behind it is, for example.

As a seasoned open source maintainer, I find this attitude quite hostile.

danjenkins:
A fork of a fork! usb-lib is forked from https://github.com/jatecl/react-native-webrtc-usb which was forked from r-n-webrtc - @sohel-khan any interest in migrating those changes back into this project?

sohel-khan:
@saghul  @danjenkins , It was not meant to that,  I searched lot of of solutions on internet for me to be implement in my project but i have not found any working library, so  i have clone react-native-webrtc-usb library which is deprecated and not maintained since 3 years, i have to dig into that and normalized back into work by upgrading gradle and react native versions. I thought someone like who need this and will find easy consume their further project so that i have published as new npm.

hthetiot:
I would not call that a "fork", since @sohel-khan removed all the git history: https://github.com/sohel-khan/react-native-webrtc-usb-lib/commits/main

saghul:
That just makes it even worse. Also that "fork" doesn't work with normal cameras because it just uses the other enumerator.

sohel-khan:
[...]

PR #1285, opened 2023-01-13
add uvc webcam support

cristiano-linvix:
Hello goodnight!

I made this library integration [AndroidUSBCamera](https://github.com/jiangdongguo/AndroidUSBCamera)

I hope it's useful for other people.

I would like to do the integration with the webcam's audio too, if I can I'll do another PR.

I'm not strong in java programming, I did what it takes to get it working.

https://github.com/react-native-webrtc/react-native-webrtc/issues/1282
https://github.com/react-native-webrtc/react-native-webrtc/issues/828

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
+++ b/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
@@ -1,8 +1,10 @@
 package com.oney.WebRTCModule;
 
+import android.content.Context;
 import android.util.Log;
 
 import com.facebook.react.bridge.ReadableMap;
+import com.jiangdg.ausbc.camera.CameraUvcStrategy;
 
 import org.webrtc.CameraEnumerator;
 import org.webrtc.CameraVideoCapturer;
@@ -22,6 +24,7 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
 
     private final CameraEnumerator cameraEnumerator;
     private final ReadableMap constraints;
+    private final CameraUvcStrategy mUvcStrategy;
 
     /**
      * The {@link CameraEventsHandler} used with
@@ -31,14 +34,16 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
      */
     private final CameraEventsHandler cameraEventsHandler = new CameraEventsHandler();
 
-    public CameraCaptureController(CameraEnumerator cameraEnumerator, ReadableMap constraints) {
+    public CameraCaptureController(CameraEnumerator cameraEnumerator, ReadableMap constraints, CameraUvcStrategy uvcStrategy) {
         super(
              constraints.getInt("width"),
-             constraints.getInt("height"), 
-             constraints.getInt("frameRate"));
+             constraints.getInt("height"),
+             constraints.getInt("frameRate")
+     );
 
         this.cameraEnumerator = cameraEnumerator;
         this.constraints = constraints;
+        this.mUvcStrategy = uvcStrategy;
     }
 
     public void switchCamera() {
@@ -86,7 +91,7 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
      * Helper function which tries to switch cameras until the desired facing mode is found.
      *
      * @param desiredFrontFacing - The desired front facing value.
-     * @param tries - How many times to try switching.
+     * @param tries              - How many times to try switching.
      */
     private void switchCamera(boolean desiredFrontFacing, int tries) {
         CameraVideoCapturer capturer = (CameraVideoCapturer) videoCapturer;
@@ -95,7 +100,7 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
             @Override
             public void onCameraSwitchDone(boolean b) {
                 if (b != desiredFrontFacing) {
-                    int newTries = tries-1;
+                    int newTries = tries - 1;
                     if (newTries > 0) {
                         switchCamera(desiredFrontFacing, newTries);
                     }
@@ -115,18 +120,35 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
      * Constructs a new {@code VideoCapturer} instance attempting to satisfy
      * specific constraints.
      *
-     * @param deviceId the ID of the requested video device. If not
-     * {@code null} and a {@code VideoCapturer} can be created for it, then
-     * {@code facingMode} is ignored.
+     * @param deviceId   the ID of the requested video device. If not
+     *                   {@code null} and a {@code VideoCapturer} can be created for it, then
+     *                   {@code facingMode} is ignored.
      * @param facingMode the facing of the requested video source such as
-     * {@code user} and {@code environment}. If {@code null}, "user" is
-     * presumed.
+     *                   {@code user} and {@code environment}. If {@code null}, "user" is
+     *                   presumed.
      * @return a {@code VideoCapturer} satisfying the {@code facingMode} or
      * {@code deviceId} constraint
      */
     private VideoCapturer createVideoCapturer(String deviceId, String facingMode) {
         String[] deviceNames = cameraEnumerator.getDeviceNames();
         List<String> failedDevices = new ArrayList<>();
+
+        // If deviceId is specified, then it takes precedence over facingMode.
+        if (deviceId != null) {
+            // if has uvc into id, instance uvc capturer
+            if (deviceId.contains("uvc")) {
+                String message = "Create UVC user-specified camera " + deviceId;
+                try {
+                    UvcCapturer videoCapturer = new UvcCapturer(deviceId, mUvcStrategy);
+                    Log.d(TAG, message + " succeeded");
+                    return videoCapturer;
+                } catch (Throwable tr) {
+                    failedDevices.add(deviceId);
+                    Log.d(TAG, message + " failed");
+                }
+            }
+        }
+
 
         String cameraName = null;
         try {
--- a/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
+++ b/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
@@ -136,14 +136,13 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
         // If deviceId is specified, then it takes precedence over facingMode.
         if (deviceId != null) {
             // if has uvc into id, instance uvc capturer
-            if (deviceId.contains("uvc")) {
+            if (deviceId.contains("uvc:")) {
                 String message = "Create UVC user-specified camera " + deviceId;
                 try {
                     UvcCapturer videoCapturer = new UvcCapturer(deviceId, mUvcStrategy);
                     Log.d(TAG, message + " succeeded");
                     return videoCapturer;
                 } catch (Throwable tr) {
-                    failedDevices.add(deviceId);
                     Log.d(TAG, message + " failed");
                 }
             }
--- a/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
+++ b/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
@@ -18,6 +18,7 @@ import com.facebook.react.bridge.UiThreadUtil;
 import com.facebook.react.bridge.WritableArray;
 import com.facebook.react.bridge.WritableMap;
 
+import com.jiangdg.ausbc.camera.CameraUvcStrategy;
 import com.oney.WebRTCModule.videoEffects.ProcessorProvider;
 import com.oney.WebRTCModule.videoEffects.VideoEffectProcessor;
 import com.oney.WebRTCModule.videoEffects.VideoFrameProcessor;
@@ -56,9 +57,15 @@ class GetUserMediaImpl {
     private Promise displayMediaPromise;
     private Intent mediaProjectionPermissionResultData;
 
+    private UvcEnumerator uvcEnumerator;
+    private CameraUvcStrategy cameraUvcStrategy;
+
     GetUserMediaImpl(WebRTCModule webRTCModule, ReactApplicationContext reactContext) {
         this.webRTCModule = webRTCModule;
         this.reactContext = reactContext;
+
+        this.cameraUvcStrategy = new CameraUvcStrategy(reactContext);
+        this.uvcEnumerator = new UvcEnumerator(this.cameraUvcStrategy);
 
         boolean camera2supported = false;
 
@@ -138,7 +145,15 @@ class GetUserMediaImpl {
 
     ReadableArray enumerateDevices() {
         WritableArray array = Arguments.createArray();
+
+        ReadableArray uvcDevices = uvcEnumerator.enumerateDevices();
         String[] devices = cameraEnumerator.getDeviceNames();
+
+        if (uvcDevices.size() > 0) {
+            for (int j = 0; j < uvcDevices.size(); j++) {
+                array.pushMap(uvcDevices.getMap(j));
+            }
+        }
 
         for (int i = 0; i < devices.length; ++i) {
             String deviceName = devices[i];
@@ -201,7 +216,9 @@ class GetUserMediaImpl {
 
             CameraCaptureController cameraCaptureController = new CameraCaptureController(
                 cameraEnumerator,
-                videoConstraintsMap);
+                videoConstraintsMap,
+                cameraUvcStrategy
+        );
 
             videoTrack = createVideoTrack(cameraCaptureController);
         }
@@ -487,7 +504,7 @@ class GetUserMediaImpl {
                  * called. This also means that the caller can reuse the SurfaceTextureHelper to initialize a new
                  * VideoCapturer once the previous VideoCapturer has been disposed. */
                 
-                if(surfaceTextureHelper != null) {
+                if (surfaceTextureHelper != null) {
                     surfaceTextureHelper.stopListening();
                     surfaceTextureHelper.dispose();
                 }
--- a/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
+++ b/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
@@ -105,7 +105,7 @@ public class UvcCapturer implements VideoCapturer {
         return new CameraRequest
                 .Builder()
                 .setCameraId(this.deviceId)
-                .setFrontCamera(false)
+                // .setFrontCamera(false)
                 .setContinuousAFModel(true)
                 .setContinuousAutoModel(true)
                 .setPreviewWidth(UVC_PREVIEW_WIDTH)
--- a/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
+++ b/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
@@ -27,8 +27,8 @@ public class UvcCapturer implements VideoCapturer {
     private SurfaceTextureHelper surfaceTextureHelper;
     private CapturerObserver capturerObserver;
 
-    private int UVC_PREVIEW_WIDTH = 1280;
-    private int UVC_PREVIEW_HEIGHT = 720;
+    private int UVC_PREVIEW_WIDTH = 1024;
+    private int UVC_PREVIEW_HEIGHT = 768;
     private int UVC_PREVIEW_FPS = 30;
 
     CameraUvcStrategy mUvcStrategy;
--- a/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
+++ b/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
@@ -31,13 +31,11 @@ public class UvcCapturer implements VideoCapturer {
     private int UVC_PREVIEW_HEIGHT = 720;
     private int UVC_PREVIEW_FPS = 30;
 
-    CameraUvcStrategy mUvcStrategy; // From Jiang Dongguo's AUSBC library
+    CameraUvcStrategy mUvcStrategy;
 
     public UvcCapturer(String deviceId, CameraUvcStrategy uvcStrategy) {
-        this.deviceId = deviceId.replaceAll("\\D", "");
-
+        this.deviceId = deviceId.replace("uvc:", "");
         Log.d(TAG, "UvcCapturer.UvcCapturer->DeviceID " + this.deviceId);
-
         mUvcStrategy = uvcStrategy;
     }
 
--- a/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
+++ b/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
@@ -27,8 +27,8 @@ public class UvcCapturer implements VideoCapturer {
     private SurfaceTextureHelper surfaceTextureHelper;
     private CapturerObserver capturerObserver;
 
-    private int UVC_PREVIEW_WIDTH = 1024;
-    private int UVC_PREVIEW_HEIGHT = 768;
+    private int UVC_PREVIEW_WIDTH = 1280;
+    private int UVC_PREVIEW_HEIGHT = 720;
     private int UVC_PREVIEW_FPS = 30;
 
     CameraUvcStrategy mUvcStrategy; // From Jiang Dongguo's AUSBC library
@@ -38,11 +38,7 @@ public class UvcCapturer implements VideoCapturer {
 
         Log.d(TAG, "UvcCapturer.UvcCapturer->DeviceID " + this.deviceId);
 
-        try {
-            mUvcStrategy = uvcStrategy;
-        } catch (Exception e) {
-            Log.e(TAG, "UvcCapturer.UvcCapturer ", e);
-        }
+        mUvcStrategy = uvcStrategy;
     }
 
     @Override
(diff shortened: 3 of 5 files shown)


android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public void changeCaptureFormat(int i, int i1, int i2) {
+
+    }
+
+    @Override
+    public void dispose() {
+        Log.d(TAG, "UvcCapturer.dispose");
```
cristiano-linvix, line 95: I need help to get to that point, to free the camera.

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
```diff
@@ -0,0 +1,73 @@
...
+
+                if (!usbDevices.isEmpty()) {
+
+                    for (int i = 0; i < usbDevices.size(); i++) {
+                        UsbDevice dev = usbDevices.get(i);
+
+                        WritableMap params = Arguments.createMap();
+
+                        params.putString("facing", "uvc");
+                        params.putString("deviceId", "uvc-" + dev.getDeviceId());
```
cristiano-linvix, line 49: I used the uvc prefix on the deviceId to be able to identify when looking for a uvc device in the CameraController.

cristiano-linvix: There in UvcCapturer I remove the uvc prefix.

8BallBomBom:
Just a few thoughts.
After an initial look there are definitely some changes that could be made.

The `UvcCapturer` should probably extend the `AbstractVideoCaptureController` same as the `ScreenCaptureController` and `CameraCaptureController`, then it can also follow naming conventions like `USBCaptureController`.
You would also not then need the changes you've made within the `CameraCaptureController`.

But also rather than tying the grabbing logic in with `getUserMedia` instead make a function for it similar to `getDisplayMedia` which would also include some extra utility functions to handle creating the `USBCaptureController` and getting tracks.

Last but not least the actual exported `ReactMethod` can be handled within the `WebRTCModule` and we'd also need to work on some basic interfacing via typescript.

@saghul What do you think?

cristiano-linvix:
Hello, thanks for the comment.

I think extending the AbstractVideoCaptureController really should be done.

Renaming the files to USBCapturer and USBEnumerator can be done if needed.

But, I don't see the need to create another function to handle the camera since it is used in getUserMedia, I think it "fits" the purpose of the function.

I await further information.

@saghul @8BallBomBom

cristiano-linvix:
Hello, how are you there?

Did you look at this?

@saghul

saghul:
Hi there! Thanks for sending the PR! I'm a bit busy now, I'll take a look as soon as I can.

cristiano-linvix:
Hello, did you analyze this PR?

saghul:
Not yet, sorry. It's been a bit busy. Hoping tkk on get to it next week though.

saghul:
Can you please rebase so the new CI runs?

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -115,19 +120,36 @@ public void onCameraSwitchError(String s) {
...
      * {@code deviceId} constraint
      */
     private VideoCapturer createVideoCapturer(String deviceId, String facingMode) {
         String[] deviceNames = cameraEnumerator.getDeviceNames();
         List<String> failedDevices = new ArrayList<>();
 
+        // If deviceId is specified, then it takes precedence over facingMode.
+        if (deviceId != null) {
+            // if has uvc into id, instance uvc capturer
+            if (deviceId.contains("uvc")) {
```
saghul, line 139: maybe we should have them start with `uvc:` instead? And then check if it _starts_ with that prefix.

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -115,19 +120,36 @@ public void onCameraSwitchError(String s) {
...
+                String message = "Create UVC user-specified camera " + deviceId;
+                try {
+                    UvcCapturer videoCapturer = new UvcCapturer(deviceId, mUvcStrategy);
+                    Log.d(TAG, message + " succeeded");
+                    return videoCapturer;
+                } catch (Throwable tr) {
+                    failedDevices.add(deviceId);
```
saghul, line 146: The fallback code won't work with this I think, because we are not enumerating the UVC cameras.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -201,7 +216,9 @@ void getUserMedia(
 
             CameraCaptureController cameraCaptureController = new CameraCaptureController(
                 cameraEnumerator,
-                videoConstraintsMap);
+                videoConstraintsMap,
+                cameraUvcStrategy
```
saghul, line 220: can't we pass the enumerator instead? feels more natural to have 1 "meta" enumerator.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
```diff
@@ -0,0 +1,116 @@
...
+    private String deviceId;
+
+    private SurfaceTextureHelper surfaceTextureHelper;
+    private CapturerObserver capturerObserver;
+
+    private int UVC_PREVIEW_WIDTH = 1024;
+    private int UVC_PREVIEW_HEIGHT = 768;
+    private int UVC_PREVIEW_FPS = 30;
+
+    CameraUvcStrategy mUvcStrategy; // From Jiang Dongguo's AUSBC library
```
saghul, line 34: Drop the commend, it's not helpful.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private static final String TAG = WebRTCModule.TAG;
+
+    private String deviceId;
+
+    private SurfaceTextureHelper surfaceTextureHelper;
+    private CapturerObserver capturerObserver;
+
+    private int UVC_PREVIEW_WIDTH = 1024;
+    private int UVC_PREVIEW_HEIGHT = 768;
```
saghul, line 31: This is a very weird default resolution. Shall we make it 1280x720 aka 720p ?

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
```diff
@@ -0,0 +1,116 @@
...
+    private int UVC_PREVIEW_FPS = 30;
+
+    CameraUvcStrategy mUvcStrategy; // From Jiang Dongguo's AUSBC library
+
+    public UvcCapturer(String deviceId, CameraUvcStrategy uvcStrategy) {
+        this.deviceId = deviceId.replaceAll("\\D", "");
```
saghul, line 37: Why is this necessary?

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
```diff
@@ -0,0 +1,116 @@
...
+
+        Log.d(TAG, "UvcCapturer.UvcCapturer->DeviceID " + this.deviceId);
+
+        try {
+            mUvcStrategy = uvcStrategy;
```
saghul, line 42: How could this fail? It's just a variable assignment.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public void startCapture(int width, int height, int fps) {
+        if (mUvcStrategy != null) {
+            // stop before to free last usb instance
+            if (mUvcStrategy.isCameraOpened()) {
+                mUvcStrategy.stopPreview();
+            }
+
+            UVC_PREVIEW_WIDTH = width;
```
saghul, line 62: why do we need to save these here?

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
```diff
@@ -0,0 +1,116 @@
...
+            UVC_PREVIEW_HEIGHT = height;
+            UVC_PREVIEW_FPS = fps;
+
+            mUvcStrategy.addPreviewDataCallBack(new IPreviewDataCallBack() {
+                @Override
+                public void onPreviewData(@Nullable byte[] bytes, @NonNull DataFormat dataFormat) {
```
saghul, line 68: Why is dataFormat not being used?

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
saghul: What do you mean?

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private CameraRequest getCameraRequest() {
+        return new CameraRequest
+                .Builder()
+                .setCameraId(this.deviceId)
+                // .setFrontCamera(false)
+                .setContinuousAFModel(true)
+                .setContinuousAutoModel(true)
+                .setPreviewWidth(UVC_PREVIEW_WIDTH)
+                .setPreviewHeight(UVC_PREVIEW_HEIGHT)
```
saghul, line 112: You are not specifying the fps here. Is there a way to do that?

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
```diff
@@ -0,0 +1,73 @@
...
+
+import org.webrtc.SurfaceViewRenderer;
+
+import java.util.List;
+
+public class UvcEnumerator {
+
+    private static final String TAG = WebRTCModule.TAG;
+
+    CameraUvcStrategy mUvcStrategy; // From Jiang Dongguo's AUSBC library
```
saghul, line 23: Drop the comment please.

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public UvcEnumerator(CameraUvcStrategy cameraUvcStrategy) {
+        try {
+            mUvcStrategy = cameraUvcStrategy;
```
saghul, line 27: How can this ever fail?

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
```diff
@@ -0,0 +1,73 @@
...
+            if (usbDevices != null) {
+
+                if (!usbDevices.isEmpty()) {
+
+                    for (int i = 0; i < usbDevices.size(); i++) {
+                        UsbDevice dev = usbDevices.get(i);
+
+                        WritableMap params = Arguments.createMap();
+
+                        params.putString("facing", "uvc");
```
saghul, line 48: Can't we get the right facing mode? If not, let's just say "environment".

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
saghul: Good idea!

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
```diff
@@ -0,0 +1,73 @@
...
+                        params.putString("label", dev.getProductName() != null ? dev.getProductName() : dev.getDeviceName());
+                        params.putString("kind", "videoinput");
+
+                        Log.d(TAG, "UvcEnumerator.enumerateDevices: " + params);
+
+                        array.pushMap(params);
+                    }
+
+                } else {
+                    Log.d(TAG, "UvcEnumerator.enumerateDevices: usbDevices is empty");
```
saghul, line 60: I'd remove this log line.

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
```diff
@@ -0,0 +1,73 @@
...
+                }
+
+            } else {
+                Log.d(TAG, "UvcEnumerator.enumerateDevices: usbDevices is null");
```
saghul, line 64: Ditto.

saghul (changes requested):
Finally managed to do a full review, please take a look!

saghul:
Question: is there a chance a camera is detected by both enumerators?

cristiano-linvix:
Hello @saghul, I'll start on the changes as soon as possible, I'm a little busy this month.

Thank you!

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
cristiano-linvix: It would be to change the uvc- for uvc: ?

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
cristiano-linvix: Okay, so I'll remove the snippet.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
cristiano-linvix: Sorry, that I didn't understand.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
cristiano-linvix: I will remove the comment.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
cristiano-linvix: WebCam resolutions are very different between models, I'll try using 720 instead of 728 and see if it works.

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
cristiano-linvix: I'll look into it.

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
cristiano-linvix: I'll switch to environment.

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
cristiano-linvix: I will.

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
cristiano-linvix: I will.

saghul:
No worries, thank you!

cristiano-linvix:
> Question: is there a chance a camera is detected by both enumerators?

It depends on the android device, if it supports UVC devices cameras, it may appear in the first enumerator too.

https://source.android.com/docs/core/camera/external-usb-cameras#implementation

saghul:
I feared that. In that case, we'll need to filter it, and prefer the non-UVC one, I'd say.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
cristiano-linvix: The data pattern that is received is always the same, you don't need to detect it.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
cristiano-linvix: Maximum FPS is 30.
Its not possible to change. :(

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
cristiano-linvix: The getCameraRequest function uses this variable to get the webcam that uses this resolution.
The startCapture function receives the size there from the controller.

cristiano-linvix:
@saghul I sent it for your analysis again, as soon as you can look, thank you.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
```diff
@@ -0,0 +1,110 @@
...
+public class UvcCapturer implements VideoCapturer {
+
+    private static final String TAG = WebRTCModule.TAG;
+
+    private String deviceId;
+
+    private SurfaceTextureHelper surfaceTextureHelper;
+    private CapturerObserver capturerObserver;
+
+    private int UVC_PREVIEW_WIDTH = 1280;
```
cristiano-linvix, line 30: UVC Capturer resolutions are different, it looks like they are screen resolutions.

Logcat
```log
getAllPreviewSizes = [Size(1920x1080@ 0.0,type:4,frame:0,index:0,[]), Size(1024x768@ 0.0,type:4,frame:0,index:1,[]), Size(1024x576@ 0.0,type:4,frame:0,index:2,[]), Size(960x720@ 0.0,type:4,frame:0,index:3,[]), Size(864x480@ 0.0,type:4,frame:0,index:4,[]), Size(800x600@ 0.0,type:4,frame:0,index:5,[]), Size(640x480@ 0.0,type:4,frame:0,index:6,[]), Size(320x240@ 0.0,type:4,frame:0,index:7,[]), Size(1920x1080@ 0.0,type:4,frame:0,index:8,[])]
```

cristiano-linvix: It really needs to be 1024x768.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
cristiano-linvix: It really needs to be 1024x768.

The webcam resolutions are different, like monitor resolution.

logcat
```log
getAllPreviewSizes = [Size(1920x1080@ 0.0,type:4,frame:0,index:0,[]), Size(1024x768@ 0.0,type:4,frame:0,index:1,[]), Size(1024x576@ 0.0,type:4,frame:0,index:2,[]), Size(960x720@ 0.0,type:4,frame:0,index:3,[]), Size(864x480@ 0.0,type:4,frame:0,index:4,[]), Size(800x600@ 0.0,type:4,frame:0,index:5,[]), Size(640x480@ 0.0,type:4,frame:0,index:6,[]), Size(320x240@ 0.0,type:4,frame:0,index:7,[]), Size(1920x1080@ 0.0,type:4,frame:0,index:8,[])]
```

cristiano-linvix:
> I feared that. In that case, we'll need to filter it, and prefer the non-UVC one, I'd say.

The function is:
```java
public bool supportsExternalCameras() {
        return this.context.getPackageManager().hasSystemFeature(PackageManager.FEATURE_CAMERA_EXTERNAL);        
}
```

Put inside the Camera Capture Controller?
I would have to pass the context into the class.

saghul:
> The function is:
> 
> ```java
> public bool supportsExternalCameras() {
>         return this.context.getPackageManager().hasSystemFeature(PackageManager.FEATURE_CAMERA_EXTERNAL);        
> }
> ```
> 
> Put inside the Camera Capture Controller? I would have to pass the context into the class.

But that would only tell you if external cameras are supported, right? The problem is if they are, and both enumerators list them. How do you plan on filtering them?

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -31,14 +34,16 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
      */
     private final CameraEventsHandler cameraEventsHandler = new CameraEventsHandler();
 
-    public CameraCaptureController(CameraEnumerator cameraEnumerator, ReadableMap constraints) {
+    public CameraCaptureController(CameraEnumerator cameraEnumerator, ReadableMap constraints, CameraUvcStrategy uvcStrategy) {
```
saghul, line 37: Can you please make the uvcStrategy the 2nd parameter?

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -86,7 +91,7 @@ protected VideoCapturer createVideoCapturer() {
      * Helper function which tries to switch cameras until the desired facing mode is found.
      *
      * @param desiredFrontFacing - The desired front facing value.
-     * @param tries - How many times to try switching.
+     * @param tries              - How many times to try switching.
```
saghul, line 94: Please avoid making any unrelated changes. It complicates the diff for no good reason. I'm working on integrating clang-format to take care of this automatically.

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -115,19 +120,35 @@ public void onCameraSwitchError(String s) {
...
      * {@code deviceId} constraint
      */
     private VideoCapturer createVideoCapturer(String deviceId, String facingMode) {
         String[] deviceNames = cameraEnumerator.getDeviceNames();
         List<String> failedDevices = new ArrayList<>();
 
+        // If deviceId is specified, then it takes precedence over facingMode.
+        if (deviceId != null) {
+            // if has uvc into id, instance uvc capturer
+            if (deviceId.contains("uvc:")) {
```
saghul, line 139: Let's be precise and check if it starts with the prefix, not if it contains it.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
saghul: I think we can create the strategy inside the enumerator, and provide a getter for it.

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
```diff
@@ -0,0 +1,65 @@
...
+
+    ReadableArray enumerateDevices() {
+        WritableArray array = Arguments.createArray();
+
+        try {
+            List<UsbDevice> usbDevices = mUvcStrategy.getUsbDeviceList(null);
+
+            if (usbDevices != null) {
+
+                if (!usbDevices.isEmpty()) {
```
saghul, line 37: Is this necessary? The for look won't run if it's empty and that's that...

android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
```diff
@@ -0,0 +1,65 @@
...
+                        Log.d(TAG, "UvcEnumerator.enumerateDevices: " + params);
+
+                        array.pushMap(params);
+                    }
+
+                }
+
+            }
+
+        } catch (Throwable tr) {
```
saghul, line 59: Can you please try to make try-catch block as narrow was possible? That is, please only protect the calls that can throw, not the whole thing.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -138,8 +145,16 @@ private void checkMandatoryConstraints(MediaConstraints peerConstraints) {
 
     ReadableArray enumerateDevices() {
         WritableArray array = Arguments.createArray();
+
+        ReadableArray uvcDevices = uvcEnumerator.enumerateDevices();
         String[] devices = cameraEnumerator.getDeviceNames();
 
+        if (uvcDevices.size() > 0) {
+            for (int j = 0; j < uvcDevices.size(); j++) {
+                array.pushMap(uvcDevices.getMap(j));
```
saghul, line 154: We need to filter the devices. maybe you can do something like `uvcEnumerator.enumerateDevices(nativeDevices);` where `nativeDevices` is `cameraEnumerator.getDeviceNames();` and you can do the filtering there, if they match.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
saghul: Alternatively, figure it out dynamically by querying the supported sizes and picking the one closest to the request resolution that is passed in startCapture, that would make it even better.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
saghul: Please pass height, width, fps as parameters to `getCameraRequest()` and get rid off these.

android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
saghul: Doesn't this dispose method work? Alternatively, we can do it in start / stop. Start re-creates what's necessary and stop destroys it.

saghul (changes requested):
Left some more comments, this is looking better!

saghul:
The PR needs rebasing and addressing some of my comments still.

cristiano-linvix:
Hello @saghul, sorry for the delay, I ended up having to take time off work because of a medical procedure, but I'll be back in April.

saghul:
No worries, no rush at all.

Hope it all goes well!

cristiano-linvix:
Hello everybody!

I'm going to close this pull request, it's been open for a long time, and we need to implement this in the latest version of the project.

When I have time again, I start the project again and make a new pull request.

Thank you all!

Issue #1286, opened 2023-01-14
onicecandidate event not found


sharma12343:
onicecandidate event definition not found in the webrtc package

8BallBomBom:
Can you provide more information?

What module version are you using?
What React Native version are you using?
Which platform? Android or iOS?

Providing more information will allow us to actually help.

sharma12343:
"react-native-webrtc": "^106.0.4",
    "react": "18.1.0",
    "react-native": "^0.70.6",
    Android

8BallBomBom:
Are you experiencing any errors?
Also the event should fire when creating an offer.

sharma12343:
No error 
Event is not fired .
Also try with below 
pc.addEventListener("icecandidate", (ev) => {
  if (ev.candidate) {
    sendMessage({
      type: "new-ice-candidate",
      candidate: event.candidate
    });
  }
}, false);

8BallBomBom:
You could hook up to events more directly but that isn't advised.
Can you provide a project that we can use to test with to reproduce the issue?
I'm failing to reproduce the issue under a new project with basic peer creation.

DenisLebedinsky:
I have same problem, 'icecandidate' event doesn't work. (platform iOS and android)

iOS:
 Event was caught in Xcode 

<img width="941" alt="Screenshot 2023-01-15 at 12 49 34" src="https://user-images.githubusercontent.com/33514371/212531654-28941fbc-0142-40f3-b0b0-af002d388bdf.png">

<img width="563" alt="Screenshot 2023-01-15 at 12 30 58" src="https://user-images.githubusercontent.com/33514371/212531402-29457110-375e-4ce1-8021-b8b2efc7a3b4.png">

waring in metro console:

<img width="647" alt="Screenshot 2023-01-15 at 12 49 52" src="https://user-images.githubusercontent.com/33514371/212531677-f4f4b731-ebd0-49fd-ad03-3cb20eec0a1d.png">

and callback doesn't work.

Android:
Also I god icaCandidate on ja
<img width="1077" alt="Screenshot 2023-01-15 at 12 55 57" src="https://user-images.githubusercontent.com/33514371/212531984-41c21464-dd37-41ef-bc55-fca37b5f9500.png">
va and event doesn't work on js

DenisLebedinsky:
When I call setupNativeEvents() after adding all listeners it works correctly

DenisLebedinsky:
#1287

[...]

PR #1287, opened 2023-01-15
replace setupNativeEvents by

DenisLebedinsky:
adding native event listener on each js listener


saghul:
What prompted this change? AFACT it's functionality equivalent.

DenisLebedinsky:
> What prompted this change? AFACT it's functionality equivalent.

event doesn't work. #1286

PR #1293, opened 2023-01-17
dtmf implemented following the w3c API

AbdulBsit:


src/RTCRtpSender.ts
```diff
@@ -22,7 +24,7 @@ export default class RTCRtpSender {
         this._peerConnectionId = info.peerConnectionId;
         this._id = info.id;
         this._rtpParameters = new RTCRtpSendParameters(info.rtpParameters);
-
+        this._dtmf = new RTCDTMFSender(info.peerConnectionId);
```
saghul, line 27: We should pass the sender ID here and then use it in the native side to match it.

saghul: Also, what do browsers have as this property, for a video track, null?

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -608,6 +608,36 @@ public boolean peerConnectionRemoveTrack(int id, String senderId) {
...
+                }
+                
+                for (RtpSender sender : peerConnection.getSenders()) {
+                    if (sender.track().kind().equals("audio")) {
+                        audioSender = sender;
+                    }
+                }
+
+                if (audioSender != null) {
+                    DtmfSender dtmfSender = audioSender.dtmf();
```
saghul, line 629: Can this be null?

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -608,6 +608,36 @@ public boolean peerConnectionRemoveTrack(int id, String senderId) {
...
+        ThreadUtils.runOnExecutor(() -> {
+            try {
+                RtpSender audioSender = null;
+                PeerConnection peerConnection = getPeerConnection(objectID);
+                if (peerConnection == null) {
+                    Log.d(TAG, "peerConnectionSendDTMF() peerConnection is null");
+                    return;
+                }
+                
+                for (RtpSender sender : peerConnection.getSenders()) {
```
saghul, line 622: Match by sender ID instead

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -309,6 +309,31 @@ @implementation WebRTCModule (RTCPeerConnection)
...
+{
+  RTCRtpSender* audioSender = nil ;
+  RTCPeerConnection *peerConnection = self.peerConnections[objectID];
+  if (!peerConnection) {
+    NSLog(@"peerConnectionSendDTMF() peerConnection is nil");
+    return;
+  }
+  for (RTCRtpSender *sender in peerConnection.senders){
+    if([[[sender track] kind] isEqualToString:@"audio"]) {
+       audioSender = sender;
```
saghul, line 322: Match by sender ID

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -309,6 +309,31 @@ @implementation WebRTCModule (RTCPeerConnection)
...
+     } 
+  }
+  if(audioSender){
+    NSOperationQueue *queue = [[NSOperationQueue alloc] init];
```
saghul, line 326: Why are you using this queue here?

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -309,6 +309,31 @@ @implementation WebRTCModule (RTCPeerConnection)
...
+    [queue addOperationWithBlock:^{
+        BOOL istoneplayed = [audioSender.dtmfSender insertDtmf:tone duration:duration interToneGap:interToneGap];
```
saghul, line 328: Is `dtmfSender` nillable?

src/RTCDTMFSender.ts
```diff
@@ -0,0 +1,27 @@
...
+    constructor(peerConnectionId: number) {
+        this._peerConnectionId = peerConnectionId;
+    }
+
+    get canInsertDTMF() {
+        return !!this._peerConnectionId;
+    }
+
+    insertDTMF(tones: string, duration?: number, interToneGap?: number) {
+        const dur = duration || DEFAULT_DURATION;
```
saghul, line 18: Use default values in the function definition instead

saghul (changes requested):
Thanks for getting this started! I left some comments, please take a look!

Also, please undo the changes to package-lock.json, you seem to be using an old npm version.

8BallBomBom:
Very slightly started on this last month over [here](https://github.com/react-native-webrtc/react-native-webrtc/compare/master...8BallBomBom:react-native-webrtc:dtmf) 😅
But then realised `DtmfSender` inside the Android/iOS WebRTC SDK doesn't provide an observer.
Which meant not being able to listen to the `tonechange` event directly.
So that would then require some patches or checking the internal `tones` string for results i'd think.

Not sure which path we should pursue 🤔
Or where to go from here.

src/RTCRtpSender.ts
AbdulBsit: Done

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
AbdulBsit: Not sure, but i think it's better to check for null in case no audioSender found

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
AbdulBsit: Done

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
AbdulBsit: Done

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
AbdulBsit: I don't know, I take code from here https://stackoverflow.com/a/60148372 , I removed it now

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
AbdulBsit: No, It should always exist in audioSender

src/RTCDTMFSender.ts
AbdulBsit: Done

src/RTCRtpSender.ts
AbdulBsit: > Also, what do browsers have as this property, for a video track, null?

Not sure about this

AbdulBsit:
> Very slightly started on this last month over [here](https://github.com/react-native-webrtc/react-native-webrtc/compare/master...8BallBomBom:react-native-webrtc:dtmf) 😅 But then realised `DtmfSender` inside the Android/iOS WebRTC SDK doesn't provide an observer. Which meant not being able to listen to the `tonechange` event directly. So that would then require some patches or checking the internal `tones` string for results i'd think.
> 
> Not sure which path we should pursue 🤔 Or where to go from here.

Oo, I haven't thought of it, So I think then we should implement it from js side?

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
AbdulBsit: I think we should use queue here, [see here](https://chromium.googlesource.com/external/webrtc/+/HEAD/sdk/android/api/org/webrtc/DtmfSender.java#30)
It says 
```
 Queues a task that sends the provided DTMF tones.
 If insertDtmf is called on the same object while an existing task for this
 object to generate DTMF is still running, the previous task is canceled.
```
So what should we do?

8BallBomBom:
Unsure exactly why the observer isn't provided but it does exist in the spec and is there [internally](https://webrtc.googlesource.com/src/+/refs/heads/main/api/dtmf_sender_interface.h#21).
But either way, unusual 🤔

saghul:
The Java / ObjC is pretty much driven by the community at this point.

So if we want to implement that we'd need to wrap that event in the C++ layer.

That said, I'd say a partial implementation is already a good thing since I suspect the usual use case here is to send DTMF to some PSTN gateway, but not receive them.

AbdulBsit:
Yeah, I agree partial will work at this time, we can implement the event in C++ in future

src/RTCRtpTransceiver.ts
```diff
@@ -17,13 +16,13 @@ export default class RTCRtpTransceiver {
...
 
     constructor(args: {
-        peerConnectionId: number,
-        isStopped: boolean,
-        direction: string,
-        currentDirection: string,
-        mid?: string,
-        sender: RTCRtpSender,
-        receiver: RTCRtpReceiver,
+        peerConnectionId: number;
```
8BallBomBom, line 19: Unsure why you've changed so much formatting 🤔 this is no longer functional.

src/RTCRtpSender.ts
```diff
@@ -1,50 +1,60 @@
...
     _peerConnectionId: number;
     _rtpParameters: RTCRtpSendParameters;
+    _dtmf: RTCDTMFSender;
 
     constructor(info: {
-        peerConnectionId: number,
-        id: string,
-        track?: MediaStreamTrack,
-        rtpParameters: RTCRtpSendParametersInit
+        peerConnectionId: number;
```
8BallBomBom, line 23: Same issue here, no longer functional.

src/RTCRtpTransceiver.ts
AbdulBsit: Oo, this is formatted by my local prettier configuration, but it is functional on my end

src/RTCRtpSender.ts
AbdulBsit: 😕 It was working on my end, apart from formatting, I have added canInsertDtmf check before sending the dtmf on native side, can you please provide the logs

AbdulBsit: also can you please confirm as provided [here](https://www.w3.org/TR/webrtc/#caninsertdtmf-algorithm) all the points are valid in your case

8BallBomBom: Just realised, nope that is my bad, as that is a type definition you can indeed use `;` instead of `,`
Results of looking over code quick 😆 can mark them as resolved 👍🏻

AbdulBsit: Oo, no problem thanks

AbdulBsit:
Hey, any update?

saghul:
Sorry I've been quite busy this week and don't have the time to review this until the next one.

AbdulBsit:
Hey any update/possibility this weekend?

saghul:
Sorry, I've been quite busy this week. I'll try to get this in asap.

Can you please drop the changes to package-lock? They are unnecessary.

AbdulBsit:
Hey, i have already reverted the package lock json in the previous commit, i just copied the whole lock file from master and add it to the commit

saghul:
Something is still broken there. Please just completely undo the changes to that file.

It still shows thousands of lines changed.

AbdulBsit:
Yeah, This is exactly why its showing thousand lines change, first when i created the PR,  the lock file changes, after that you suggest me to revert it, I reverted it fully, and that's why it's showing the 10,200 line change, you can try the master one and my PR one, its exactly the same 10,200 line deletion and 10,200 addition which results in 0 change so that's why its showing the total change number 20,444 for lock file

src/RTCRtpSender.ts
```diff
@@ -1,50 +1,60 @@
...
         } catch (e) {
             return;
         }
 
         this._track = track;
     }
 
-    static getCapabilities(kind: 'audio' | 'video'): RTCRtpCapabilities {
-        if (kind === 'audio') {
+    static getCapabilities(kind: "audio" | "video"): RTCRtpCapabilities {
```
saghul, line 51: Please avoid all reformatting of existing code.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -607,6 +607,59 @@ public boolean peerConnectionRemoveTrack(int id, String senderId) {
...
+                PeerConnectionObserver peerConnectionObserver = mPeerConnectionObservers.get(peerConnectionId);
+                
+                if (peerConnectionObserver == null) {
+                    Log.d(TAG, "peerConnectioncanInsertDTMF() peerConnectionObserver is null");
+                    return false;
+                }
+                RtpSender audioSender = peerConnectionObserver.getSender(senderId);
+                
+                if (audioSender != null) {
+                    DtmfSender dtmfSender = audioSender.dtmf();
```
saghul, line 623: This is not guaranteed to be null, you need to check for it.

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -607,6 +607,59 @@ public boolean peerConnectionRemoveTrack(int id, String senderId) {
...
+        ThreadUtils.runOnExecutor(() -> {
+            try {
+                PeerConnectionObserver peerConnectionObserver = mPeerConnectionObservers.get(peerConnectionId);
+                if (peerConnectionObserver == null) {
+                    Log.d(TAG, "peerConnectionSendDTMF() peerConnectionObserver is null");
+                    return;
+                }
+                RtpSender audioSender = peerConnectionObserver.getSender(senderId);
+                if (audioSender != null) {
+                    DtmfSender dtmfSender = audioSender.dtmf();
```
saghul, line 647: I checked the native code. It can be null. Please add a check for it.

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -308,6 +308,62 @@ @implementation WebRTCModule (RTCPeerConnection)
...
+        audioSender = sender;
+        break;
+      } 
+    } 
+
+    if(audioSender){
+      ret = [audioSender.dtmfSender canInsertDtmf];
+    }
+
+    NSLog(@"peerConnectionCanInsertDTMF() audioSender is nil");
```
saghul, line 334: This line will be printed always because you don't return early above.

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -308,6 +308,62 @@ @implementation WebRTCModule (RTCPeerConnection)
                   }
                 }];
 }
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(peerConnectionCanInsertDTMF:peerConnectionId:(nonnull NSNumber *)peerConnectionId senderId:(nonnull NSString *)senderId)
+{ 
+__block BOOL ret = NO;
+dispatch_sync(self.workerQueue, ^{
+    RTCPeerConnection *peerConnection = self.peerConnections[peerConnectionId];
+    if (!peerConnection) {
+        NSLog(@"peerConnectionCanInsertDTMF() peerConnection is nil");
```
saghul, line 317: Please use the RN logger instead of NSLog.

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -308,6 +308,62 @@ @implementation WebRTCModule (RTCPeerConnection)
...
+    if([audioSender.dtmfSender canInsertDtmf]){
+      BOOL istoneplayed = [audioSender.dtmfSender insertDtmf:tone duration:duration interToneGap:interToneGap];
+      NSLog(@"DTMF Tone played :: [%s]", istoneplayed ? "true" : "false");
+      return;
+    }
+      NSLog(@"peerConnectionSendDTMF() canInsertDtmf is false");
+      return;
+  }
+  NSLog(@"peerConnectionSendDTMF() audioSender is nil");
+  return;
```
saghul, line 365: this return is not necessary

src/RTCDTMFSender.ts
```diff
@@ -0,0 +1,36 @@
...
+    insertDTMF(
+        tones: string,
+        duration = DEFAULT_DURATION,
+        interToneGap = DEFAULT_INTER_TONE_GAP
+    ) {
+        WebRTCModule.peerConnectionSendDTMF(
+            tones,
+            duration,
+            interToneGap < 30 ? 30 : interToneGap,
+            this._peerConnectionId,
```
saghul, line 32: Please pass the peerConnectionId and senderId first to the native function.

src/RTCDTMFSender.ts
```diff
@@ -0,0 +1,36 @@
...
+    _peerConnectionId: number;
+    _senderId: string;
+    constructor(peerConnectionId: number, senderId: string) {
+        this._peerConnectionId = peerConnectionId;
+        this._senderId = senderId;
+    }
+
+    get canInsertDTMF(): boolean {
+        // Blocking!
+        return WebRTCModule.peerConnectionCanInsertDTMF(
```
saghul, line 17: This is not a PC function but a sender one, can you please rename the native functions to senderCanInsertDTMF and senderInsertDTMF

src/RTCRtpSender.ts
```diff
@@ -1,50 +1,60 @@
-import { NativeModules } from 'react-native';
-
-import MediaStreamTrack from './MediaStreamTrack';
-import RTCRtpCapabilities, { senderCapabilities, DEFAULT_AUDIO_CAPABILITIES } from './RTCRtpCapabilities';
-import RTCRtpSendParameters, { RTCRtpSendParametersInit } from './RTCRtpSendParameters';
+import { NativeModules } from "react-native";
```
saghul, line 1: Please don't reformat the code to match your styling. Make your code conform to the surroundings.

src/RTCRtpTransceiver.ts
saghul: Yes, undo all formatting please.

src/RTCDTMFSender.ts
```diff
@@ -0,0 +1,36 @@
+import { NativeModules, Platform } from "react-native";
+
+const { WebRTCModule } = NativeModules;
+const DEFAULT_DURATION = 300;
+const DEFAULT_INTER_TONE_GAP = 70;
+
+export default class RTCDTMFSender {
```
saghul, line 7: I think it should be easy to implement toneBuffer (minus cancellation) by getting a callback from the native function when done.

Let's aim for the most complete API we can deliver.

saghul (changes requested):
Left a full review, please take a look!

8BallBomBom:
@AbdulBsit When you get a chance could you review the requested changes please.
Then we can look at merging 👍🏻

src/RTCDTMFSender.ts
AbdulBsit: Android (org.webrtc) don't provide as such callback, it does provide method to get bufferTones https://chromium.googlesource.com/external/webrtc/+/HEAD/sdk/android/api/org/webrtc/DtmfSender.java

8BallBomBom:
Was not expecting this to be revived, one of them things on the long back-burner list 😅 
Can you run the included format tool and ensure all lint errors are fixed.
Then we can review and finally merge.

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -309,6 +309,106 @@ @implementation WebRTCModule (RTCPeerConnection)
     [peerConnection addIceCandidate:candidate completionHandler:handler];
 }
 
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(senderCanInsertDTMF:(nonnull NSNumber *)peerConnectionId senderId:(nonnull NSString *)senderId) {
+    __block BOOL ret = NO;
+    dispatch_sync(self.workerQueue, ^{
+        RTCPeerConnection *peerConnection = self.peerConnections[peerConnectionId];
+        if(!peerConnection) {
+            RCTLogWarn(@"senderCanInsertDTMF() peerConnection is nil");
+            return;
```
saghul, line 318: Here and elsewhere in the function, return `ret`

ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
```diff
@@ -309,6 +309,106 @@ @implementation WebRTCModule (RTCPeerConnection)
...
+    return @(ret);
+}
+
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getDTMFToneBuffer:(nonnull NSNumber *)peerConnectionId senderId:(nonnull NSString *)senderId) {
+    __block NSString *toneBuffer = @"";
+    dispatch_sync(self.workerQueue, ^{
+        RTCPeerConnection *peerConnection = self.peerConnections[peerConnectionId];
+        if(!peerConnection) {
+            RCTLogWarn(@"getDTMFToneBuffer() peerConnection is nil");
+            return;
```
saghul, line 352: Here and elsewhere in the function, return the toneBuffer.

src/RTCDTMFSender.ts
saghul: 👍

src/RTCDTMFSender.ts
```diff
@@ -0,0 +1,55 @@
+import { NativeModules } from 'react-native';
+
+const { WebRTCModule } = NativeModules;
+const DEFAULT_DURATION = 300;
+const DEFAULT_INTER_TONE_GAP = 70;
+const MIN_DURATION = 70;
+const MAX_DURATION = 6000;
+const MIN_INTER_TONE_GAP = 50;
+
+export default class RTCDTMFSender {
```
saghul, line 10: Even though we don't implement the `tonechange` event, we should extend from event target here.

src/RTCDTMFSender.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    get toneBuffer(): string {
+        return WebRTCModule.getDTMFToneBuffer(
+            this._peerConnectionId,
+            this._senderId
+        );
+    }
+
+    insertDTMF(
+        tones: string,
```
saghul, line 33: Check if the tones are in the valid character set and throw an exception otherwise: https://developer.mozilla.org/en-US/docs/Web/API/RTCDTMFSender/insertDTMF#invalidcharactererror

src/RTCDTMFSender.ts
```diff
@@ -0,0 +1,55 @@
...
+        duration = DEFAULT_DURATION,
+        interToneGap = DEFAULT_INTER_TONE_GAP
+    ) {
+        if (duration < MIN_DURATION) {
```
saghul, line 37: I think you can rewrite this like so: `Math.min(Math.max(duration, MIN_DURATION), MAX_DURATION)`

src/RTCRtpSender.ts
```diff
@@ -22,6 +25,7 @@ export default class RTCRtpSender {
         this._peerConnectionId = info.peerConnectionId;
         this._id = info.id;
         this._rtpParameters = new RTCRtpSendParameters(info.rtpParameters);
+        this._dtmf = new RTCDTMFSender(info.peerConnectionId, info.id);
```
saghul, line 28: Don't initialize this eagerly, it should be `null` if the track doesn't do DTMF, like in a sender for video: https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/dtmf#value

saghul (changes requested):
Good to see this revived! Left some comments, PTAL! Please run the formatter too, while you're at it.

saghul:
> We would really appreciate if this gets merged and available in one of the future versions :)

Someone needs to complete the work. Would you like to pick this up?

8BallBomBom:
Can you fix the lint errors and also mark all the review notes which you've resolved as resolved.
Then another review can take place 👍🏻

ios/RCTWebRTC/WebRTCModule.m
```diff
@@ -50,7 +50,11 @@ - (instancetype)init {
         if (fieldTrials == nil) {
             // Fix for dual-sim connectivity:
             // https://bugs.chromium.org/p/webrtc/issues/detail?id=10966
-            fieldTrials = @{kRTCFieldTrialUseNWPathMonitor : kRTCFieldTrialEnabledValue};
+            fieldTrials = @{kRTCFieldTrialUseNWPathMonitor : kRTCFieldTrialEnabledValue,
+                // Initialize field trial for solving audio issues after hold when using CallKit.
+                // See: https://bugs.chromium.org/p/webrtc/issues/detail?id=8126#c35
+                @"WebRTC-Audio-iOS-Holding" : kRTCFieldTrialEnabledValue
```
saghul, line 56: Please undo this, it's unrelated and known to cause crashes.

saghul (commented):
Not sure what's up with it but please undo the package-lock changes.

ios/RCTWebRTC/WebRTCModule.m
AbdulBsit: Ooh, sorry i was making the patch for my use, forgot to switch the branch, it gets pushed in to this, I'll revert this, and as soon as a i get time, i will try completing this PR

saghul: Sounds good!

ghost:
Hi, just wondering — is this PR going to be merged, or is there anything left we can help finish to get it merged?

saghul:
There is unaddressed feeback, it's not in a mergeable state.

PR #1299, opened 2023-01-24
android: stop screen capture when media projection is stopped

davidliu:
`ScreenCapturerAndroid` currently doesn't stop when the MediaProjection is stopped (only stops when `stopCapture` is called). This stops the capturer when that happens so we can release any resources associated with capture.


saghul (approved):

saghul:
Good catch!

Issue #1310, opened 2023-02-02
Crash on Android  9 NoSuchMethodError on java/nio/FloatBuffer.position


pietgk:
#### Expected Behavior

After upgrading our app to the latest M106 version we have a working app on iOS and android 10 and higher.
We expect the App to work (not crash) on android 9.

#### Observed Behavior

On Android 9 we get a crash (the same as reported [here](https://community.jitsi.org/t/jitsi-android-sdk-crash/120099) )

```
NoSuchMethodError 
No virtual method position(I)Ljava/nio/FloatBuffer; in class Ljava/nio/FloatBuffer; or its super classes (declaration of 'java.nio.FloatBuffer' appears in /system/framework/core-oj.jar))
```

Stacktrace

```
org.webrtc.GlUtil in createFloatBuffer at line 47
org.webrtc.GlGenericDrawer in <clinit> at line 75
org.webrtc.YuvConverter in <init> at line 111
org.webrtc.YuvConverter in <init> at line 118
org.webrtc.SurfaceTextureHelper in create at line 92
com.oney.WebRTCModule.GetUserMediaImpl in createVideoTrack at line 376
com.oney.WebRTCModule.GetUserMediaImpl in getUserMedia at line 206
com.oney.WebRTCModule.WebRTCModule in lambda$getUserMedia$10$WebRTCModule at line 742
com.oney.WebRTCModule.-$$Lambda$WebRTCModule$4OOJLDjizOGGW7tkUrJ04dyPO_A in run at line 8
java.util.concurrent.ThreadPoolExecutor in runWorker at line 1167
java.util.concurrent.ThreadPoolExecutor$Worker in run at line 641
java.lang.Thread in run at line 764
```

From a more specific [similar Error](https://github.com/msgpack/msgpack-java/issues/516) explaining FloatBuffer change in java 8 and in java 11.

And verifying that the jar files in android/libs/ are compiled for JRE11 (Major version 55), 

We think that the library being compiled against the JRE 11 runtime and osVersion 9 seems to use the Java 8 context where the classByteBuffer inherits the position method from Buffer and has a return type of java.nio.Buffer. 
(On Java 11 the method is overridden with an implementation that returns java.nio.ByteBuffer.)

We do not understand why it works on osVersion higher than 9 (maybe /system/framework/core-oj.jar is the difference...).

We also looked further into 
- https://github.com/react-native-webrtc/react-native-webrtc/issues/1026
[...]

PR #1320, opened 2023-02-16
update android Faq with issue '#1310' info

pietgk:
Add knowledge from Crash on Android 9 NoSuchMethodError on java/nio/FloatBuffer.position into AndroidInstallation.md


8BallBomBom:
Definitely an interesting issue, never managed to reproduce with the latest React-Native but it would seem that due to upgrades of gradle over time the issue has gone away.

100% worth noting though, thanks for the contribution 👍🏻

saghul (approved):

saghul:
Cheers!

Issue #1309, opened 2023-02-01
Cannot display stream in Android occasionally


wjaykim:
#### Steps to reproduce the issue

I'm using Livekit solution, which uses react-native-webrtc for implementation of their React Native sdk.

1. Remote user starts screen sharing
2. Remote user disables screen sharing
3. Remote user enables screen sharing again

When I do the steps above, it fails to display remote user's screen share stream quietly. One log I found was:
```
No video stream for react tag: 0a2bd99d-453e-4db5-8fad-837bc5f1e7e3
```

So I dug more deeply, then I found that `streamReactTag` value maybe changed without `streamId` is changed. It prevents `streamReactTag` to update in this line: https://github.com/react-native-webrtc/react-native-webrtc/blob/002e871853a8fbd6e05c8260b71de5af974ba761/src/RTCPeerConnection.ts#L248-L256

So in order to temporarily fix this issue, I added
```ts
                if (stream) {
                  stream._reactTag = streamInfo.streamReactTag;
                }
```
next to this line: https://github.com/react-native-webrtc/react-native-webrtc/blob/002e871853a8fbd6e05c8260b71de5af974ba761/src/RTCPeerConnection.ts#L258

Then remote screen share stream is displayed successfully. I'm not sure if the fix is the right way, because in ios, remote stream is displayed without the fix with the same `streamId` and different `streamReactTag`.

Needing for help!

#### Platform Information

* **React Native Version**: 0.71.1
* **Plugin Version**: 106.0.4
* **OS**: Android
* **OS Version**:

fukemy:
I have same problem, instead of create new videoTrack, I just use RTCRtpSender to replace track, then it's work on everycase

streamer45:
Not 100% sure if related to the OP's issue but I can reproduce something very similar on our side as well when switching (removing/adding) between simulcasted video tracks:

```
 LOG  rn-webrtc:pc:DEBUG 6 ctor +1m
 LOG  rn-webrtc:pc:DEBUG 6 setLocalDescription +207ms
 LOG  rn-webrtc:pc:DEBUG 6 setLocalDescription OK +26ms
[...]

PR #1326, opened 2023-02-20
android,ios: fix matching remote streams in onTrack

saghul:
Fixes: https://github.com/react-native-webrtc/react-native-webrtc/issues/1309

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -49,7 +49,8 @@ class PeerConnectionObserver implements PeerConnection.Observer {
     private int transceiverNextId = 0;
 
     private PeerConnection peerConnection;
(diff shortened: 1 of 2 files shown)


Romick2005:
It doesn't help for my case. The issue exist when connection is made between Android -> Android ( no remote video for the caller, but both video local and remote works for the callee). For iOS->Android, Android->iOS, iOS->iOS it works fine.

With this fix both Android devices does not show remote video. In previous version only caller does not have remote video.

But at least now it is consistent ).

saghul:
Do you see a new stream in ontrack, with a different react tag?

streamer45:
Gave this a go and cannot reproduce the issue any longer. Incoming streams/tracks with previously seen ids are now played as expected.

saghul:
> Gave this a go and cannot reproduce the issue any longer. Incoming streams/tracks with previously seen ids are now played as expected.

Thanks for confirming! @Romick2005 i suspect yours is a different issue.

Issue #1347, opened 2023-03-18
MediaStream addTrack select wrong track id when remote tracks has the duplicated id


aitboudad:
Hi, 

I've two PeerConnection instances that contains a video track with the same `id=janus`, when I create a new `MediaStream` from the second `PeerConnection` instance, the selected video track is always retrieved the first PeerConnection instance.

```
const pc1 = new RTCPeerConnection();
....

const pc2 = new RTCPeerConnection();
pc2.addEventListener( 'track', event => {
  remoteMediaStream = new MediaStream();
  remoteMediaStream.addTrack(event.track);
} );
...

<RTCView
  mirror={true}
  objectFit={'cover'}
  streamURL={remoteMediaStream.toURL()}
  zOrder={0}
/>
```

Solution:

The solution is to pass the peerConnection id to [mediaStreamAddTrack](https://github.com/react-native-webrtc/react-native-webrtc/blob/master/src/MediaStream.ts#L86) in order to getTrack from the right peer connection [getTrack](https://github.com/react-native-webrtc/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java#L412-L418)

saghul:
Can you log the track objects?

They should have different reactTag values.

Adding the PC ID is not a valid solution because you can add a track that doesn't belong to a PC to a stream.

aitboudad:
> Can you log the track objects?

Peer instance 1 Track (console `e.track`):
```json
{
  "_constraints":{},
  "_enabled":true,
  "_muted":false,
  "_peerConnectionId":1,
  "_readyState":"live",
  "_settings":{},
  "id":"janus0",
  "kind":"video",
  "label":"",
  "remote":true
}
```

Peer instance 2 Track (console `e.track`):
```json
{
  "_constraints":{},
  "_enabled":true,
  "_muted":false,
  "_peerConnectionId":2,
  "_readyState":"live",
  "_settings":{},
  "id":"janus0",
  "kind":"video",
  "label":"",
  "remote":true
}
```


> They should have different reactTag values.

yes, they have a difference `reactTag`

> Adding the PC ID is not a valid solution because you can add a track that doesn't belong to a PC to a stream.

I see, then I think rely on an unique internal id would solve the issue

aitboudad:
[...]

PR #1350, opened 2023-03-20
streams: fix matching remote tracks

saghul:
We generate random UUIs for our local tracks, but collisions are possible for remote tracks. The unicity of tracks IDs is guaranteed only in the scope of a single PeerConnection.

Thus, use the PeerConnection internal ID when we go looking for a track. For local tracks we'll use an ID of -1.

Fixes: https://github.com/react-native-webrtc/react-native-webrtc/issues/1347


8BallBomBom:
Will run some tests in a little while 👍🏻

aitboudad:
I confirm that the issue has been fixed, tested on Android.
Thank you for the effort👍🏻!

Issue #1346, opened 2023-03-17
Type Script : Getting error when calling the "await mediaDevices.getUserMedia({ audio:true, video: true });"


VikramSingh1994:
**@saghul** , please guide.
I have created new project using react native version "0.71.4" and typescripts (App.tsx).
I have install  "react-native-webrtc": "^106.0.7".
then
**import { mediaDevices, RTCView } from 'react-native-webrtc';**
and call the method -
**await mediaDevices.getUserMedia({ audio:true, video: true });**
then getting error **in iOS** (But working in Android):-

 **ERROR  Invariant Violation: `new NativeEventEmitter()` requires a non-null argument**., js engine: hermes
 **ERROR  [TypeError: Cannot read property 'mediaDevices' of undefined]**

#### Expected Behavior
It should be work like js based react native project.

#### Observed Behavior
react-native-webrtc not work in type scripts based project.

#### Platform Information
**iOS 16**

"dependencies": {
    "react": "18.2.0",
    "react-native": "0.71.4",
    "react-native-webrtc": "^106.0.7"
  },
  "devDependencies": {
    "@babel/core": "^7.20.0",
    "@babel/preset-env": "^7.20.0",
    "@babel/runtime": "^7.20.0",
    "@react-native-community/eslint-config": "^3.2.0",
    "@tsconfig/react-native": "^2.0.2",
    "@types/jest": "^29.2.1",
    "@types/react": "^18.0.24",
    "@types/react-test-renderer": "^18.0.0",
    "babel-jest": "^29.2.1",
    "eslint": "^8.19.0",
    "jest": "^29.2.1",
    "metro-react-native-babel-preset": "0.73.8",
    "prettier": "^2.4.1",
    "react-test-renderer": "18.2.0",
    "typescript": "4.8.4"
  },
  "jest": {
    "preset": "react-native"
  }

saghul:
Did you call registerGlobals() at the start of your app?

8BallBomBom:
Can't say I've experienced this myself during testing but after looking around 😅
It would appear that if you don't run `pod install` in your project folder then some things won't link up properly.
As a result you'll get a whole range of random errors.

saghul:
> After some extra testing on iOS it would indeed seem to be reproducible.
> Only takes creating a new react-native project, adding the webrtc module and then forgetting to run pod install.

[...]

PR #1352, opened 2023-03-20
Quick null check for both ios and android.

8BallBomBom:
Won't fix [this](https://github.com/react-native-webrtc/react-native-webrtc/issues/1346) issue but will make developers aware that they will need to run pod install or npm install as it would appear some linking commands don't always execute properly when adding packages to your project.

Not actually sure if that is intentional or not.
Also would it be worth mentioning anything about alternative package managers like yarn?

### DIFF
--- a/src/RTCRtpTransceiver.ts
+++ b/src/RTCRtpTransceiver.ts
@@ -5,7 +5,6 @@ import RTCRtpSender from './RTCRtpSender';
 
 const { WebRTCModule } = NativeModules;
 
-
 export default class RTCRtpTransceiver {
     _peerConnectionId: number;
     _sender: RTCRtpSender;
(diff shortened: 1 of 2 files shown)


saghul (approved):

8BallBomBom:
Ahhhh, not very often i do that but will instead do drafts then squash the lot once ready 👍🏻

saghul (approved):

8BallBomBom:
Don't think i actually have them permissions 🤔

8BallBomBom:
Saved my bacon, thanks 😅

saghul:
> Don't think i actually have them permissions 🤔

Well, let me fix that then!

PR #1372, opened 2023-05-01
android: bump API level to 24, aka Android 7

saghul:
We have already de-facto done it since some of the code after the migration to UP requires API level 24.

### DIFF
--- a/Documentation/AndroidInstallation.md
+++ b/Documentation/AndroidInstallation.md
@@ -4,6 +4,8 @@ Starting with React Native 0.60 due to a new auto linking feature you no longer 
 
 See a sample app in the `examples/GumTestApp` directory.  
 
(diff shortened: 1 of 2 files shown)


bfelbo:
This is great to know, thanks! Is the last release to confidently support older Android versions then [1.106.1](https://github.com/react-native-webrtc/react-native-webrtc/releases/tag/1.106.1)?

saghul:
If by old you mean plan B, then 1.XXX will do yeah. The migration to Unified Plan introduced the API level 24 requirement, but we missed to bump it until now.

bfelbo:
I mean Android 5 and Android 6 as we have users on those versions and haven't yet migrated to UP. Sounds good, thanks for clarifying!

Issue #1353, opened 2023-03-21
Error when calling from iOS 16.4 to Android 13


w4ugit:
I am making an app to make calls between two devices. I use p2p. When I make a call from Android, the call is accepted and everything works. When I make a call from iOS - the call is accepted, but we do not see each other, only our streams. I don't see any errors in the console.

But when I do exactly that with Android 12 - everything works. I also want to note that everything works from iOS to iOS. 

First device: Iphone 8 (iOS 16.4)
Two device: Android Google Pixel 4XL (Android 13)

System:
    OS: macOS 13.3
    CPU: (8) arm64 Apple M2
    Memory: 93.80 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 19.7.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.5.0 - /opt/homebrew/bin/npm
    Watchman: 2023.03.06.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.2, iOS 16.2, macOS 13.1, tvOS 16.1, watchOS 9.1
    Android SDK: Not Found
  IDEs:
    Android Studio: 2022.1 AI-221.6008.13.2211.9619390
    Xcode: 14.2/14C18 - /usr/bin/xcodebuild
  Languages:
    Java: javac 13 - /usr/bin/javac

saghul:
What version of this plugin are you using?

w4ugit:
"react-native-webrtc": "^106.0.7"

w4ugit:
I also tested it on Oppo, also with Android 13 - it works

saghul:
Can you try making this change manually and see if things change? https://github.com/react-native-webrtc/react-native-webrtc/issues/1332#issuecomment-1448246584

w4ugit:
This should be done in the file?
```
/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```

saghul:
Correct.

w4ugit:
did not help, the problem remained

w4ugit:
I can provide it tomorrow

w4ugit:
[...]

PR #1375, opened 2023-05-11
api: refactor nonstandard createOffer options

saghul:
~~They don't work properly with transceivers and they are deprecated anyway.~~

Keep only the legacy ones mentioned in 4.4.3.2 but implement them in JS
since the native ones seem to not function correctly in certain
circumstances.

Closes: https://github.com/react-native-webrtc/react-native-webrtc/issues/1353

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -871,7 +871,7 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
-    public void peerConnectionCreateOffer(int id, ReadableMap options, Promise promise) {
+    public void peerConnectionCreateOffer(int id, Promise promise) {
         ThreadUtils.runOnExecutor(() -> {
             PeerConnection peerConnection = getPeerConnection(id);
 
@@ -910,7 +910,7 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
                 public void onSetSuccess() {}
             };
 
-            peerConnection.createOffer(observer, constraintsForOptions(options));
+            peerConnection.createOffer(observer, new MediaConstraints());
         });
     }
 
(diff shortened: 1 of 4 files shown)


davidliu:
`iceRestart` is still a valid option to be passed in here, so I don't think we can remove this entirely.

8BallBomBom:
@davidliu Can't you call the `restartIce` method on the peer connection which then forces the next created offer to perform an ice restart?

Reference over [here](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/restartIce).
But even then we'd be going against spec as the `iceRestart` option for offer creation hasn't been deprecated.

saghul:
I think I'll need to abandon this because we also depend on the deprecated options 🤦 

For restartIce, however, you can call the restartIce method on the PC and then createOffer.

8BallBomBom:
Possible re-visit for the future as spec changes might occur? 🤔

saghul:
> Possible re-visit for the future as spec changes might occur? 

It might still be A Good Thing (TM) to drop the native bits and do addTransceiver or call restartIce in JS, to simplify the current implementation and fix the offerToReceiveXXX not working.

saghul:
Alright, I gave it another go, please take another look!

@davidliu Ah my bad, it's on another section of the spec so I missed it, I'll bring it back. Can you check if calling `pc.restartIce()` works for you? Then I could make it that way instead of the native cruft that was there before for constraints. Spec wise they seem to be equivalent.

davidliu:
Calling `pc.restartIce` right before creating the offer does work for me. 

Spec wise, I think the difference is that the the `negotiationNeeded` flag gets updated with `restartIce`, and can generate a `negotiationneeded` event.

I do think it's safer to just pass the map down and avoid handling it ourselves. [Looks like the legacy options are still handled within WebRTC](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/pc/sdp_offer_answer.cc;l=2402?q=sdp_offer_answer.cc&ss=chromium), though they only [look for the capital **O** options](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/sdk/media_constraints.cc;l=107?q=media_constraints.cc&ss=chromium). Perhaps better to just remap the lowercase options to the uppercase version.

saghul:
Turns out the problem was that we were not mapping the new transceivers into JS objects. https://github.com/react-native-webrtc/react-native-webrtc/pull/1411 solves it.

PR #1411, opened 2023-06-30
pc: fix handling legacy options in createOffer

saghul:
They may create new transceivers, so make sure we create the corresponding JS objects.

Fixes: https://github.com/react-native-webrtc/react-native-webrtc/issues/1353
Ref: https://github.com/react-native-webrtc/react-native-webrtc/pull/1375


saghul:
@jallamsetty1 we need this for JM or receiving video in P2P fails unless we have it on, because the legacy createOffer options didn't work, can you PTAL?

Issue #1033, opened 2021-07-05
MacOS Catalyst Support


ajnozari:
#### Expected behavior
Building for macOS Big Sur or higher should be possible.

#### Observerd behavior
XCFramework is not detected and loaded for catalyst.  This does not affect iOS simulators

#### Steps to reproduce the problem
install the package and attempt to build with the target set to macOS.

#### Platform information
The xcframework needs to be told to link them for macOS.  In fact it should be able to use the same framework used for the simulator.  Ninja can even do this, but I'm not sure how to add macOS as a target to the xcFramework build script that is used.  I've tried several times to get it to build but I'm just not familiar enough with Ninja and how your build script is setup.

* **React Native version**: 0.64.2
* **Plugin version**: 0.89.3
* **OS**: MacOS
* **OS version**: 10.15+

saghul:
PRs are welcome, it’s unlikely I’ll have the time to work on this myself.

ajnozari:
So I was able to get ninja to start generating the macos catalyst slice but it failed halfway.  

I’m gonna give the errors a poke over time and see how far along I can push the build, however it seems to be an OpenGL function call which is deprecated and won’t run on catalyst, which might end this attempt until upstream webrtc themselves remove any OpenGL code.

I’ll post the exact error later when I can get it to generate the build error again. I know it’s been mentioned in other issues that this webRTC does use OpenGL so I’m wondering if maybe that means ninja is trying to swap something out on the fly and it’s not using Metal like apple now expects.

saghul:
I just saw that Signal has contributed patches for building for Catalyst to WebRTC and they landed in master already.

There is a mention about some problems on x64 but arm64 seems to work fine.

I hope this matures enough in the next 1-2 WebRTC releases so we can pick it up.

Flam3rboy:
Any updates on this?

saghul:
Not really, sorry, I didn't get the time to investigate.

Flam3rboy:
If this doesn't work, I'll have to use a fallback option. 
@saghul Do you know if it works if I embed a react-native-WebView and use the system browser WebRTC capabilities?

saghul:
I don't know. You'd have to ask the WebView maintainers.

Flam3rboy:
still searching for solutions:
```
'WebRTC/RTCCameraVideoCapturer.h' file not found
```

Flam3rboy:
[...]

PR #1420, opened 2023-07-12
add catalyst support to build script

SamuelScheit:
As mentioned in #1033, chromium's webrtc supports catalyst as a target, which has now been added to the build webrtc script.
Also the ios_deployment_target is set to 14.0 for and only for catalyst targets. (remains 12.0 for ios)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,4 @@ examples/GumTestApp_macOS/package-lock.json
 *.zip
 lib/
 src/*.js
-
+build_webrtc/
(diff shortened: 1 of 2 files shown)


saghul:
Is that all that's needed for building the right XCFramework slice? Nice!

What webrtc version did this start working on?

SamuelScheit:
I did not test other versions, but it works with the latest version of the webrtc repo

saghul:
Ok. I'll give it a try when we bump to 114 or 115.

8BallBomBom:
@saghul Possible to be included with the M124 update?

saghul:
If I need to do another build I'll include it.

I forgot about this PR, sorry!

PR #1426, opened 2023-07-20
Add pending and current descriptions

shepeliev:
This PR fixes getting `RTCPeerConnection.localDescription` and `RTCPeerConnection.remoteDescription` according to WebRTC docs:

> On a more fundamental level, the returned value is the value of [RTCPeerConnection.pendingLocalDescription](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/pendingLocalDescription) if that property isn't null; otherwise, the value of [RTCPeerConnection.currentLocalDescription](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/currentLocalDescription) is returned. See [Pending and current descriptions](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Connectivity#pending_and_current_descriptions) in the WebRTC Connectivity page for details on this algorithm and why it's used.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -502,9 +502,27 @@ class PeerConnectionObserver implements PeerConnection.Observer {
     @Override
     public void onSignalingChange(PeerConnection.SignalingState signalingState) {
         ThreadUtils.runOnExecutor(() -> {
+            WritableMap newLocalSdpMap = Arguments.createMap();
+            WritableMap newRemoteSdpMap = Arguments.createMap();
+            SessionDescription newLocalSdp = peerConnection.getLocalDescription();
+            SessionDescription newRemoteSdp = peerConnection.getRemoteDescription();
+
+            if (newLocalSdp != null) {
+                newLocalSdpMap.putString("type", newLocalSdp.type.canonicalForm());
+                newLocalSdpMap.putString("sdp", newLocalSdp.description);
+            }
+
+            if (newRemoteSdp != null) {
+                newRemoteSdpMap.putString("type", newRemoteSdp.type.canonicalForm());
+                newRemoteSdpMap.putString("sdp", newRemoteSdp.description);
+            }
+
             WritableMap params = Arguments.createMap();
             params.putInt("pcId", id);
             params.putString("signalingState", signalingStateString(signalingState));
+            params.putMap("localSdp", newLocalSdpMap);
+            params.putMap("remoteSdp", newRemoteSdpMap);
+
             webRTCModule.sendEvent("peerConnectionSignalingStateChanged", params);
         });
     }
--- a/src/RTCPeerConnection.ts
+++ b/src/RTCPeerConnection.ts
@@ -63,18 +63,6 @@ const PEER_CONNECTION_EVENTS = [
 let nextPeerConnectionId = 0;
 
 export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_CONNECTION_EVENTS) {
-    currentLocalDescription: RTCSessionDescription | null = null;
-    pendingLocalDescription: RTCSessionDescription | null = null;
-    get localDescription(): RTCSessionDescription | null {
-        return this.pendingLocalDescription || this.currentLocalDescription;
-    };
-    
-    currentRemoteDescription: RTCSessionDescription | null = null;
-    pendingRemoteDescription: RTCSessionDescription | null = null;
-    get remoteDescription(): RTCSessionDescription | null {
-        return this.pendingRemoteDescription || this.currentRemoteDescription;
-    };
-
     signalingState: RTCSignalingState = 'stable';
     iceGatheringState: RTCIceGatheringState = 'new';
     connectionState: RTCPeerConnectionState = 'new';
@@ -84,6 +72,11 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
     _transceivers: { order: number, transceiver: RTCRtpTransceiver }[];
     _remoteStreams: Map<string, MediaStream>;
     _pendingTrackEvents: any[];
+
+    _currentLocalDescription: RTCSessionDescription | null = null;
+    _pendingLocalDescription: RTCSessionDescription | null = null;
+    _currentRemoteDescription: RTCSessionDescription | null = null;
+    _pendingRemoteDescription: RTCSessionDescription | null = null;
 
     constructor(configuration) {
         super();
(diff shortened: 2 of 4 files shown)


8BallBomBom:
LGTM 👍🏻
Welcome and Thanks for the contribution.

src/RTCPeerConnection.ts
```diff
@@ -157,15 +166,23 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
             desc = null;
         }
 
+        this.pendingLocalDescription = new RTCSessionDescription(desc);
```
saghul, line 169: LEt's store this in some `_pendingLocalDescription` and add a getter, so it's not settable.

src/RTCPeerConnection.ts
```diff
@@ -157,15 +166,23 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+
+        let localDescriptionResult;
+        try {
+            localDescriptionResult = await WebRTCModule.peerConnectionSetLocalDescription(this._pcId, desc);
+        } finally {
+            this.pendingLocalDescription = null;
```
saghul, line 175: AFAIS the current one is still there after sertting it:

Screen Shot 2023-07-20 at 16 56 58

src/RTCPeerConnection.ts
```diff
@@ -157,15 +166,23 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+        }
         const {
             sdpInfo,
             transceiversInfo
-        } = await WebRTCModule.peerConnectionSetLocalDescription(this._pcId, desc);
+        } = localDescriptionResult;
 
         if (sdpInfo.type && sdpInfo.sdp) {
-            this.localDescription = new RTCSessionDescription(sdpInfo);
+            this.currentLocalDescription = new RTCSessionDescription(sdpInfo);
```
saghul, line 183: Ditto, let's make this private and add a getter.

src/RTCPeerConnection.ts
```diff
@@ -63,8 +63,17 @@ const PEER_CONNECTION_EVENTS = [
 let nextPeerConnectionId = 0;
 
 export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_CONNECTION_EVENTS) {
-    localDescription: RTCSessionDescription | null = null;
-    remoteDescription: RTCSessionDescription | null = null;
+    currentLocalDescription: RTCSessionDescription | null = null;
+    pendingLocalDescription: RTCSessionDescription | null = null;
+    get localDescription(): RTCSessionDescription | null {
```
saghul, line 68: Please move these to after the constructor.

saghul (changes requested):
Thanks for contributing! Left some comments.

src/RTCPeerConnection.ts
saghul: I think we need to reset them when the signalling state goes back to stable. While in 'have-local-offer' it should still be pending.

shepeliev: Yep, you're right. I'll rework it.

saghul: Please check the spec, the are probably implications when a rollback happens.

src/RTCPeerConnection.ts
```diff
@@ -569,6 +615,42 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+
+            if (localSdp.type && localSdp.sdp) {
+                switch (this.signalingState) {
+                    case 'stable':
+                        this._currentLocalDescription = new RTCSessionDescription(localSdp);
+                        this._pendingLocalDescription = null;
+                        break;
+                    case 'have-local-offer':
+                    case 'have-local-pranswer':
+                        this._currentLocalDescription = null;
```
saghul, line 628: Isn't this block redundant with the handling in setLocalDescription?

src/RTCPeerConnection.ts
```diff
@@ -569,6 +615,42 @@ export default class RTCPeerConnection extends defineCustomEventTarget(...PEER_C
...
+
+            if (remoteSdp.type && remoteSdp.sdp) {
+                switch (this.signalingState) {
+                    case 'stable':
+                        this._currentRemoteDescription = new RTCSessionDescription(remoteSdp);
+                        this._pendingRemoteDescription = null;
+                        break;
+                    case 'have-remote-offer':
+                    case 'have-remote-pranswer':
+                        this._currentRemoteDescription = null;
```
saghul, line 645: Ditto but for sRD.

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -502,9 +502,27 @@ public void onRenegotiationNeeded() {
...
+
+            if (newRemoteSdp != null) {
+                newRemoteSdpMap.putString("type", newRemoteSdp.type.canonicalForm());
+                newRemoteSdpMap.putString("sdp", newRemoteSdp.description);
+            }
+
             WritableMap params = Arguments.createMap();
             params.putInt("pcId", id);
             params.putString("signalingState", signalingStateString(signalingState));
+            params.putMap("localSdp", newLocalSdpMap);
```
saghul, line 523: See below, I wonder if we only want to do this for the transition to "stable".

saghul (commented):
Left a couple of comments, this is looking good!

src/RTCPeerConnection.ts
shepeliev: I’d rather say handling it in `setLocalDescription` is redundant, but I’ve left it there just in case SDP updated while 

```js
await WebRTCModule.peerConnectionSetLocalDescription(this._pcId, desc);
```

resolved.

Assume we call `setLocalDescription(offer)`. The fact is that we receive `onsignalingstatechange(have-local-offer)` event earlier than above promise resolved. So if we don't set current/pending SDP here, in client's JS code `peerConnection.pendingLocalDescription` would return `null`. 

Of course, we could handle all this stuff in `set(Local|Remote)Description`, but it would make us wrap 
```js
await WebRTCModule.peerConnectionSetLocalDescription(this._pcId, desc);
```
in try catch and restore SDPs in case of exception. I think in current way code looks more concise.

saghul: Makes sense, thanks for th explanation!

In that case let's remove the handling here so it's centralized in the other place?

shepeliev: Do you mean remove from `peerConnectionSignalingStateChanged` listener and move everything to the `set(Local|Remote)Description`?

saghul: The opposite, as you suggested

saghul:
Thing look very good, great work @shepeliev ! I'll take it for a spin with Jitsi Meet to see we have no regressions before merging. I'll be off this week so it will need to wait until the next one, sorry!

saghul:
Excellent work mate!

8BallBomBom:
It has been a pleasure, always great to see things lining up with the spec 👍🏻

Issue #1369, opened 2023-04-26
Audio constraints support for iOS


aonnikov:
Is that possible to support audio constraints for iOS? It appears that audio constraints are completely ignored, but I suppose that it is due to the fact, that AGC, AEC, and NS are disabled by default in favor of internal (platform specific) implementations.

I'm interested in disabling any processing on iOS. I guess that this flag is responsible for that, but there is no way to override AudioDeviceModule (similar to Android).
https://github.com/jitsi/webrtc/blob/v111.0.1/sdk/objc/native/api/audio_device_module.mm#L20

Do you have any suggestions?

saghul:
Hey there! I think this might be possible with some APIs that landed not too long ago.

https://github.com/jitsi/webrtc/blob/7a6c522ecf9e0a53b5fbbe5ef4a82c407c060b35/sdk/objc/api/peerconnection/RTCPeerConnectionFactory.h#L52
https://github.com/jitsi/webrtc/blob/M111/sdk/objc/components/audio/RTCAudioDevice.h

A patch adding the ability to pass a custom RTCAudioDevice to WebRTCOptions and then sets it on the PC factry would be acceptable.

saghul:
.

PR #1431, opened 2023-08-02
ios: add ability to customize RTCAudioDevice

saghul:
Fixes: https://github.com/react-native-webrtc/react-native-webrtc/issues/1369


8BallBomBom:
Ahhhh definitely nice to see 👍🏻
Brings things closer to Android.
[WebRTCModule.java - L60](https://github.com/react-native-webrtc/react-native-webrtc/blob/c3effb3a9caa0ac2a3ff33ba8cb08959e63692be/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java#L60)

PR #1434, opened 2023-08-10
[plan-b] Add releaseWebrtcModule() global function

kjvenalainen:
Adds `releaseWebrtcModule()` as a global function.

When making subsequent calls, it was observed that the internal WebRTC audio processing module, and associated dependencies were reused across calls. This does not match the behaviour of Chromium, where the WebRTC dependencies are properly torn down after all calls.

Additionally, dependencies remained resident in memory even when the app was idle and backgrounded, which is not ideal from a memory footprint point of view.

PS: This PR is opened against `plan-b` branch precisely because that branch is based on M106 which is where these changes were originally developed and tested.

### DIFF
--- a/Documentation/BasicUsage.md
+++ b/Documentation/BasicUsage.md
@@ -15,7 +15,8 @@ import {
 	MediaStream,
 	MediaStreamTrack,
 	mediaDevices,
-	registerGlobals
+	registerGlobals,
+	releaseWebrtcModule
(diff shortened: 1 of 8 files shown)


saghul:
We won't be making any changes to the plan-b branch.

There won't be any new plan-b release either.

I suggest you implement these on a local fork.

PR #1443, opened 2023-09-01
Improve typing of RTCIceCandidate and MediaStream

Thaigun:
The type of the constructor parameter is currently 
```
{  candidate?: string, sdpMLineIndex?: null, sdpMid?: null }
```
which causes TypeScript errors despite valid parameters like

```
new RTCIceCandidate({candidate: 'candidateString', sdpMLineIndex: 0, sdpMid: 'sdpMidString'});
```

This fix allows constructing the class without TypeScript issues.

As far as I can tell, shouldn't cause issues to any existing implementations.

EDIT:
Added typing for MediaStream constructor, too.

### DIFF
--- a/src/RTCIceCandidate.ts
+++ b/src/RTCIceCandidate.ts
@@ -1,14 +1,21 @@
+interface RTCIceCandidateInfo {
+    candidate?: string;
+    sdpMLineIndex: number | null;
+    sdpMid: string | null;
+}
+
 export default class RTCIceCandidate {
     candidate: string;
     sdpMLineIndex: number;
     sdpMid: string;
 
-    constructor({ candidate = '', sdpMLineIndex = null, sdpMid = null }) {
+    constructor(info: RTCIceCandidateInfo) {
+        const { candidate, sdpMLineIndex, sdpMid } = info;
         if (sdpMLineIndex === null || sdpMid === null) {
             throw new TypeError('`sdpMLineIndex` and `sdpMid` must not null');
         }
 
-        this.candidate = candidate;
+        this.candidate = candidate ?? '';
         this.sdpMLineIndex = sdpMLineIndex;
         this.sdpMid = sdpMid;
     }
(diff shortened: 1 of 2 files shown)


src/RTCIceCandidate.ts
```diff
@@ -1,14 +1,22 @@
...
-        if (sdpMLineIndex === null || sdpMid === null) {
+    constructor(info: RTCIceCandidateInfo) {
+        const { candidate, sdpMLineIndex, sdpMid } = info;
+        // == null checks for both null and undefined, but not for other falsy values
+        if (sdpMLineIndex == null || sdpMid == null) {
             throw new TypeError('`sdpMLineIndex` and `sdpMid` must not null');
         }
 
-        this.candidate = candidate;
+        this.candidate = candidate ?? '';
```
saghul, line 19: Is this necessary? Doesn't the default value above cover it?

saghul (approved):
LGTM with a question.

src/RTCIceCandidate.ts
Thaigun: There's no default value anymore. 

Please note that I added an other typing to this PR, even though it doesn't match the branch name etc...

8BallBomBom:
Welcome and Thanks for contributing 👍🏻 
Happy to see TS improvements, plenty to fix up to a more strict standard.
Brief check but LGTM 😃

Thaigun:
Thank you for reacting swiftly, it definitely encourages me to make more contributions as I run to more problems in my implementation.

8BallBomBom:
Couple of lint related errors, otherwise good to go 👍🏻

Thaigun:
Ah, linter didn't like my solution to check for `undefined` or `null` with `==`. Now looks more like how it used to be with default values in the deconstruction.

saghul:
FYI you can opt out of the linter if necessary with eslint-disable-line if warranted.

8BallBomBom:
Gonna push this through, plenty more to clean up but we will get there bit by bit 👍🏻

PR #1456, opened 2023-10-10
fix(android): release JavaAudioDeviceModule via onCatalystInstanceDestroy method

wjaykim:
When WebRTCModule is initialized twice, app crashes.
To reproduce:
1. before WebRTCModule is initialized(such as `Application.onCreate()`), inject new `audioDeviceModule` created by `JavaAudioDeviceModule.builder` in `WebRTCModuleOptions.getInstance()`.
2. Apply Codepush update, which triggers React Context to be recreated.
3. WebRTCModule is initialized again, then it crashes when executing `createPeerConnectionFactory` (https://github.com/react-native-webrtc/react-native-webrtc/blob/master/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java#L97)

I think its because `nativeAudioDeviceModule` cannot be created twice.

If I release existing audio device module before executing `createPeerConnectionFactory`, the app does not crash.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -42,6 +42,7 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
     PeerConnectionFactory mFactory;
     VideoEncoderFactory mVideoEncoderFactory;
     VideoDecoderFactory mVideoDecoderFactory;
+    AudioDeviceModule mAudioDeviceModule;
 
     // Need to expose the peer connection codec factories here to get capabilities
     private final SparseArray<PeerConnectionObserver> mPeerConnectionObservers;
@@ -103,6 +104,7 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
         // Saving the encoder and decoder factories to get codec info later when needed.
         mVideoEncoderFactory = encoderFactory;
         mVideoDecoderFactory = decoderFactory;
+        mAudioDeviceModule = adm;
 
         getUserMediaImpl = new GetUserMediaImpl(this, reactContext);
     }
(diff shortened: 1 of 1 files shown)


saghul:
Isn't there a lifecycle event / function called when the module is destroyed instead?

saghul:
This should be done in `onCatalystInstanceDestroy`.

wjaykim:
Yes, it could be, but what if audioDeviceModule was created where regardless of rn native module context?

I thought if unreleased audioDeviceModule in shared WebRTCModuleOptions makes app to destroy, it’s safer to release it before creating peerconnection by rn webrtc side.

saghul:
Well, we don't know when and where the ADM was created. But since it's used for RN, what would the purpose of that ADM be besides RN?

In general we should release all resources when Catalyst is destroyed, which we are not doing today.

If someone has a very specific use case that warrants not doing that we could add a boolean that indicates if the ADM should be released or not. Or maybe we could take an `onInitialize` callback which the module will call when it gets initialized, so the app can do its thing.

I think we have many options, but the current one looks very odd.

wjaykim:
I see. Then would it be better if we call release in WebRTCModule’s `onCatalystInstanceDestroy`?

8BallBomBom:
Releasing everything via the `onCatalystInstanceDestroy` method for Android and the `dealloc` or `invalidate` method for iOS would probably be the best route to ensure everything gets cleaned up on reload 🤔

saghul:
> I see. Then would it be better if we call release in WebRTCModule’s `onCatalystInstanceDestroy`?

Yes indeed.

wjaykim:
IDK what resources need to be released, so I've made change releasing ADM only in `onCatalystInstanceDestroy`. Please take a look🙃

saghul (approved):

8BallBomBom:
LGTM 👍🏻 

We can always do another PR to include everything else.
As for what needs to be released 🤔 peer connections, streams and camera as far as i'm aware.

PR #1457, opened 2023-10-10
Revert "pc,api: add pending and current descriptions"

saghul:
This reverts commit c3effb3a9caa0ac2a3ff33ba8cb08959e63692be.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -502,27 +502,9 @@ class PeerConnectionObserver implements PeerConnection.Observer {
     @Override
     public void onSignalingChange(PeerConnection.SignalingState signalingState) {
         ThreadUtils.runOnExecutor(() -> {
-            WritableMap newLocalSdpMap = Arguments.createMap();
-            WritableMap newRemoteSdpMap = Arguments.createMap();
-            SessionDescription newLocalSdp = peerConnection.getLocalDescription();
-            SessionDescription newRemoteSdp = peerConnection.getRemoteDescription();
-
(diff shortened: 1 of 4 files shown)


8BallBomBom:
Can't say i've had issues with this change so far 🤔
But if it is causing issues then best to be included in another release 👍🏻 
Should we look at doing more test cases?

saghul:
I marked it as a draft because I need to do a full bisect but this broke remote video for us.

I suspect it's because the SDP won't be updated right after calling sLD, and that breaks code that expects that.

Hopefully I can get to the bottom of it in the upcoming days.

8BallBomBom:
Currently building some new examples so i'll do some testing on one of the new clean projects.
See if anything pops up 👍🏻

saghul:
Sorry @shepeliev I think doing everything based on signalling state changes won't fly. The event will be processed after sLD and the SDP won't be updated right after sLD resolves, which is an expectation apps have.

Let's give it another try, with the wisdom of hindsight :-)

8BallBomBom:
As they say, try, try again!
The groundwork has been laid 👍🏻 

This one did kind of catch me a bit off guard.
The way i've implemented the use of this module, can't say i've experienced any problems.
After looking around at some examples, it is understandable why this caused issues.

But as we're always fighting for spec compliance, a few nudges here and there, should be off to the races 😸

saghul:
We do extensive SDP shenanigans, so at least there a way for me to know if we have messed up something there 😅 

Granted, it's hard to catch 'em all...

shepeliev:
Looks like we have to sync pending SDP in sLD too. I'll back to this as soon as be able. Anyway, it should be aligned with specs as we have different behavior on Web and iOS/Android in our project.

PR #1459, opened 2023-10-13
Added UVC Camera support

tommiilmonen:
Hello

I added UVC device support on Android to this library.
I checked [cristiano-linvix](https://github.com/cristiano-linvix)'s closed pull request https://github.com/react-native-webrtc/react-native-webrtc/pull/1285 about the same feature, and did some improvements.

The audio support from UVC device is not added in this pull request, since it's implementation is not as straight forward IMO.

EDIT: Fix typo


android/src/main/java/com/oney/WebRTCModule/UVCCamera2Enumerator.java
```diff
@@ -0,0 +1,101 @@
...
+    public UVCCamera2Enumerator(Context context) {
+        super(context);
+        this.uvcMonitor = new UVCMonitor(context);
+        this.context = context;
+    }
+
+    @Override
+    public String[] getDeviceNames() {
+        ArrayList<String> devicesNames = new ArrayList<>(Arrays.asList(super.getDeviceNames()));
+        UsbDevice uvcDevice = this.uvcMonitor.getUvcCameraDevice();
```
tommiilmonen, line 45: We could get the uvc device from the strategy itself, but it has certain weaknesses. Ie: The camera needs to be registered and will need a permission from the user, which IMO should only be needed when user wants to stream the camera).

Thus, i created a wrapper (UVCMonitor), which can provide UVC device without prompting permission from the user.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -70,8 +70,8 @@ class GetUserMediaImpl {
         }
 
         if (camera2supported) {
-            Log.d(TAG, "Creating video capturer using Camera2 API.");
-            cameraEnumerator = new Camera2Enumerator(reactContext);
+            Log.d(TAG, "Creating video capturer using Camera2 API with UVC support.");
+            cameraEnumerator = new UVCCamera2Enumerator(reactContext);
```
tommiilmonen, line 74: I'd like if the user could explicitly tell the module to use certain implementation (Camera2Enumerator vs UVCCamera2Enumartor). I'm not sure how to do this correctly, since i only wanted to add implementations and not change the other source code. What do you think?

android/src/main/java/com/oney/WebRTCModule/UVCCamera2Enumerator.java
saghul: You mean enumerating the devices would require permissions?

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
saghul: Great question.

Generally speaking I think it should work out of the box, so it Just Works (TM).

A question I have is: is there a chance we get duplicate devices? As in, the builtin enumerator would detect the camera, then the UVC one would proceed to detect the same. Is that possible at all?

android/src/main/java/com/oney/WebRTCModule/UVCMonitor.java
```diff
@@ -0,0 +1,62 @@
...
+
+import androidx.annotation.Nullable;
+
+import com.jiangdg.usb.USBMonitor;
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class UVCMonitor {
```
saghul, line 18: Is any permission necessary for this to work?

android/src/main/java/com/oney/WebRTCModule/UVCMonitor.java
```diff
@@ -0,0 +1,62 @@
...
+                .stream()
+                .filter(d -> SUPPORTED_USB_CLASSES.contains(d.getDeviceClass()))
+                .collect(Collectors.toList());
+        if (devices.size() > 1) {
+            Log.w(WebRTCModule.TAG, "Found more than one UVC Cameras. Using the one found first.");
+        }
+        return devices.size() > 0 ? devices.get(0) : null;
+    }
+
+    private static class DeviceConnectListener implements USBMonitor.OnDeviceConnectListener {
```
saghul, line 46: Maybe add some logs here?

android/src/main/java/com/oney/WebRTCModule/UVCVideoCapturer.java
```diff
@@ -0,0 +1,117 @@
...
+    public void switchCamera(CameraSwitchHandler cameraSwitchHandler, String s) {
+        cameraSwitchHandler.onCameraSwitchError("Not supported yet with uvc");
+    }
+
+    private CameraRequest createCameraRequest(int width, int height) {
+        return new CameraRequest
+                .Builder()
+                .setPreviewWidth(width)
+                .setPreviewHeight(height)
+                .setAudioSource(CameraRequest.AudioSource.SOURCE_AUTO)
```
saghul, line 94: Audio?

android/src/main/java/com/oney/WebRTCModule/UVCVideoCapturer.java
```diff
@@ -0,0 +1,117 @@
...
+
+        public PreviewDataCallback(Consumer<VideoFrame> videoFrameConsumer) {
+            this.videoFrameConsumer = videoFrameConsumer;
+        }
+
+        @Override
+        public void onPreviewData(@Nullable byte[] bytes, int width, int height, @NonNull DataFormat dataFormat) {
+            if (DataFormat.NV21.equals(dataFormat)) {
+                NV21Buffer nv21Buffer = new NV21Buffer(bytes, width, height, null);
+                VideoFrame frame = new VideoFrame(nv21Buffer, 0, System.nanoTime());
```
saghul, line 110: Doewn't the frame need to be released after passing it along?

saghul:
I really like how concise this is! Left some comments!

android/src/main/java/com/oney/WebRTCModule/UVCCamera2Enumerator.java
tommiilmonen: Yes, we would need to request a permission to use the attached UVC device in order to even enumerate it.

This approach (where permission is asked when actual streaming is started) is much more user friendly imo.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
tommiilmonen: Relating to your question about duplicate devices:

Yes, i think there is a slight chance that the built-in Camera2Enumerator could provide the external camera as well (This depends on the device and the manufacturer as far as i know).

I don't see this as an issue by the following reasoning:
If the external camera is enumerated twice (ie. by built-in and this ad-hoc), it should work whether the user chooses one or the other. Built-in enumerated uvc device will use the built-in capturer and ad-hoc will use the ad-hoc, since these appear as two distinct devices.

Of course, optimally the ad-hoc enumerated device should not be added to the enumerated devices if the built-in detects it, but currently i don't know how to robustly filter it out since none of my phones support external camera directly.

android/src/main/java/com/oney/WebRTCModule/UVCCamera2Enumerator.java
saghul: I see. Users would need to enumerate the device first to know it's there, wouldn't they?

tommiilmonen: I think i explained this a bit confusingly, sorry.

I'll try again,
CameraUvcStrategy (which we use in the capturer) has a getCurrentDevice(): UsbDevice method, which could be used for enumerating purposes. But, in order to use that method we would need a runtime permission from the user.
By using the UVCMonitor, no runtime- or app permissions are required for the enumeration part.

saghul: Understood now, that sounds great!

android/src/main/java/com/oney/WebRTCModule/UVCMonitor.java
tommiilmonen: Actually we do not need onXXX listeners at all, so i refactored UVCMonitor to use android's native UsbManager.

android/src/main/java/com/oney/WebRTCModule/UVCVideoCapturer.java
tommiilmonen: Removed this to prevent confusion, since audio is not yet captured from the uvc device.

android/src/main/java/com/oney/WebRTCModule/UVCVideoCapturer.java
tommiilmonen: Yes, good catch!

tommiilmonen:
@saghul

Thanks for the comments. Checked and fixed them. Can this be approved and merged?

saghul:
I'll take another look soon!

saghul:
I have been busy and unable to test. Have you tested it? More testing would certainly help.

tommiilmonen:
I have tested the implementation and it is ready from my part.

saghul:
Thanks for giving it a try!

cristiano-linvix:
Hi guys, everything good?

I'm coming back with this topic, do you need help with anything?

In my case, I need audio too, but I don't know how to implement it, I managed to only make video in the other [PR](https://github.com/react-native-webrtc/react-native-webrtc/pull/1285) I made.

I have external UVC cameras here to test if needed!

Good job @tommiilmonen, you need help?

cristiano-linvix:
I was unable to use UVC with an external USB device.
In the other PR, access to an external USB device was working, perhaps because of the implementation in this [link](https://github.com/cristiano-linvix/react-native-webrtc/blob/53414f9481724ae29e7b86abf318acc87de39af1/android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java#L33).

Can we implement support?

@tommiilmonen

tommiilmonen:
I would prefer adding the UsbClass value 255 to the supported types instead of using the uvcStrategy.
I can add this value to the supported types if/when you can verify that it works @cristiano-linvix ?

I didn't wanna use uvcStrategy for some reason that i can not remember anymore :(
(I think it had to do with registering for usb events which required permissions iirc)

saghul:
> I didn't wanna use uvcStrategy for some reason that i can not remember anymore :(
> (I think it had to do with registering for usb events which required permissions iirc)

It would be good to know this.

cristiano-linvix:
> I would prefer adding the UsbClass value 255 to the supported types instead of using the uvcStrategy. I can add this value to the supported types if/when you can verify that it works @cristiano-linvix ?

It finds the device, but for some strange reason, it doesn't launch the preview.

I'm trying to look at the PR I made and compare to find a possible difference between the two to correct.

This message appears in the Log
```bash
2023-12-05 03:06:36.612 2843-4159 CameraUvc br.com.linvix.apps.selfcheckout E Find no uvc devices, if you want some special device please use getUsbDeviceList() or add device info into default_device_filter.xml

```

@tommiilmonen Can you contact me via email or discord so we can try to finalize this PR?

cristiano-linvix:
I located an interesting file, which allows me to configure devices recognized by the lib.

https://github.com/jiangdongguo/AndroidUSBCamera/blob/master/app/src/main/res/xml/default_device_filter.xml

By placing the device there, the webcam works.

But, device cannot return in `getUvcCameraDevice`.

One difficulty I had here was because on my device there is a device that appears that I have no idea what it is, and it appears in the list.

I think we should list all available uvc cameras and pass the ID to be able to launch the correct camera with the camera's deviceId.

tommiilmonen:
So if i understood correct, we just remove the filter fromhttps://github.com/react-native-webrtc/react-native-webrtc/pull/1459/files#diff-3503f0dd8e37ca538dec357613394b5cc012f88d5dea027386f7b8924a1bb830 line 43and we are good to go (provided user has set the right xml for weirdo cameras)?  Lähetetty Windowsin Sähköpostiista Lähettäjä: Cristiano CruzLähetetty: tiistai 5. joulukuuta 2023 15.45Vastaanottaja: react-native-webrtc/react-native-webrtcKopio: Tommi Ilmonen; MentionAihe: Re: [react-native-webrtc/react-native-webrtc] Added UVC Camera support (PR #1459) I located an interesting file, which allows me to configure devices recognized by the lib.https://github.com/jiangdongguo/AndroidUSBCamera/blob/master/app/src/main/res/xml/default_device_filter.xmlBy placing the device there, the webcam works.But, device cannot return in getUvcCameraDevice.One difficulty I had here was because on my device there is a device that appears that I have no idea what it is, and it appears in the list.I think we should list all available uvc cameras and pass the ID to be able to launch the correct camera with the camera's deviceId.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

tommiilmonen:
> I located an interesting file, which allows me to configure devices recognized by the lib.
> 
> https://github.com/jiangdongguo/AndroidUSBCamera/blob/master/app/src/main/res/xml/default_device_filter.xml
> 
> By placing the device there, the webcam works.
> 
> But, device cannot return in `getUvcCameraDevice`.
> 
> One difficulty I had here was because on my device there is a device that appears that I have no idea what it is, and it appears in the list.
> 
> I think we should list all available uvc cameras and pass the ID to be able to launch the correct camera with the camera's deviceId.

Okay. I will add the usb class 255 to the supported devices. This way your setup will also work provided that you have set the definition xml to the srcs.

cristiano-linvix:
Here we go, things I've discovered so far:

Creating the `default_device_filter.xml` file and placing the camera configuration with UVC support, it starts to recognize/work the camera.
```xml
<?xml version="1.0" encoding="utf-8"?>
<usb>
    <usb-device class="239" subclass="2" />	<!-- all device of UVC -->
    <!-- a few android 9.0 -->
    <usb-device class="14" subclass="9" />
    <usb-device class="2" subclass="0" />
    <usb-device class="6" subclass="-1" />
    <usb-device class="39" subclass="0" />
    <usb-device product-id="4836" vendor-id="9573" />
    <usb-device product-id="2229" vendor-id="1133" />
    <usb-device product-id="640" vendor-id="1409" />
    <usb-device product-id="258" vendor-id="9228" />
    <usb-device product-id="16981" vendor-id="19" />
    <usb-device product-id="5120" vendor-id="1409" />

    <!--  Logitech C270  -->
    <usb-device product-id="6144" vendor-id="742" />
</usb>
```

Things we need to check:
There can be two UVC cameras on the device, we should allow the user to choose the camera he wants, there is a setCameraId property in the CameraRequest.
In the function where it maps cameras we need to return all available cameras.

I can help you with this path of allowing multiple cameras.

> I didn't wanna use uvcStrategy for some reason that i can not remember anymore :(
> (I think it had to do with registering for usb events which required permissions iirc)

Can we test your reason?
The lib documentation recommends using uvcStrategy to locate USB devices, can we merge both?

tommiilmonen:
How can there be two uvc cameras simoultaneosly?

cristiano-linvix:
There may be 2 cameras connected, but not in use simultaneously, he may want to switch between cameras (front or back for example).

cristiano-linvix:
I sent you an email on my discord to talk about it.

saghul:
Having the discussion here would benefit everyone, folks.

I will not blindly merge this so you're going to have to repeat it here anyway.

cristiano-linvix:
@saghul
Can you give some guidance on how we could implement audio as well?

android/src/main/java/com/oney/WebRTCModule/UVCCamera2Enumerator.java
```diff
@@ -0,0 +1,107 @@
...
+        try {
+            CameraUvcStrategy uvcStrategy = new CameraUvcStrategy(context);
+            List<UsbDevice> uvcCameras = uvcStrategy.getUsbDeviceList(null); // null := no filtering
+            if (uvcCameras != null) {
+                uvcCameras.forEach(camera -> devicesNames.add(UVC_PREFIX + camera.getDeviceName()));
+            }
+            uvcStrategy.unRegister();
+        } catch (Exception e) {
+            Log.e(WebRTCModule.TAG, "Error in uvc camera enumeration", e);
+        }
```
tommiilmonen, line 53: Okay we fixed issues that @cristiano-linvix was having.

We changed the enumeration to use CameraUvcStrategy, which detects uvc cameras more robustly than my own implementation. Also, it now provides all available UVC devices instead of just only one.

@saghul

cristiano-linvix:
Working perfectly!

This screenshot is already using a WebRTC connection
Client Windows (Chrome) calling Android

Tested with the Logitech C270 camera

Captura de Tela 2023-12-05 às 16 10 05

android/src/main/java/com/oney/WebRTCModule/UVCCamera2Enumerator.java
saghul: Does it have any permissions implications?

saghul:
Great news!

saghul:
> @saghul
> 
> Can you give some guidance on how we could implement audio as well?

That is a whole other beast, currently out of the scope of this PR.

Let's have this first, we'll see after.

cristiano-linvix:
> That is a whole other beast, currently out of the scope of this PR.
> 
> Let's have this first, we'll see after.

Ok, thanks.

android/src/main/java/com/oney/WebRTCModule/UVCCamera2Enumerator.java
cristiano-linvix: I was asked for camera permission when I started streaming to webrtc.

No other permissions were requested.

My `AndroidManifest.xml`

```xml
    <!-- react-native-webrtc -->
    <uses-feature android:name="android.hardware.camera" />
    <uses-feature android:name="android.hardware.camera.autofocus" />
    <uses-feature android:name="android.hardware.audio.output" />
    <uses-feature android:name="android.hardware.microphone" />

    <!-- react-native-webrtc -->
    <uses-permission android:name="android.permission.CAMERA" />
    <uses-permission android:name="android.permission.RECORD_AUDIO" />
    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
    <uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
    <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
```

davidliu:
Tested it as well, worked pretty smoothly without any extra steps required.

saghul:
Excellent! So there are no extra permissions shenanigans now?

saghul:
I'm ready to cut a new release if I get confirmation on my point above.

tommiilmonen:
I'll try to answer during next week.

8BallBomBom:
Managed to get my hands on a cam, will do some testing over the next day or so 👍🏻 
As for the permissions, unless anything has majorly changed since the last time.
Should be good 😸 but will test.

davidliu:
> Excellent! So there are no extra permissions shenanigans now?

There's nothing on the dev side, there's a runtime permission that gets automatically shown when requesting the usb camera through `getUserMedia` and the camera stream is properly linked up after the permission is given.

There is a small change to the android manifest however. The [underlying library](https://github.com/jiangdongguo/AndroidUSBCamera/blob/master/libausbc/src/main/AndroidManifest.xml) adds a `<uses-feature android:name="android.hardware.usb.host"/>`, which could be problematic for play store listings. I think adding a `android:required="false"` to our android manifest will override this and fix it.

saghul:
> there's a runtime permission that gets automatically shown when requesting the usb camera through `getUserMedia`

Does it also show when using `getUserMedia` without specifying a device ID?



> There is a small change to the android manifest however. The [underlying library](https://github.com/jiangdongguo/AndroidUSBCamera/blob/master/libausbc/src/main/AndroidManifest.xml) adds a `<uses-feature android:name="android.hardware.usb.host"/>`, which could be problematic for play store listings.

Why do you think so? Nothing seems to be mentioned here: https://developer.android.com/guide/topics/manifest/uses-feature-element

saghul:
> I think adding a `android:required="false"` to our android manifest will override this and fix it.

Sounds good!

davidliu:
> Does it also show when using `getUserMedia` without specifying a device ID?

Using `{video: true}` defaults to my device's front camera.

> Why do you think so? Nothing seems to be mentioned here: https://developer.android.com/guide/topics/manifest/uses-feature-element

Never had to deal with the `use-feature` tag myself, but the page mentions that the play store may not show an app for users who don't meet the feature requirements. 

https://developer.android.com/guide/topics/manifest/uses-feature-element#market-feature-filtering

Setting the required to false will let the play store know not to filter on this requirement.

saghul:
Makes sense. Can you suggest the code to do that?

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
davidliu, line 74: Not sure if there's any issues if for some reason a device doesn't support USB hosting, but to be safe, it might good to add a check for `PackageManager.hasSystemFeature(PackageManager.FEATURE_USB_HOST)` and fall back to the regular Camera2Enumerator if it can't support USB cameras.

davidliu:
@tommiilmonen can you change the [base AndroidManifest.xml](https://github.com/react-native-webrtc/react-native-webrtc/blob/master/android/src/main/AndroidManifest.xml) to look like this?

```
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
  package="com.oney.WebRTCModule"
  xmlns:tools="http://schemas.android.com/tools">

    <uses-feature android:name="android.hardware.usb.host" 
      android:required="false"
      tools:node="replace"/>

</manifest>
```

This properly replaces the libausb `use-feature` declaration with our own..

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
saghul: Good call.

8BallBomBom: Very good call, starting to think i might not have any devices with that feature 🤔

8BallBomBom:
> There's nothing on the dev side, there's a runtime permission that gets automatically shown when requesting the usb camera through `getUserMedia` and the camera stream is properly linked up after the permission is given.

Same result as default, if no permissions are already given then they get requested at runtime 😺

davidliu:
Oh shoot, I think I misunderstood @saghul's comment and was a little unclear myself.

There's a permission popup specific to the usb device that shows up when you want to use it (lasts until the usb camera is unplugged), separate from the existing permissions. It's needed even if you have the camera/mic permissions.

This is correctly not shown when other cameras are chosen.

8BallBomBom:
Ahhhhhh yes, that could be a bit of an issue 🤔

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
tommiilmonen: This should not be an issue since the implementation extends the Camera2Enumerator class (and uvc device enumeration exceptions are handled). But i will add the FEATURE_USB_HOST check just in case.

saghul:
As long as it only shows up when trying to use the camera we should be fine!

tommiilmonen:
I added the use-feature override to manifest xml. @davidliu 

Also now
`PackageManager.hasSystemFeature(PackageManager.FEATURE_USB_HOST)`
is checked when determining the enumerator instance.

What comes to the permissions, it goes exactly as david described.

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
8BallBomBom: Doesn't hurt sometimes to try and be a bit more safe.
Can be the difference between pushing forward and dropping off a cliff 😄

saghul:
Can you run `npm run format` and commit the result please?

android/src/main/java/com/oney/WebRTCModule/UVCCamera2Enumerator.java
8BallBomBom: Will need to add something to the docs/handbook for this.
```xml
<uses-feature android:name="android.hardware.usb.host" />
```
But all good 👍🏻

tommiilmonen:
Sure, formatted now.

saghul (approved):

saghul:
It's in, thanks a lot!

cristiano-linvix:
Awesome!

8BallBomBom:
Forgot to update on this but managed to test with an old Logitech cam without issues 👍🏻 
Now we need to have a look into the audio issue 🤔 over [here](https://github.com/react-native-webrtc/react-native-webrtc/issues/1485).

cristiano-linvix:
> Forgot to update on this but managed to test with an old Logitech cam without issues 👍🏻 Now we need to have a look into the audio issue 🤔 over [here](https://github.com/react-native-webrtc/react-native-webrtc/issues/1485).

I'll try to look at this today or tomorrow, let me know in the issue if it worked or not.

imamabdulazis:
> Oh shoot, I think I misunderstood @saghul's comment and was a little unclear myself.
> 
> There's a permission popup specific to the usb device that shows up when you want to use it (lasts until the usb camera is unplugged), separate from the existing permissions. It's needed even if you have the camera/mic permissions.
> 
> This is correctly not shown when other cameras are chosen.
> 

I think the USB camera is still not working, the permission to use the USB camera is not showing as mentioned before.

saghul:
Note that UVC support was disabled on Android 14 due to a bug in the library.

imamabdulazis:
@saghul I was using android 11, but I think the issue not only on this library since I was using [Jitsi meet](https://github.com/jitsi/jitsi-meet) react native which is using this library to their camera.
I check the new version has been used the webrtc version for uvc support.
But, when I plug-in the USB camera the permission doesn't come up.

I already patch this package before and working fine for version `react-native-webrtc": "111.0.1"`, but found and issue when `https://github.com/jiangdongguo/AndroidUSBCamera` is update their package, the old library has no longer supported in a few function.

<details>
  <summary>Patch code</summary>
  
  ```js
 --- a/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
+++ b/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
@@ -1,15 +1,23 @@
 package com.oney.WebRTCModule;
 
+import android.app.AlertDialog;
+import android.content.DialogInterface;
 import android.util.Log;
 
+import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 
+import android.content.Context;
+
+import com.jiangdg.ausbc.camera.CameraUvcStrategy;
+
 import org.webrtc.CameraEnumerator;
 import org.webrtc.CameraVideoCapturer;
 import org.webrtc.VideoCapturer;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 public class CameraCaptureController extends AbstractVideoCaptureController {
     /**
@@ -21,6 +29,9 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
 
     private final CameraEnumerator cameraEnumerator;
     private final ReadableMap constraints;
+    private final CameraUvcStrategy mUvcStrategy;
+
+    private final UvcEnumerator uvcEnumerator;
 
     /**
      * The {@link CameraEventsHandler} used with
@@ -30,11 +41,14 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
      */
     private final CameraEventsHandler cameraEventsHandler = new CameraEventsHandler();
 
-    public CameraCaptureController(CameraEnumerator cameraEnumerator, ReadableMap constraints) {
+    public CameraCaptureController(CameraEnumerator cameraEnumerator, ReadableMap constraints,
+                                   CameraUvcStrategy uvcStrategy, UvcEnumerator uvcEnumerator) {
         super(constraints.getInt("width"), constraints.getInt("height"), constraints.getInt("frameRate"));
 
         this.cameraEnumerator = cameraEnumerator;
         this.constraints = constraints;
+        this.mUvcStrategy = uvcStrategy;
+        this.uvcEnumerator = uvcEnumerator;
     }
 
     public void switchCamera() {
@@ -79,10 +93,11 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
     }
 
     /**
-     * Helper function which tries to switch cameras until the desired facing mode is found.
+     * Helper function which tries to switch cameras until the desired facing mode
+     * is found.
      *
      * @param desiredFrontFacing - The desired front facing value.
-     * @param tries - How many times to try switching.
+     * @param tries              - How many times to try switching.
      */
     private void switchCamera(boolean desiredFrontFacing, int tries) {
         CameraVideoCapturer capturer = (CameraVideoCapturer) videoCapturer;
@@ -111,18 +126,40 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
      * Constructs a new {@code VideoCapturer} instance attempting to satisfy
      * specific constraints.
      *
-     * @param deviceId the ID of the requested video device. If not
-     * {@code null} and a {@code VideoCapturer} can be created for it, then
-     * {@code facingMode} is ignored.
+     * @param deviceId   the ID of the requested video device. If not
+     *                   {@code null} and a {@code VideoCapturer} can be created for
+     *                   it, then
+     *                   {@code facingMode} is ignored.
      * @param facingMode the facing of the requested video source such as
-     * {@code user} and {@code environment}. If {@code null}, "user" is
-     * presumed.
+     *                   {@code user} and {@code environment}. If {@code null},
+     *                   "user" is
+     *                   presumed.
      * @return a {@code VideoCapturer} satisfying the {@code facingMode} or
      * {@code deviceId} constraint
      */
     private VideoCapturer createVideoCapturer(String deviceId, String facingMode) {
         String[] deviceNames = cameraEnumerator.getDeviceNames();
         List<String> failedDevices = new ArrayList<>();
+        ReadableArray uvcCameraList = uvcEnumerator.enumerateDevices();
+
+        Log.d(TAG, "UVC_CHECKKKKK: " + uvcCameraList.size());
+
+        // !UVC
+        if (uvcCameraList.size() > 0) {
+            ReadableMap uvcCamera = uvcCameraList.getMap(0);
+            String uvcDeviceId = uvcCamera.getString("deviceId");
+            Log.d(TAG, "uvcDevice " + uvcCamera.getString("deviceId"));
+            // if has uvc into id, instance uvc capturer
+            String message = "Create UVC user-specified camera " + uvcDeviceId;
+            try {
+                UvcCapturer videoCapturer = new UvcCapturer(uvcDeviceId, mUvcStrategy);
+                Log.d(TAG, message + " succeeded");
+                return videoCapturer;
+            } catch (Throwable tr) {
+                Log.d(TAG, message + " failed");
+            }
+        }
+        // !UVC
 
         String cameraName = null;
         try {
@@ -132,6 +169,8 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
             Log.d(TAG, "failed to find device with id: " + deviceId);
         }
 
+        Log.d(TAG, "CAMERA_NAME: " + cameraName);
+
         // If deviceId is specified, then it takes precedence over facingMode.
         if (cameraName != null) {
             VideoCapturer videoCapturer = cameraEnumerator.createCapturer(cameraName, cameraEventsHandler);
diff --git a/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java b/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
index b945603..bd2252c 100644
--- a/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
+++ b/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
@@ -6,6 +6,7 @@ import android.content.Intent;
 import android.media.projection.MediaProjectionManager;
 import android.util.DisplayMetrics;
 import android.util.Log;
+import com.jiangdg.ausbc.camera.CameraUvcStrategy;
 
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.BaseActivityEventListener;
@@ -55,16 +56,22 @@ class GetUserMediaImpl {
     private Promise displayMediaPromise;
     private Intent mediaProjectionPermissionResultData;
 
+    private UvcEnumerator uvcEnumerator;
+    private CameraUvcStrategy cameraUvcStrategy;
+
     GetUserMediaImpl(WebRTCModule webRTCModule, ReactApplicationContext reactContext) {
         this.webRTCModule = webRTCModule;
         this.reactContext = reactContext;
+        this.cameraUvcStrategy = new CameraUvcStrategy(reactContext);
+        this.uvcEnumerator = new UvcEnumerator(this.cameraUvcStrategy);
 
         boolean camera2supported = false;
 
         try {
             camera2supported = Camera2Enumerator.isSupported(reactContext);
         } catch (Throwable tr) {
-            // Some devices will crash here with: Fatal Exception: java.lang.AssertionError: Supported FPS ranges cannot
+            // Some devices will crash here with: Fatal Exception: java.lang.AssertionError:
+            // Supported FPS ranges cannot
             // be null. Make sure we don't.
             Log.w(TAG, "Error checking for Camera2 API support.", tr);
         }
@@ -104,7 +111,8 @@ class GetUserMediaImpl {
         PeerConnectionFactory pcFactory = webRTCModule.mFactory;
         MediaConstraints peerConstraints = webRTCModule.constraintsForOptions(audioConstraintsMap);
 
-        // PeerConnectionFactory.createAudioSource will throw an error when mandatory constraints contain nulls.
+        // PeerConnectionFactory.createAudioSource will throw an error when mandatory
+        // constraints contain nulls.
         // so, let's check for nulls
         checkMandatoryConstraints(peerConstraints);
 
@@ -134,8 +142,15 @@ class GetUserMediaImpl {
 
     ReadableArray enumerateDevices() {
         WritableArray array = Arguments.createArray();
+        ReadableArray uvcDevices = uvcEnumerator.enumerateDevices();
         String[] devices = cameraEnumerator.getDeviceNames();
 
+        if (uvcDevices.size() > 0) {
+            for (int j = 0; j < uvcDevices.size(); j++) {
+                array.pushMap(uvcDevices.getMap(j));
+            }
+        }
+
         for (int i = 0; i < devices.length; ++i) {
             String deviceName = devices[i];
             boolean isFrontFacing;
@@ -191,8 +206,7 @@ class GetUserMediaImpl {
 
             Log.d(TAG, "getUserMedia(video): " + videoConstraintsMap);
 
-            CameraCaptureController cameraCaptureController =
-                    new CameraCaptureController(cameraEnumerator, videoConstraintsMap);
+            CameraCaptureController cameraCaptureController = new CameraCaptureController(cameraEnumerator,videoConstraintsMap, cameraUvcStrategy,uvcEnumerator);
 
             videoTrack = createVideoTrack(cameraCaptureController);
         }
@@ -204,7 +218,7 @@ class GetUserMediaImpl {
             return;
         }
 
-        createStream(new MediaStreamTrack[] {audioTrack, videoTrack}, (streamId, tracksInfo) -> {
+        createStream(new MediaStreamTrack[] { audioTrack, videoTrack }, (streamId, tracksInfo) -> {
             WritableArray tracksInfoWritableArray = Arguments.createArray();
 
             for (WritableMap trackInfo : tracksInfo) {
@@ -255,8 +269,8 @@ class GetUserMediaImpl {
 
         this.displayMediaPromise = promise;
 
-        MediaProjectionManager mediaProjectionManager =
-                (MediaProjectionManager) currentActivity.getApplication().getSystemService(
+        MediaProjectionManager mediaProjectionManager = (MediaProjectionManager) currentActivity.getApplication()
+                .getSystemService(
                         Context.MEDIA_PROJECTION_SERVICE);
 
         if (mediaProjectionManager != null) {
@@ -279,7 +293,7 @@ class GetUserMediaImpl {
         if (track == null) {
             displayMediaPromise.reject(new RuntimeException("ScreenTrack is null."));
         } else {
-            createStream(new MediaStreamTrack[] {track}, (streamId, tracksInfo) -> {
+            createStream(new MediaStreamTrack[] { track }, (streamId, tracksInfo) -> {
                 WritableMap data = Arguments.createMap();
 
                 data.putString("streamId", streamId);
@@ -389,10 +403,12 @@ class GetUserMediaImpl {
     }
 
     /**
-     * Set video effect to the TrackPrivate corresponding to the trackId with the help of VideoEffectProcessor
+     * Set video effect to the TrackPrivate corresponding to the trackId with the
+     * help of VideoEffectProcessor
      * corresponding to the name.
+     * 
      * @param trackId TrackPrivate id
-     * @param name VideoEffectProcessor name
+     * @param name    VideoEffectProcessor name
      */
     void setVideoEffect(String trackId, String name) {
         TrackPrivate track = tracks.get(trackId);
@@ -409,8 +425,8 @@ class GetUserMediaImpl {
                     return;
                 }
 
-                VideoEffectProcessor videoEffectProcessor =
-                        new VideoEffectProcessor(videoFrameProcessor, surfaceTextureHelper);
+                VideoEffectProcessor videoEffectProcessor = new VideoEffectProcessor(videoFrameProcessor,
+                        surfaceTextureHelper);
                 videoSource.setVideoProcessor(videoEffectProcessor);
 
             } else {
@@ -448,10 +464,13 @@ class GetUserMediaImpl {
          * Initializes a new {@code TrackPrivate} instance.
          *
          * @param track
-         * @param mediaSource            the {@code MediaSource} from which the specified
+         * @param mediaSource            the {@code MediaSource} from which the
+         *                               specified
          *                               {@code code} was created
-         * @param videoCaptureController the {@code AbstractVideoCaptureController} from which the
-         *                               specified {@code mediaSource} was created if the specified
+         * @param videoCaptureController the {@code AbstractVideoCaptureController} from
+         *                               which the
+         *                               specified {@code mediaSource} was created if
+         *                               the specified
          *                               {@code track} is a {@link VideoTrack}
          */
         public TrackPrivate(MediaStreamTrack track, MediaSource mediaSource,
@@ -472,10 +491,14 @@ class GetUserMediaImpl {
                 }
 
                 /*
-                 * As per webrtc library documentation - The caller still has ownership of {@code
-                 * surfaceTextureHelper} and is responsible for making sure surfaceTextureHelper.dispose() is
-                 * called. This also means that the caller can reuse the SurfaceTextureHelper to initialize a new
-                 * VideoCapturer once the previous VideoCapturer has been disposed. */
+                 * As per webrtc library documentation - The caller still has ownership of
+                 * {@code
+                 * surfaceTextureHelper} and is responsible for making sure
+                 * surfaceTextureHelper.dispose() is
+                 * called. This also means that the caller can reuse the SurfaceTextureHelper to
+                 * initialize a new
+                 * VideoCapturer once the previous VideoCapturer has been disposed.
+                 */
 
                 if (surfaceTextureHelper != null) {
                     surfaceTextureHelper.stopListening();
diff --git a/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java b/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
new file mode 100644
index 0000000..a6696cb
--- /dev/null
+++ b/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/UvcCapturer.java
@@ -0,0 +1,132 @@
+package com.oney.WebRTCModule;
+
+import android.content.Context;
+import android.util.Log;
+import android.widget.Toast;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.jiangdg.ausbc.callback.IPreviewDataCallBack;
+import com.jiangdg.ausbc.camera.CameraUvcStrategy;
+import com.jiangdg.ausbc.camera.bean.CameraRequest;
+import com.jiangdg.ausbc.render.env.RotateType;
+
+import org.webrtc.CapturerObserver;
+import org.webrtc.NV21Buffer;
+import org.webrtc.SurfaceTextureHelper;
+import org.webrtc.SurfaceViewRenderer;
+import org.webrtc.VideoCapturer;
+import org.webrtc.VideoFrame;
+
+public class UvcCapturer implements VideoCapturer {
+
+    private static final String TAG = WebRTCModule.TAG;
+
+    private String deviceId;
+
+    private SurfaceTextureHelper surfaceTextureHelper;
+    private CapturerObserver capturerObserver;
+
+    private int UVC_PREVIEW_WIDTH = 1024;
+    private int UVC_PREVIEW_HEIGHT = 768;
+    private int UVC_PREVIEW_FPS = 30;
+
+    CameraUvcStrategy mUvcStrategy;
+
+    public UvcCapturer(String deviceId, CameraUvcStrategy uvcStrategy) {
+        this.deviceId = deviceId.replace("uvc:", "");
+        Log.d(TAG, "UvcCapturer.UvcCapturer->DeviceID " + this.deviceId);
+        mUvcStrategy = uvcStrategy;
+    }
+
+    @Override
+    public void initialize(SurfaceTextureHelper surfaceTextureHelper, Context context, CapturerObserver capturerObserver) {
+        this.surfaceTextureHelper = surfaceTextureHelper;
+        this.capturerObserver = capturerObserver;
+    }
+
+    @Override
+    public void startCapture(int width, int height, int fps) {
+        if (mUvcStrategy != null) {
+            // stop before to free last usb instance
+            if (mUvcStrategy.isCameraOpened()) {
+                mUvcStrategy.stopPreview();
+            }
+            Log.d(TAG, "SIZEE__WIDTH: "+mUvcStrategy.getAllPreviewSizes(1.0));
+
+            UVC_PREVIEW_WIDTH = width;
+            UVC_PREVIEW_HEIGHT = height;
+            UVC_PREVIEW_FPS = fps;
+
+//            mUvcStrategy.addPreviewDataCallBack(new IPreviewDataCallBack() {
+//                @Override
+//                public void onPreviewData(@Nullable byte[] bytes, int i, int i1, @NonNull DataFormat dataFormat) {
+//                    NV21Buffer nv21Buffer = new NV21Buffer(bytes, UVC_PREVIEW_WIDTH, UVC_PREVIEW_HEIGHT, null);
+//                    VideoFrame frame = new VideoFrame(nv21Buffer, 0, System.nanoTime());
+//                    capturerObserver.onFrameCaptured(frame);
+//                }
+//            });
+            mUvcStrategy.addPreviewDataCallBack(new IPreviewDataCallBack() {
+                @Override
+                public void onPreviewData(@Nullable byte[] bytes, @NonNull DataFormat dataFormat) {
+                            NV21Buffer nv21Buffer = new NV21Buffer(bytes, UVC_PREVIEW_WIDTH, UVC_PREVIEW_HEIGHT, null);
+                    VideoFrame frame = new VideoFrame(nv21Buffer, 0, System.nanoTime());
+                    capturerObserver.onFrameCaptured(frame);
+                }
+            });
+
+
+            mUvcStrategy.startPreview(getCameraRequest(), this.surfaceTextureHelper.getSurfaceTexture());
+        }
+    }
+
+    @Override
+    public void stopCapture() throws InterruptedException {
+        if (mUvcStrategy != null) {
+            if (mUvcStrategy.isCameraOpened()) {
+                mUvcStrategy.stopPreview();
+            }
+        }
+    }
+
+    @Override
+    public void changeCaptureFormat(int i, int i1, int i2) {
+
+    }
+
+    @Override
+    public void dispose() {
+        Log.d(TAG, "UvcCapturer.dispose");
+        if (mUvcStrategy != null) mUvcStrategy.stopPreview();
+    }
+
+    @Override
+    public boolean isScreencast() {
+        return false;
+    }
+
+    private CameraRequest getCameraRequest() {
+        return new CameraRequest
+                .Builder()
+                .setCameraId(this.deviceId)
+                // .setFrontCamera(false)
+                .setContinuousAFModel(true)
+                .setContinuousAutoModel(true)
+                .setPreviewWidth(UVC_PREVIEW_WIDTH)
+                .setPreviewHeight(UVC_PREVIEW_HEIGHT)
+                .create();
+//                .Builder()
+//                .setPreviewWidth(800) // camera preview width
+//                .setPreviewHeight(800) // camera preview height
+//                .setRenderMode(CameraRequest.RenderMode.OPENGL) // camera render mode
+//                .setDefaultRotateType(RotateType.ANGLE_0) // rotate camera image when opengl mode
+//                .setAudioSource(CameraRequest.AudioSource.SOURCE_AUTO) // set audio source
+//                .setPreviewFormat(CameraRequest.PreviewFormat.FORMAT_YUYV) // set preview format, MJPEG recommended
+//                .setAspectRatioShow(true) // aspect render,default is true
+//                .setCaptureRawImage(false) // capture raw image picture when opengl mode
+//                .setRawPreviewData(false)  // preview raw image when opengl mode
+//                .create();
+    }
+
+}
\ No newline at end of file
diff --git a/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java b/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
new file mode 100644
index 0000000..d2db1cf
--- /dev/null
+++ b/node_modules/react-native-webrtc/android/src/main/java/com/oney/WebRTCModule/UvcEnumerator.java
@@ -0,0 +1,65 @@
+package com.oney.WebRTCModule;
+
+import android.content.Context;
+import android.hardware.Camera;
+import android.hardware.usb.UsbDevice;
+import android.util.Log;
+import android.widget.Toast;
+
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.ReadableArray;
+import com.facebook.react.bridge.WritableArray;
+import com.facebook.react.bridge.WritableMap;
+import com.jiangdg.ausbc.camera.CameraUvcStrategy;
+
+import org.webrtc.SurfaceViewRenderer;
+
+import java.util.List;
+
+public class UvcEnumerator {
+
+  private static final String TAG = WebRTCModule.TAG;
+
+  CameraUvcStrategy mUvcStrategy;
+
+  public UvcEnumerator(CameraUvcStrategy cameraUvcStrategy) {
+    mUvcStrategy = cameraUvcStrategy;
+  }
+
+  ReadableArray enumerateDevices() {
+    WritableArray array = Arguments.createArray();
+
+    try {
+      List<UsbDevice> usbDevices = mUvcStrategy.getUsbDeviceList(null);
+
+      if (usbDevices != null) {
+
+        if (!usbDevices.isEmpty()) {
+
+          for (int i = 0; i < usbDevices.size(); i++) {
+            UsbDevice dev = usbDevices.get(i);
+
+            WritableMap params = Arguments.createMap();
+
+            params.putString("facing", "environment");
+            params.putString("deviceId", "uvc:" + dev.getDeviceId());
+            params.putString("groupId", "");
+            params.putString("label", dev.getProductName() != null ? dev.getProductName() : dev.getDeviceName());
+            params.putString("kind", "videoinput");
+
+            Log.d(TAG, "UvcEnumerator.enumerateDevices: " + params);
+
+            array.pushMap(params);
+          }
+
+        }
+
+      }
+
+    } catch (Throwable tr) {
+      Log.d(TAG, "Error on UvcEnumerator.enumerateDevices", tr);
+    }
+
+    return array;
+  }
+}
\ No newline at end of file
  ```
</details>

imamabdulazis:
> Oh shoot, I think I misunderstood @saghul's comment and was a little unclear myself.
> 
> There's a permission popup specific to the usb device that shows up when you want to use it (lasts until the usb camera is unplugged), separate from the existing permissions. It's needed even if you have the camera/mic permissions.
> 
> This is correctly not shown when other cameras are chosen.
> 

I did the test with example project too, and the permission doesn't come up.
https://github.com/react-native-webrtc/react-native-webrtc/tree/master/examples/GumTestApp

saghul:
@tommiilmonen shouldn't a permission dialog pop up? Is anything missing from the sample app perhaps?

saqibkafeel:
why no one is sharing the react native code to enable the UVC from that side directly and pass UVC camera stream to server for proper video call feature?
please share proper react native code to enable the Usb/UVC webcam 
thanks

saghul:
It's enabled by default. Unless you are in Abdroid 14, where it's disabled because it crashes.

When you do getUserMedia UVC cameras will be!enumerated.

saqibkafeel:
> It's enabled by default. Unless you are in Abdroid 14, where it's disabled because it crashes.
> 
> When you do getUserMedia UVC cameras will be!enumerated.

ok provide the react native code how i can enable the usb camera instead android builtin cam , and pass it to server for proper video call,  i am using it on android 9

saghul:
I am not going to do that, sorry. Look into the WebRTC APIs for enumerating video devices (you'll see both there) and then using one of them.

This plugin is just a WebRTC implementation for RN.

saqibkafeel:
> I am not going to do that, sorry. Look into the WebRTC APIs for enumerating video devices (you'll see both there) and then using one of them.
> 
> This plugin is just a WebRTC implementation for RN.

ok  you just post the react native code to enable usb/uvc camera instead built-in android camera, fot your provided example code is not opening my usb camera which loading in other android app for testing like this [app]
(https://play.google.com/store/apps/details?id=com.shenyaocn.android.usbcamera&hl=en&gl=US)

for now your example code using this code in react native side to getUserMedia but it is not loading my usb camera it is only loading my front face cam from builtin cam 

here is your example app RN side code:

`const [stream, setStream] = useState(null);
  const start = async () => {
    console.log('start');
    if (!stream) {
      let s;
      try {
        s = await mediaDevices.getUserMedia({ video: true });
        setStream(s);
      } catch(e) {
        console.error(e);
      }
    }
  };`

 please provide the solution accordingly!

tommiilmonen:
The sample linked above does not explicitly use the external camera

So, 
to explicitly use the UVC:

1. Enumerate devices and get the id of the uvc camera. Its in format uvc-camera:<id>
2. Explicitly call with the device id: getUserMedia({ video: { deviceId: <uvc-cameraid> } })

saqibkafeel:
> 2\. uvcId

now what will be the RN code to get the "uvcId" because each camera should be different uvcId and we need to implement it for all usb cam attached with android phone

tommiilmonen:
This should work
(await mediaDevices.enumerateDevices()).filter(source => (source?.label || '').startsWith('uvc-camera:'))

saqibkafeel:
> This should work (await mediaDevices.enumerateDevices()).filter(source => (source?.label || '').startsWith('uvc-camera:'))

no luck it just giving error 
 ERROR  [TypeError: undefined is not a function]
 
 here is my code implementation 
 
```
const [stream, setStream] = useState(null);
  const start = async () => {
    console.log('start');
    if (!stream) {
      let s;
      try {
        //s = await mediaDevices.getUserMedia({ video: true });
        s = (await mediaDevices.enumerateDevices()).filter(source => (source?.label || '').startsWith('uvc-camera:'));
       
        setStream(s);
      } catch(e) {
        console.error(e);
      }
    }
  };
```

saqibkafeel:
now i managed to enable the permission and detect my usb camera into example app but stream.toUrl() is showing empty screen after set deviceid to my usb cam

WhatsApp Image 2024-03-24 at 6 13 09 PM

here is the full RN side code:

```


import React, {useState} from 'react';
import {
  Button,
  SafeAreaView,
  StyleSheet,
  ScrollView,
  View,
  Text,
  StatusBar,
} from 'react-native';
import { Colors } from 'react-native/Libraries/NewAppScreen';
import { mediaDevices, RTCView } from 'react-native-webrtc';

const App: () => React$Node = () => {
  const [stream, setStream] = useState(null);
  const start = async () => {
    console.log('start');
   
    if (!stream) {
      var s = null;
      try {
        
        let devices = await mediaDevices.enumerateDevices();   
        console.log(devices); // the result of this log all devices array attached in this thread
        s = await mediaDevices.getUserMedia({video: { deviceId: devices[2].deviceId } });   
        console.log(s);
         console.log(s._tracks[0]);
        // console.log(s.toURL());
         setStream(s);
      } catch(e) {
        console.log('error');
        console.error(e);
      }
    }
  };
  const stop = () => {
    console.log('stop');
    if (stream) {
      stream.release();
      setStream(null);
    }
  };
  return (
    <>
      <StatusBar barStyle="dark-content" />
      <SafeAreaView style={styles.body}>
      {
        stream &&
          <RTCView
            streamURL={stream && stream.toURL()}
            style={styles.stream} 
            />
      }
        <View
          style={styles.footer}>
          <Button
            title = "Start"
            onPress = {start} />
          <Button
            title = "Stop"
            onPress = {stop} />
        </View>
      </SafeAreaView>
    </>
  );
};

const styles = StyleSheet.create({
  body: {
    backgroundColor: Colors.white,
    ...StyleSheet.absoluteFill
  },
  stream: {
    flex: 1
  },
  footer: {
    backgroundColor: Colors.lighter,
    position: 'absolute',
    bottom: 0,
    left: 0,
    right: 0
  },
});

export default App;
```

and here is the response of all enumerated devices detected including my usb cam

```
[
    {
        "deviceId": "0",
        "facing": "environment",
        "groupId": "",
        "kind": "videoinput",
        "label": "0"
    },
    {
        "deviceId": "1",
        "facing": "front",
        "groupId": "",
        "kind": "videoinput",
        "label": "1"
    },
    {
        "deviceId": "2",
        "facing": "front",
        "groupId": "",
        "kind": "videoinput",
        "label": "uvc-camera:/dev/bus/usb/001/051"
    },
    {
        "deviceId": "audio-1",
        "groupId": "",
        "kind": "audioinput",
        "label": "Audio"
    }
]
```

when i am setting deviceId to 0 and 1 the stream is appearing instead of white screen but when i am setting my device id to usb camera the stream.toUrl() is not empty but the screen is white stream is not showing!

can you guys help what i did wrong here? why the usb camera stream is not visible ? i have tried to set width and height in video tag but no luck ! 

also in enumarated devices array my showing facing:"front" but set it to stream the facingMode is showing "User"

here is the result after set the usb camera in stream

`{"_id": "46878239-152e-4ba1-b58c-3579cebef7cd", "_reactTag": "46878239-152e-4ba1-b58c-3579cebef7cd", "_tracks": [{"_constraints": [Object], "_enabled": true, "_muted": false, "_peerConnectionId": undefined, "_readyState": "live", "_settings": [Object], "id": "ee9429af-10a2-4530-93ff-4d7bb557ab74", "kind": "video", "label": "", "remote": false}]}`

and here is constraints and setting expansion

`{"_constraints": {"deviceId": "2", "facingMode": "user", "frameRate": 30, "height": 720, "width": 1280}, "_enabled": true, "_muted": false, "_peerConnectionId": undefined, "_readyState": "live", "_settings": {"frameRate": 30, "height": 720, "width": 1280}, "id": "ee9429af-10a2-4530-93ff-4d7bb557ab74", "kind": "video", "label": "", "remote": false}
`
as you can see the "facingMode": "user" is showing here ! any fix i can here i have tried to set it via set attribute but it is setting to environment but not to front( also on environment there is no luck)

saqibkafeel:
> tommiilmonen
> saghul
[saghul](https://github.com/saghul)
[tommiilmonen](https://github.com/tommiilmonen)
any update on this issue ?

saghul:
> can you guys help what i did wrong here? why the usb camera stream is not visible ? i have tried to set width and height in video tag but no luck !

Please share adb logcat. Are you using Android 14 by any chance?

> also in enumarated devices array my showing facing:"front" but set it to stream the facingMode is showing "User"

The UVC cameras don't really have a facingMode, so they use the default. You can ignore that.

cristiano-linvix:
Hello. I hope everyone is well.
In my case, I do it this way below.
But, i have a problem with audio.

The change you need to implement is to get the device id which is UVC.
In some cases, the camera resolution needs to be configured so that it locates the camera correctly.

```
            // get devices
            const devices: MediaDeviceInfo[] = (await mediaDevices.enumerateDevices()) as MediaDeviceInfo[];

            // check if you have a camera or video device
            HasCamera.current = devices.filter((d) => d.kind === 'videoinput').length > 0;
            HasMicrophone.current = devices.filter((d) => d.kind === 'audioinput').length > 0;

            if (!HasMicrophone.current) {
                // console.warn('mic not found');
            }

            if (!HasCamera.current) {
                // console.warn('camera not found');
            }

            // get stream
            LocalStream.current = await mediaDevices.getUserMedia({
                // !in version 118+ of webrtc that implements uvc, the audio breaks in the stream
                // !it does not accept the camera and ends up not closing the connection with the peer
                // video: HasCamera.current
                //     ? {
                //           width: 640,
                //           height: 480,
                //       }
                //     : false,
                video: false,
                audio: HasMicrophone.current
                    ? {
                          echoCancellation: true,
                          autoGainControl: true,
                          noiseSuppression: true,
                      }
                    : false,
            });
```

saghul:
UVC is only used for cameras, I don't see how that is related to audio.

tommiilmonen:
I can add working sample add code fragment here thursday i think.

saqibkafeel:
> Hello. I hope everyone is well. In my case, I do it this way below. But, i have a problem with audio.
> 
> The change you need to implement is to get the device id which is UVC. In some cases, the camera resolution needs to be configured so that it locates the camera correctly.
> 
> ```
>             // get devices
>             const devices: MediaDeviceInfo[] = (await mediaDevices.enumerateDevices()) as MediaDeviceInfo[];
> 
>             // check if you have a camera or video device
>             HasCamera.current = devices.filter((d) => d.kind === 'videoinput').length > 0;
>             HasMicrophone.current = devices.filter((d) => d.kind === 'audioinput').length > 0;
> 
>             if (!HasMicrophone.current) {
>                 // console.warn('mic not found');
>             }
> 
>             if (!HasCamera.current) {
>                 // console.warn('camera not found');
>             }
> 
>             // get stream
>             LocalStream.current = await mediaDevices.getUserMedia({
>                 // !in version 118+ of webrtc that implements uvc, the audio breaks in the stream
>                 // !it does not accept the camera and ends up not closing the connection with the peer
>                 // video: HasCamera.current
>                 //     ? {
>                 //           width: 640,
>                 //           height: 480,
>                 //       }
>                 //     : false,
>                 video: false,
>                 audio: HasMicrophone.current
>                     ? {
>                           echoCancellation: true,
>                           autoGainControl: true,
>                           noiseSuppression: true,
>                       }
>                     : false,
>             });
> ```

i am getting the usb cam permission to access but the thing is the stream is not loading in RTCView like it is loading for builtin android camera;s, for usb cam it is showing only white blank screen i need video load in RTCView

saqibkafeel:
> I can add working sample add code fragment here thursday i think.

it should be in Reactive Native will be helpfull

cristiano-linvix:
> UVC is only used for cameras, I don't see how that is related to audio.

I think the same, however, if I activate the UVC camera, I have a problem with the device's audio and the peer rejects the connection.
But, this has nothing to do with this topic, I will open an issue when I investigate further and if it is really necessary.

saqibkafeel:
> now i managed to enable the permission and detect my usb camera into example app but stream.toUrl() is showing empty screen after set deviceid to my usb cam
> 
> WhatsApp Image 2024-03-24 at 6 13 09 PM
> 
> here is the full RN side code:
> 
> ```
> 
> 
> import React, {useState} from 'react';
> import {
>   Button,
>   SafeAreaView,
>   StyleSheet,
>   ScrollView,
>   View,
>   Text,
>   StatusBar,
> } from 'react-native';
> import { Colors } from 'react-native/Libraries/NewAppScreen';
> import { mediaDevices, RTCView } from 'react-native-webrtc';
> 
> const App: () => React$Node = () => {
>   const [stream, setStream] = useState(null);
>   const start = async () => {
>     console.log('start');
>    
>     if (!stream) {
>       var s = null;
>       try {
>         
>         let devices = await mediaDevices.enumerateDevices();   
>         console.log(devices); // the result of this log all devices array attached in this thread
>         s = await mediaDevices.getUserMedia({video: { deviceId: devices[2].deviceId } });   
>         console.log(s);
>          console.log(s._tracks[0]);
>         // console.log(s.toURL());
>          setStream(s);
>       } catch(e) {
>         console.log('error');
>         console.error(e);
>       }
>     }
>   };
>   const stop = () => {
>     console.log('stop');
>     if (stream) {
>       stream.release();
>       setStream(null);
>     }
>   };
>   return (
>     <>
>       <StatusBar barStyle="dark-content" />
>       <SafeAreaView style={styles.body}>
>       {
>         stream &&
>           <RTCView
>             streamURL={stream && stream.toURL()}
>             style={styles.stream} 
>             />
>       }
>         <View
>           style={styles.footer}>
>           <Button
>             title = "Start"
>             onPress = {start} />
>           <Button
>             title = "Stop"
>             onPress = {stop} />
>         </View>
>       </SafeAreaView>
>     </>
>   );
> };
> 
> const styles = StyleSheet.create({
>   body: {
>     backgroundColor: Colors.white,
>     ...StyleSheet.absoluteFill
>   },
>   stream: {
>     flex: 1
>   },
>   footer: {
>     backgroundColor: Colors.lighter,
>     position: 'absolute',
>     bottom: 0,
>     left: 0,
>     right: 0
>   },
> });
> 
> export default App;
> ```
> 
> and here is the response of all enumerated devices detected including my usb cam
> 
> ```
> [
>     {
>         "deviceId": "0",
>         "facing": "environment",
>         "groupId": "",
>         "kind": "videoinput",
>         "label": "0"
>     },
>     {
>         "deviceId": "1",
>         "facing": "front",
>         "groupId": "",
>         "kind": "videoinput",
>         "label": "1"
>     },
>     {
>         "deviceId": "2",
>         "facing": "front",
>         "groupId": "",
>         "kind": "videoinput",
>         "label": "uvc-camera:/dev/bus/usb/001/051"
>     },
>     {
>         "deviceId": "audio-1",
>         "groupId": "",
>         "kind": "audioinput",
>         "label": "Audio"
>     }
> ]
> ```
> 
> when i am setting deviceId to 0 and 1 the stream is appearing instead of white screen but when i am setting my device id to usb camera the stream.toUrl() is not empty but the screen is white stream is not showing!
> 
> can you guys help what i did wrong here? why the usb camera stream is not visible ? i have tried to set width and height in video tag but no luck !
> 
> also in enumarated devices array my showing facing:"front" but set it to stream the facingMode is showing "User"
> 
> here is the result after set the usb camera in stream
> 
> `{"_id": "46878239-152e-4ba1-b58c-3579cebef7cd", "_reactTag": "46878239-152e-4ba1-b58c-3579cebef7cd", "_tracks": [{"_constraints": [Object], "_enabled": true, "_muted": false, "_peerConnectionId": undefined, "_readyState": "live", "_settings": [Object], "id": "ee9429af-10a2-4530-93ff-4d7bb557ab74", "kind": "video", "label": "", "remote": false}]}`
> 
> and here is constraints and setting expansion
> 
> `{"_constraints": {"deviceId": "2", "facingMode": "user", "frameRate": 30, "height": 720, "width": 1280}, "_enabled": true, "_muted": false, "_peerConnectionId": undefined, "_readyState": "live", "_settings": {"frameRate": 30, "height": 720, "width": 1280}, "id": "ee9429af-10a2-4530-93ff-4d7bb557ab74", "kind": "video", "label": "", "remote": false} ` as you can see the "facingMode": "user" is showing here ! any fix i can here i have tried to set it via set attribute but it is setting to environment but not to front( also on environment there is no luck)

any update on this issue ? when i am setting UVC stream in RTCView it is showing blank screen

tommiilmonen:
I literally told you i can send the sample thursday....

saghul:
If you keep harassing people like this you'll be blocked from this org.

Nobody has the obligation to help you.

saqibkafeel:
> I literally told you i can send the sample thursday....

oh actually i was on GMT+5 , bro your help will be helpfull , i need to be implement it in RN side to load UVC camera stream to RTCView like builtin cam
Thanks in advance.

saqibkafeel:
> harassing

where is the word "harassing" coming? bro i am doing a project which deadline is passed, and i need to update the client on regular basis and in reply client share me many discussions in issues by you guys where you are saying UVC cam is fully functional now what i reply to client ?

saghul:
That is your problem, not ours.

saqibkafeel:
> That is your problem, not ours.

don't be rude like a kid, if you guys done implementation of UVC camera which you are claiming in your each and every conversation so make it proper implemented in your example where the UVC camera stream showing blank screen! 
Thanks.

8BallBomBom:
@saqibkafeel You are welcome to submit a PR with more helpful guides and examples.
Any further discussion can be taken over [here](https://react-native-webrtc.discourse.group/) as long as you can actually  control your behaviour.

saghul:
Don't commit to deadlines you can't meet. It's that simple.

Software is software and may have bugs, that's life.

Constantly asking for updates when you've been told when you'd get them is not acceptable.

PR #1460, opened 2023-10-16
Fix onxxx event handlers not being defined after TS refactor

saghul:


8BallBomBom:
I'll be blunt, this one really kicked me in the arse, was thinking something was wrong with my dev setup.
Been switching between node versions a lot recently using `nvm` which resulted in some weirdness with executables not being linked properly and not switching when they should, absolute headache...
But on top of that something else weird happened where packages were not updating correctly...
and even worse, working on packages that decide to change git author details...

The conclusion though 🤔 
I never experienced the issues which kept me running around in circles due to not using the `onXXX` methods.
Way back I changed everything over to `addEventListener` in the docs and all of my projects.
Big oops, now they are included in testing 😅

robertbarclay:
ETA on when this might ship? also seeing issues. We have been patching in the EventTarget already, but this seems cleaner of a fix here. 🎉

8BallBomBom:
Best bet would be to use `111.0.3` until the next update which will most likely go live soon 👍🏻

PR #1462, opened 2023-10-17
M118

saghul:
- ~~Android build is still missing~~ Done!

### DIFF
--- a/ios/RCTWebRTC/WebRTCModule+Transceivers.m
+++ b/ios/RCTWebRTC/WebRTCModule+Transceivers.m
@@ -3,6 +3,7 @@
 #import <React/RCTBridge.h>
 #import <React/RCTBridgeModule.h>
 
+#import <WebRTC/RTCRtpCodecCapability.h>
 #import <WebRTC/RTCRtpReceiver.h>
 #import <WebRTC/RTCRtpSender.h>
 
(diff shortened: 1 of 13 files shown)


ios/RCTWebRTC/WebRTCModule+Transceivers.m
```diff
@@ -178,6 +185,72 @@ @implementation WebRTCModule (Transceivers)
...
+        for (NSDictionary *entry in availableCodecs) {
+            NSDictionary *availableCodecDict = [entry objectForKey:@"dict"];
+            if ([codecDict isEqualToDictionary:availableCodecDict]) {
+                [codecsToSet addObject:[entry objectForKey:@"codec"]];
+                break;
+            }
+        }
+    }
+
+    transceiver.codecPreferences = codecsToSet;
```
saghul, line 250: @davidliu Looks like we need to use setCodecPreferences here, but it takes RTCRtpCapabilities so some of this conversion is not necessary I think?

davidliu: Yeah, I think I might have changed the code compared to when the initial draft pr was made, so I'll need to recheck this commit. The M118 build is available to try out for ios?

saghul: Yep feel free to add commits to this PR!

saghul:
Android is ready for testing.

ios/RCTWebRTC/WebRTCModule+Transceivers.m
davidliu: `setCodecPreferences` takes RTCRtpCodecCapabilities rather than just RTCRtpCapabilities, so the conversion is needed. Fixed up the usage, so it compiles now and should be working as intended.

saghul:
Thanks @davidliu I'll begin testing tomorrow!

saghul:
Sorry for the delay y'all, let's do this!

PR #1468, opened 2023-10-31
Change event handlers over to getters/setters

8BallBomBom:
Will need some testing as this was thrown together 👍🏻 
Might be a good idea moving things up to the  constructor 🤔

### DIFF
--- a/src/MediaDevices.ts
+++ b/src/MediaDevices.ts
@@ -1,4 +1,4 @@
-import { EventTarget, Event, defineEventAttribute } from 'event-target-shim';
+import { EventTarget, Event, getEventAttributeValue, setEventAttributeValue } from 'event-target-shim';
 import { NativeModules } from 'react-native';
 
 import getDisplayMedia from './getDisplayMedia';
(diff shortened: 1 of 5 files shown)


saghul:
The docs suggest stronger type annotations which I see you had and then removed. What happened there?

8BallBomBom:
I can add them back but pretty much a whole bunch of errors about the amount of arguments not being acceptable 🤔

saghul:
I see. If this fixes it, let's go with it and improve it later if necessary.

8BallBomBom:
No worries, will run some tests and ensure the events are firing as expected but as for the error others have been experiencing, that is a bit of a pickle, not managed to reproduce 🤔

saghul:
Since it looks like this might not be necessary after all, shall we close it and revisit if / when needed?

8BallBomBom:
Best bet, close it, revisit if and when needed due to package rot/updates over time 😅

saghul:
> Maybe once the deprecated code has actually been removed?

Unlikely to happen anytime soon, so sure :-)

8BallBomBom:
Never really liked getters/setters anyway 🤣 all that extra code seems kinda dirty.

Issue #1476, opened 2023-11-11
WebRTCModule.peerConnectionInit returns false when using TURN server with iceServers key "urls"


xaleel:
I spent many days trying to debug this issue, now I feel silly it's this simple, but I don't believe this is the expected behavior.

#### Expected Behavior: 
`peerConnection` initializes

#### Observed Behavior: 
App crashes with error message `Failed to initialize PeerConnection, check the native logs!`. 

#### Steps to reproduce the issue:
```typescript
const peerConstraints = {
	iceServers: [
		{
			url: "stun:stun.l.google.com:19302",
			// urls: "stun:stun.l.google.com:19302",
			// both work fine
		},
		{
			urls: "TURN:freeturn.net:3478", // this fails
			// urls: ["TURN:freeturn.net:3478"], // this also fails
			// url: "TURN:freeturn.net:3478", // (without the 's') this works
			// url: ["TURN:freeturn.net:3478"], // this also works
			username: "free",
			credential: "free",
		},
	],
	iceTransportPolicy: "all",
	rtcpMuxPolicy: "negotiate"
};
export default function CameraComponent(props: any) {
	const [peerConnection, setPeerConnection] = useState(new RTCPeerConnection(peerConstraints));
	...
}
```
According to [w3 docs](https://www.w3.org/TR/webrtc/#dom-rtciceserver) ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/RTCPeerConnection#iceservers)) the `RTCIceServer` Dictionary's key should be "urls". Using "url" should still be supported, in my opinion.

#### Platform Information  
* **React Native Version**: `0.72.6` 
* **WebRTC Module Version**: `^111.0.6`
* **Platform OS + Version**: Android 9, API 28, Samsung SM-G950F, **expo** `~49.0.18`

xaleel:
Additional info:
With the configuration that fails, I added debug lines to `node_modules\react-native-webrtc\src\RTCPeerConnection.ts`
```typescript
export default class RTCPeerConnection extends EventTarget<RTCPeerConnectionEventMap> {
   ...
    constructor(configuration) {
        super();

        this._pcId = nextPeerConnectionId++;

        console.log("this._pcId", this._pcId);
        let peerCon = WebRTCModule.peerConnectionInit(configuration, this._pcId);
        console.log("peerCon", peerCon);

        if (!peerCon) {
[...]

PR #1477, opened 2023-11-12
pc: sanitize ICE servers

saghul:
- Make sure the `urls` property is used, while supporting the deprecated `url`.
- Make sure all servers are lower-case since the C++ layer does a case insensitive parse.

Fixes: https://github.com/react-native-webrtc/react-native-webrtc/issues/1476

### DIFF
--- a/src/RTCPeerConnection.ts
+++ b/src/RTCPeerConnection.ts
@@ -45,6 +45,21 @@ type RTCDataChannelInit = {
     id?: number
 };
 
+type RTCIceServer = {
+    credential?: string,
+    url?: string, // Deprecated.
+    urls?: string | string[],
+    username?: string
+};
+
+type RTCConfiguration = {
+    bundlePolicy?: 'balanced' | 'max-compat' | 'max-bundle',
+    iceCandidatePoolSize?: number,
+    iceServers?: RTCIceServer[],
+    iceTransportPolicy?: 'all' | 'relay',
+    rtcpMuxPolicy?: 'negotiate' | 'require'
+};
+
 type RTCPeerConnectionEventMap = {
     connectionstatechange: Event<'connectionstatechange'>
     icecandidate: RTCIceCandidateEvent<'icecandidate'>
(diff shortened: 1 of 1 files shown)


8BallBomBom (approved):
Looks good to me, testing without hitches in general terms 👍🏻 
Seems to be sanitising as advertised.

Only hitch i'm getting are due to providing wonky ice server connection info but that is expected behavior.
There is a complete halt though if providing wonky params on peer creation but that can be ignored or addressed in another pr 😸 no biggie.

Then again maybe we could go down a more verbose route for everything when any incorrect params are used, future improvement?

saghul:
What kind of wonky input makes it go boom?

8BallBomBom:
Not related to the `iceServers` but related to the initialization of peers in general.
You can cause the entire initialization to halt just as simple as that.
No errors or anything.

```javascript
let peer = new RTCPeerConnection( [] );
```

But then we are talking if doing things wrong 😨
Maybe we could look at sanitising and building in some extra type checks on everything?
As said, future improvement, might help some people along the way 😸

8BallBomBom:
> That's a weird one. That should fail the type check...

Exactly my thought 🤔

PrakashkumarM:
Hi guys.
I'm experiencing the same issue even after applying the latest fix: "react-native-webrtc": "github:react-native-webrtc/react-native-webrtc#pull/1477/head". Interestingly, the problem only occurs when connected to the debugger. I appreciate any suggestions or solutions you might have.

saghul:
What object are you passing to the PC constructor? Are you using the Chrome debugger?

PrakashkumarM:
This is my peer connection setup.
const ICE_SERVERS = [
  {
    urls: "turn:turn.call.test.com:3500",
    username: "user",
    credential: "password",
  },
  // Add more ICE servers as needed
];

  // PeerConnection setup
  const peerConnection = useRef(
    new RTCPeerConnection({
      sdpSemantics: "unified-plan",
      offerToReceiveVideo: true,
      offerToReceiveAudio: true,
      iceServers: ICE_SERVERS,
    })
  );
  
  Yes I'm using chrome debugger

saghul:
> sdpSemantics: "unified-plan",

This is not necessary.



> offerToReceiveVideo: true,
> offerToReceiveAudio: true,

This is not used by the PC constraints, it's something you may pass to createLocalDescription.



> Yes I'm using chrome debugger

That will not work because we use some sync methods. You need to disable it or the code won't work at all, which I suspect is the reson for your problem here.

PrakashkumarM:
Kindly suggest an alternative debugger. I'm currently using React Native Debugger.

saghul:
Flipper, or judiciously placed console.log statements 😅

PR #1484, opened 2023-12-04
Replaced deprecated `AVCaptureDeviceTypeBuiltInMicrophone` with `AVCaptureDeviceTypeMicrophone`

curiouscosmos:
`AVCaptureDeviceTypeBuiltInMicrophone` is deprecated  https://developer.apple.com/documentation/avfoundation/avcapturedevicetypebuiltinmicrophone

`AVCaptureDeviceTypeMicrophone` is replacement as per the docs.


8BallBomBom:
Thanks for the contribution 👍🏻 

Not entirely sure about this one as we do still have slight older iOS support and this would make things iOS 17+ as a default requirement far as i can tell 🤔

saghul:
The new identifier is iOS >= 17 only, so we cannot have that, sorry. iOS 12 is still supported.

PR #1520, opened 2024-03-14
Fix #1518 - Support null values in either sdpMid or sdpMLineIndex (but not both)

### LINKED ISSUE #518: Integrate with Openvidu
Hi.
for web using openvidu but I have difficulty connecting openvidu with react-native-webrtc.

I can publish videos from mobile(react-native-webrtc) to the web (openvidu), but I failed to receive videos from the web to mobile.

can you help me to fix the code?

https://github.com/randymuhroji/react-native-webrtc-openvidu

### LINKED ISSUE #1518: `sdpMid` and `sdpMLineIndex` are  both required although they are optional/nullable in the spec
#### Expected Behavior:
`RTCIceCandidate` should allow initiation with either `sdpMid`,  or `sdpMLineIndex`
Spec defines them as nullable:
https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate#instance_properties

Only when both `sdpMid` &  `sdpMLineIndex` are null should an error be raised:
image

#### Observed Behavior:
When trying to connect to a signaling server which doesn't send 'sdpMid', but does send `sdpMLineIndex` as part of the ice candidate message, there is the following error:
```
cannot parse incoming message from signaling server [TypeError: `sdpMLineIndex` and `sdpMid` must not null]
```

#### Steps to reproduce the issue:
1) Connect to a signaling server not necessarily sending 'sdpMid' field, such as gstreamer webrtcsink's signaling server
2) See the issue

#### Platform Information  
* **React Native Version**: React Native 0.73.1
* **WebRTC Module Version**: 118.0.2
* **Platform OS + Version**: Android 12

omerts:
Fixes #1518 
I have tested this on Android, but not on iOS as I have no access to a mac.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -1275,16 +1275,16 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
                 return;
             }
 
-            if (!(candidateMap.hasKey("sdpMid") && candidateMap.hasKey("sdpMLineIndex")
-                        && candidateMap.hasKey("sdpMid"))) {
+            if (!candidateMap.hasKey("sdpMid") && !candidateMap.hasKey("sdpMLineIndex")) {
                 promise.reject("E_TYPE_ERROR", "Invalid argument");
                 return;
             }
 
-            IceCandidate candidate = new IceCandidate(candidateMap.getString("sdpMid"),
-                    candidateMap.getInt("sdpMLineIndex"),
-                    candidateMap.getString("candidate"));
-
+            IceCandidate candidate = new IceCandidate(
+                candidateMap.hasKey("sdpMid") && !candidateMap.isNull("sdpMid") ? candidateMap.getString("sdpMid")  : "",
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -1275,16 +1275,16 @@ public void peerConnectionAddICECandidate(int pcId, ReadableMap candidateMap, Pr
                 return;
             }
 
-            if (!(candidateMap.hasKey("sdpMid") && candidateMap.hasKey("sdpMLineIndex")
-                        && candidateMap.hasKey("sdpMid"))) {
+            if (!candidateMap.hasKey("sdpMid") && !candidateMap.hasKey("sdpMLineIndex")) {
                 promise.reject("E_TYPE_ERROR", "Invalid argument");
```
8BallBomBom, line 1279: Might be a bit better to have a more specific error like everywhere else 🤔

8BallBomBom (changes requested):
Could you please run the provided code format script and submit the changes.
Looks good but will need testing.

saghul:
Can you check if the native code allows this in both platforms?

8BallBomBom:
Haven't currently got any Apple devices but will try and run some tests shortly 👍🏻

saghul (approved):

PR #1525, opened 2024-03-28
android, ios: remove track when removed from peerconnection

davidliu:
Not sure if this was an intentional omission, but I noticed that a track is kept around after it has been removed from the peer connection. This was causing issues when the track with the same id was added back to the peer connection, as any calls on the track at the JS layer would be pointed at the stale track. Are there any issues with removing the track here?

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -483,10 +483,7 @@ class PeerConnectionObserver implements PeerConnection.Observer {
             params.putInt("pcId", this.id);
             params.putString("receiverId", receiver.id());
 
-            MediaStreamTrack track = remoteTracks.remove(receiver.id());
-            if (track != null) {
-                track.dispose();
-            }
+            remoteTracks.remove(receiver.id());
 
             webRTCModule.sendEvent("peerConnectionOnRemoveTrack", params);
         });
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -483,6 +483,11 @@ class PeerConnectionObserver implements PeerConnection.Observer {
             params.putInt("pcId", this.id);
             params.putString("receiverId", receiver.id());
 
+            MediaStreamTrack track = remoteTracks.remove(receiver.id());
+            if (track != null) {
+                track.dispose();
+            }
+
             webRTCModule.sendEvent("peerConnectionOnRemoveTrack", params);
         });
     };
--- a/ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
@@ -922,6 +922,8 @@ RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(peerConnectionRemoveTrack
         params[@"pcId"] = peerConnection.reactTag;
         params[@"receiverId"] = rtpReceiver.receiverId;
 
+        [peerConnection.remoteTracks removeObjectForKey:rtpReceiver.receiverId];
+
         [self sendEventWithName:kEventPeerConnectionOnRemoveTrack body:params];
     });
 }


saghul:
Oh, so we remove it from JS but not native? That does sound like an overright.

8BallBomBom:
Ouch that is a bit of a big oversight, might also explain some of the weird peer crashes 🤔

8BallBomBom (approved):

8BallBomBom:
Forgot to get back on this but tested without issue 👍🏻

android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
```diff
@@ -483,6 +483,8 @@ public void onRemoveTrack(RtpReceiver receiver) {
             params.putInt("pcId", this.id);
             params.putString("receiverId", receiver.id());
 
+            remoteTracks.remove(receiver.id());
```
carlosmg95, line 486: I don't know what I am doing wrong, but after this change, when I pause a consumer and then I resume it, the track is black.

saghul: What do pause and resume do under the hood?

carlosmg95: They call `setRemoteDescription` from RTCPeerConnection, so `onRemoveTrack` is triggered when the consumer is paused and `onAddTrack` is triggered when the consumer is resumed. But the track is black in RTCView.

saghul: I'm not sure how this chance would affect that. When a new track comes it should be fresh, no?

carlosmg95: It should be, but I have checked that if I comment this line, it works ok. I will try to pause/resume a track in a simpler example.

saghul: Ping @davidliu

8BallBomBom: Bit of an unusual one, some edge cases we haven't tested 🤔 
Hopefully understanding this correctly but you are referring to a mediasoup consumer?
and then pausing/resuming the consumer/track is causing `onRemoveTrack` and `onAddTrack` to be triggered.

carlosmg95: Yeah, it is a mediasoup consumer. The way is:
#### From mediasoup-client
1. I call `consumer.pause()`.
2. It emits a `@pause` event.
3. The event is listened by Transport and call `this.pausePendingConsumers();` that calls `this._handler.pauseReceiving(localIds);`.
4. In this case, the handle is _ReactNativeUnifiedPlan_ that, among other things, calls `await this._pc.setRemoteDescription(offer);`.
#### From react-native-webrtc
5. The new remote description is set.
6. It triggers `onRemoveTrack` because, according to the doc: onRemoveTrack is "Triggered when a previously added remote track is removed by the peer, as a result of setRemoteDescription.".

The way after `consumer.resume()` is very similar.

Without the new line `remove(receiver.id());`, when I call `consumer.resume()` I see the video again, but with that line, I cannot see it, it shows a black video.

8BallBomBom: Alrighty 🤔 Got to get my head around this for a second.
Not experiencing any issues with this PR but then I'm not doing as you are doing @carlosmg95 😅 
Am i reading this wrong?

Isn't this.
```java
remoteTracks.remove(receiver.id());
```
Meant to be this.
```java
final MediaStreamTrack track = receiver.track();
remoteTracks.remove(track.id());
```

saghul: Very sharp eye, you are correct.

8BallBomBom: Thought something was off 😅
My bad, have a habit of pulling, testing, if nothing breaks then the all clear is given.
Will take a closer look in future 😉

carlosmg95: Thank you guys for listening to me, and all the people who work in this library. I use it every day.

carlosmg95: In my case, they have the same value.

8BallBomBom: Wasn't exactly sure about that one, not something I've dived too deep into 😅

PR #1531, opened 2024-04-04
android: don't declare foreground service permissions

saghul:
People who are not interested in using the feature will be asked about it in the Play Console otherwise.

Document what apps need to do in order to get screen-sharing working on Android.


santhoshvai:
awesome!, thanks for this change

Issue #1533, opened 2024-04-04
Drop UVC support?


saghul:
I know, I know, we have landed it not that long ago... However:

- It's a huge thing upon close inspection, composed of multiple libraries which are shared in binary form (yes, the code is there but the aar incldues them, we'd need to fork it to do our own builds)
- It inlcudes a lot of features we don't need / want like audio support
- Looks unmaintained, based on the number of issues and the lack of a response from the maintainer
- Doesn't work on Android 14 (somewhat related to the above)
- xz

As an alternative, we could add an option for injecting a custom camera enumerator / capturer, so apps that want to use this library can continue doing so, without having to include them here.

@davidliu @8BallBomBom

8BallBomBom:
Building in the ability to specify a custom capture device for video/audio would definitely be a better path forward as it would mean any integrations would be handled by another contributor/package.

Less issues in the long run.

Would need some planning but a good way forward nevertheless 🤔

saghul:
FTR, I am observing crashes in production in Jitsi Meet:

```
Fatal Exception: java.lang.NoSuchFieldError: no "J" field "mNativePtr" in class "Lcom/jiangdg/uvc/UVCCamera;" or its superclasses
       at com.jiangdg.uvc.UVCCamera.nativeCreate(UVCCamera.java)
       at com.jiangdg.uvc.UVCCamera.<init>(UVCCamera.java:179)
       at com.jiangdg.ausbc.camera.CameraUvcStrategy.createCamera(CameraUvcStrategy.kt:132)
       at com.jiangdg.ausbc.camera.CameraUvcStrategy.startPreviewInternal(CameraUvcStrategy.kt:114)
       at com.jiangdg.ausbc.camera.ICameraStrategy.handleMessage(ICameraStrategy.kt:103)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:223)
       at android.os.HandlerThread.run(HandlerThread.java:67)
```

davidliu:
I agree, having the ability to inject camera enumerators seems like the way to go here.

saghul:
PR: https://github.com/react-native-webrtc/react-native-webrtc/pull/1534

I'll work on the API to inject external enumerators / capturers in a subsequent PR.

Also, heads up @tommiilmonen , I'm sorry it didn't quite work out.

8BallBomBom:
[...]

PR #1534, opened 2024-04-04
android: drop UVC camera support

saghul:
Ref: https://github.com/react-native-webrtc/react-native-webrtc/issues/1533


android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -141,16 +118,30 @@ private void checkMandatoryConstraints(MediaConstraints peerConstraints) {
         peerConstraints.mandatory.addAll(valid);
     }
 
+    private CameraEnumerator getCameraEnumerator() {
+        if (cameraEnumerator == null) {
+            if (Camera2Enumerator.isSupported(reactContext)) {
```
saghul, line 123: Note: the astute reader might notice we no longer try-except this. It used to throw, but not anymore, it was fixed upstream.

davidliu (approved):

PR #1541, opened 2024-04-10
android,ios: don't reject promise for getStats when peer connection doesn't exist

davidliu:
Browsers don't seem to fail for `getStats()` on closed peer connections/senders/receivers (and the spec doesn't seem to allow for rejecting it). For RTCRtpSender/Receivers, they return an empty stats report, while there's a lone entry for peer connections:

```
{
    "id": "P",
    "timestamp": 1712746990577.823,
    "type": "peer-connection",
    "dataChannelsClosed": 0,
    "dataChannelsOpened": 2
}
```

To closer align with the spec, this PR just has them all return empty maps if the native equivalent can't be found.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -255,9 +255,9 @@ class PeerConnectionObserver implements PeerConnection.Observer {
 
     public void senderGetStats(String senderId, Promise promise) {
         RtpSender targetSender = null;
-        for (RtpSender r : peerConnection.getSenders()) {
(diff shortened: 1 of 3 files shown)


saghul (approved):

8BallBomBom (approved):
Might be worth having the same warnings on the android side, seems to be a common difference across most of the project 🤔 

Something to address overall in a future pr.

Either way LGTM 😄

davidliu:
What's the RCTLog equivalent for Android? Seems like we just use regular android logging everywhere.

8BallBomBom:
No worries, just nit picking, tend mostly to leave a trail of things to get back to 😄 
This PR just reminded me of the difference in logging between platforms and how they don't match up.

PR #1543, opened 2024-04-11
compat: Add RTCVideoView to the new Interop Layer - RN 0.72+

8BallBomBom:
There is plenty to do when it comes to converting everything over to the latest RN standards but until then this will allow people to use the New Architecture in their project.

More info over [here](https://github.com/reactwg/react-native-new-architecture/discussions/135).
Will test soon 😄

### DIFF
--- a/react-native.config.js
+++ b/react-native.config.js
@@ -1,11 +1,20 @@
 'use strict';
 
+module.exports = {
+    project: {
+        android: {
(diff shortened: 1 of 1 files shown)


davidliu (approved):
Nice!

saghul (approved):
LGTM if that works!

8BallBomBom:
From what it appears, should be a simple change as I've done to get things working but so far that doesn't seem to work.
Still investigating though 🤔

PR #1545, opened 2024-04-11
ios: fix not being able to deactivate encodings

saghul:
This must have been one of the nastiest bugs to catch.

The JS layer passes the encoding parameters as JSON, which includes the "active" element set to the boolean value `true`. RN converts that to an NSDictionary, but since NSDictionary doesn't support booleans (BOOL is a C type, not an Obj-C type) the boolean values get converted to NSNumber. Assigning any NSNumber to a BOOL will always result in a true value, because it's actually an object! In order to get the actual boolean value one has to use `boolValue`: `[theNumber boolValue]`, which returns BOOL.

So yeah, if your simulcast layer suspension code is mysteriously not working on iOS... this is why :-P


8BallBomBom:
Ooooooofff good catch 😅

PR #1547, opened 2024-04-12
ios,android - remove track when removed from peer fix

8BallBomBom:
Will need testing but should make sure tracks are being removed from peer connections 😃

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
+++ b/android/src/main/java/com/oney/WebRTCModule/PeerConnectionObserver.java
@@ -483,7 +483,8 @@ class PeerConnectionObserver implements PeerConnection.Observer {
             params.putInt("pcId", this.id);
             params.putString("receiverId", receiver.id());
 
-            remoteTracks.remove(receiver.id());
+            final MediaStreamTrack track = receiver.track();
+            remoteTracks.remove(track.id());
 
             webRTCModule.sendEvent("peerConnectionOnRemoveTrack", params);
         });
--- a/ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCPeerConnection.m
@@ -927,7 +927,8 @@ RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(peerConnectionRemoveTrack
         params[@"pcId"] = peerConnection.reactTag;
         params[@"receiverId"] = rtpReceiver.receiverId;
 
-        [peerConnection.remoteTracks removeObjectForKey:rtpReceiver.receiverId];
+        RTCMediaStreamTrack *track = rtpReceiver.track;
+        [peerConnection.remoteTracks removeObjectForKey:track.trackId];
 
         [self sendEventWithName:kEventPeerConnectionOnRemoveTrack body:params];
     });


8BallBomBom:
@carlosmg95 Can you check to see if this resolves the issue you are experiencing?
Thanks for the report and info provided.

carlosmg95:
> @carlosmg95 Can you check to see if this resolves the issue you are experiencing? Thanks for the report and info provided.

I'm sorry but my issue is still there. In my case, `receiver.id()` and `track.id()` have the same value, so the final result is the same. I don't understand very well why the track is removed when the mediasoup consumer is paused. I'll check my code, the issue may be there.

Thank you very much for doing it so quickly and for this library!

8BallBomBom:
No worries, thanks for helping.
Bit stumped as to why the id values are the same 🤔 
Figured the transceiver/consumer id and the track id should be unique.

But then there is the issue you are experiencing.
We will need to investigate further 😄

davidliu:
@carlosmg95 can you debug and check if onAddTrack is called when the track is resumed?

davidliu:
@carlosmg95  can you debug and check if onAddTrack is called when you resume the track?

carlosmg95:
> @carlosmg95 can you debug and check if onAddTrack is called when you resume the track?

Yeah, it is called. And the track.id that is put in remoteTracks is the same that I had removed.

saghul (approved):

saghul:
I don't understand 2 things here: 1) why are they the same ID? 2) What does keeping that track there accomplish? Perhaps we need to revisit the ontrack logic a bit.

8BallBomBom:
Exactly to be honest I'm not sure how we should move forward on this.
We really should be removing the tracks once peer connections are disposed.
But then I'm thinking surely we shouldn't be removing the tracks when they are paused?
Feels like fishy logic, events being called when they shouldn't be.

saghul:
There is no concept of pausing at our level. The track is removed.

davidliu:
Looks like by default the receiver/sender is equal to the track id if it's a part of a transceiver that's coming in with an associated track:

Receiver: https://github.com/jitsi/webrtc/blob/0c4b9fd122265548f6b7fbb6784a88ba61c5a1cb/pc/sdp_offer_answer.cc#L3846

Sender:
https://github.com/jitsi/webrtc/blob/0c4b9fd122265548f6b7fbb6784a88ba61c5a1cb/pc/rtp_transmission_manager.cc#L154

I think the reason why I wrote it this way to use the receiver id was because I assumed the track was already removed and thus `receiver.track.id` would be unavailable. This will need some testing to make sure.

---- 

I think we might need a reproducible example for this; with my setup, tracks can be unpublished and republished with the same track id, but I'm not running into any issues with the video.

@carlosmg95 just to make sure, are you unattaching from video stream while the track is paused (removed), and then reattaching when the video track is resumed (added)? Or are you keeping the RTCVideoView set on the same stream url the entire time?

Wondering if there's some weird thing with all these intermediate layers, where having the old removed remote tracks still somehow allows it to attach to a stream id at the native layer, and then when the track comes back, the video view gets repopulated.

carlosmg95:
> Looks like by default the receiver/sender is equal to the track id if it's a part of a transceiver that's coming in with an associated track:
> 
> Receiver: https://github.com/jitsi/webrtc/blob/0c4b9fd122265548f6b7fbb6784a88ba61c5a1cb/pc/sdp_offer_answer.cc#L3846
> 
> Sender: https://github.com/jitsi/webrtc/blob/0c4b9fd122265548f6b7fbb6784a88ba61c5a1cb/pc/rtp_transmission_manager.cc#L154
> 
> I think the reason why I wrote it this way to use the receiver id was because I assumed the track was already removed and thus `receiver.track.id` would be unavailable. This will need some testing to make sure.
> 
> I think we might need a reproducible example for this; with my setup, tracks can be unpublished and republished with the same track id, but I'm not running into any issues with the video.
> 
> @carlosmg95 just to make sure, are you unattaching from video stream while the track is paused (removed), and then reattaching when the video track is resumed (added)? Or are you keeping the RTCVideoView set on the same stream url the entire time? (Or even attempting to attach while the video track is removed?)
> 
> Wondering if there's some weird thing with all these intermediate layers, where having the old removed remote tracks still somehow allows it to attach to a stream id at the native layer, and then when the track comes back, the video view gets repopulated.

In the original code, I unattach and attach because the code is like:
```
if (paused) {
  return <PausedIcon />
}
return <RTCView ... />
```
but I tried with a simpler code when the RTCView is always visible and the stream doesn't change, and the result is the same.

davidliu:
@carlosmg95 would it be possible to upload a minimal reproducible example? Would speed up debugging this issue a lot.

carlosmg95:
> @carlosmg95 would it be possible to upload a minimal reproducible example? Would speed up debugging this issue a lot.

I cannot upload my current code, but I will try to develop an example tomorrow.

8BallBomBom:
No worries, even if you can provide private access to something that would help.
Reproduction helps debug and resolve issues a lot quicker, specially them edge cases 😅

carlosmg95:
Hi,

I have made an example based on GumTestApp example. You have run `npm i mediasoup-client protoo-client random-string` and change the App.js file by:
```
/**
 * Sample React Native App
 * https://github.com/facebook/react-native
 *
 * @format
 * @flow strict-local
 */

import {Device, parseScalabilityMode} from 'mediasoup-client';
import protooClient from 'protoo-client';
import randomString from 'random-string';
import React, {useRef, useState} from 'react';
import {
  Button,
  SafeAreaView,
  StyleSheet,
  View,
  StatusBar,
  TextInput,
} from 'react-native';
import {Colors} from 'react-native/Libraries/NewAppScreen';
import {registerGlobals, MediaStream, RTCView} from 'react-native-webrtc';

const App = () => {
  const [consumer, setConsumer] = useState(null);
  const [stream, setStream] = useState(null);
  const textRef = useRef(null);

  const _joinRoom = async (protoo) => {
    registerGlobals();
    const device = new Device();
    const deviceInfo = {
      flag: 'unknown',
      name: 'ReactNative',
      version: 'ReactNative',
    };
    const routerRtpCapabilities = await protoo.request(
      'getRouterRtpCapabilities',
    );
    await device.load({routerRtpCapabilities});
    const transportInfo = await protoo.request('createWebRtcTransport', {
      forceTcp: false,
      producing: false,
      consuming: true,
      sctpCapabilities: undefined,
    });

    const {id, iceParameters, iceCandidates, dtlsParameters, sctpParameters} =
      transportInfo;

    const recvTransport = device.createRecvTransport({
      id,
      iceParameters,
      iceCandidates,
      dtlsParameters: {
        ...dtlsParameters,
        // Remote DTLS role. We know it's always 'auto' by default so, if
        // we want, we can force local WebRTC transport to be 'client' by
        // indicating 'server' here and vice-versa.
        role: 'auto',
      },
      sctpParameters,
      iceServers: [],
    });

    recvTransport.on(
      'connect',
      (
        {iceParameters, dtlsParameters},
        callback,
        errback, // eslint-disable-line no-shadow
      ) => {
        protoo
          .request('connectWebRtcTransport', {
            transportId: recvTransport.id,
            iceParameters,
            dtlsParameters,
          })
          .then(callback)
          .catch(errback);
      },
    );

    const {peers} = await protoo.request('join', {
      displayName: 'displayName',
      device: deviceInfo,
      rtpCapabilities: device.rtpCapabilities,
    });
    console.info(peers);

    protoo.on('request', async (request, accept, reject) => {
      switch (request.method) {
        case 'newConsumer': {
          const {peerId, producerId, id, kind, rtpParameters, appData} =
            request.data;

          try {
            const consumer = await recvTransport.consume({
              id,
              producerId,
              kind,
              rtpParameters,
              // NOTE: Force streamId to be same in mic and webcam and different
              // in screen sharing so libwebrtc will just try to sync mic and
              // webcam streams from the same remote peer.
              streamId: `${peerId}-${appData.share ? 'share' : 'mic-webcam'}`,
              appData: {...appData, peerId}, // Trick.
            });

            // // Store in the map.
            // this._consumers.set(consumer.id, consumer);

            // consumer.on('transportclose', () => {
            //   this._consumers.delete(consumer.id);
            // });

            parseScalabilityMode(
              consumer.rtpParameters.encodings[0].scalabilityMode,
            );

            setConsumer(consumer);
            const localStream = new MediaStream();
            localStream.addTrack(consumer.track);
            setStream(localStream);

            // We are ready. Answer the protoo request so the server will
            // resume this Consumer (which was paused for now if video).
            accept();
          } catch (error) {
            console.error('"newConsumer" request failed:%o', error);

            throw error;
          }

          break;
        }
      }
    });
  };

  const pause = () => {
    console.log('pause');
    if (consumer) {
      consumer.pause();
    }
  };
  const resume = () => {
    console.log('resume');
    if (consumer) {
      consumer.resume();
    }
  };
  const start = async () => {
    console.log('start');
    if (!stream) {
      try {
        const peerId = randomString({length: 8}).toLowerCase();
        const url = `wss://v3demo.mediasoup.org:4443/?roomId=${textRef.current}&peerId=${peerId}&consumerReplicas=undefined`;
        const protooTransport = new protooClient.WebSocketTransport(url);
        const _protoo = new protooClient.Peer(protooTransport);
        _protoo.on('open', () => _joinRoom(_protoo));
      } catch (e) {
        console.error(e);
      }
    }
  };
  const stop = () => {
    console.log('stop');
    if (stream) {
      stream.release();
      setConsumer(null);
      setStream(null);
    }
  };
  return (
    <>
      <StatusBar barStyle="dark-content" />
      <SafeAreaView style={styles.body}>
        {stream && <RTCView streamURL={stream.toURL()} style={styles.stream} />}
        <View style={styles.footer}>
          {!stream ? (
            <>
              <TextInput
                onChange={(event) => {
                  textRef.current = event.nativeEvent.text.trim();
                }}
              />
              <Button title="Start" onPress={start} />
            </>
          ) : (
            <>
              <Button title="Pause" onPress={pause} />
              <Button title="Resume" onPress={resume} />
              <Button title="Stop" onPress={stop} />
            </>
          )}
        </View>
      </SafeAreaView>
    </>
  );
};

const styles = StyleSheet.create({
  body: {
    backgroundColor: Colors.white,
    ...StyleSheet.absoluteFill,
  },
  stream: {
    flex: 1,
  },
  footer: {
    backgroundColor: Colors.lighter,
    position: 'absolute',
    bottom: 0,
    left: 0,
    right: 0,
  },
});

export default App;
```

The app will show you an input and a "Start" button. You have to open a web browser and go to the demo mediasoup server "https://v3demo.mediasoup.org/", write the "roomId" (it appears in the url) in the app input, and press "Start".

By this moment, you should see in the app the video that the web is sending, and the buttons "Pause" and "Resume".

Press "Pause" and you will see the video goes black. Then press "Resume", the video won't change, it will stay black.

It a very simple app that doesn't meet all mediasoup requirements, but it is enough to reproduce the issue.

saghul:
> Looks like by default the receiver/sender is equal to the track id if it's a part of a transceiver that's coming in with an associated track:
> 
> Receiver: https://github.com/jitsi/webrtc/blob/0c4b9fd122265548f6b7fbb6784a88ba61c5a1cb/pc/sdp_offer_answer.cc#L3846

Doesn't that show it's equal to the stream's ID?

saghul:
> I think the reason why I wrote it this way to use the receiver id was because I assumed the track was already removed and thus `receiver.track.id` would be unavailable. This will need some testing to make sure.

AFAIS that won't be the case. The C++ layer for AudioRtpReceiver even accesses the track on its destructor. So I'd say it's safe to use the track ID and avoid relying on internal coincidences which may not hold true forever.

davidliu:
> AFAIS that won't be the case. The C++ layer for AudioRtpReceiver even accesses the track on its destructor. So I'd say it's safe to use the track ID and avoid relying on internal coincidences which may not hold true forever.

Sounds good. We can treat this PR as a fix unrelated to Carlos' issue and address in a separate PR when we figure that out.

saghul:
SGTM!

8BallBomBom:
Closed as no longer needed.

PR #1561, opened 2024-05-02
ios: add all available camera device types

### LINKED ISSUE #1560: ios: mediaDevices.enumerateDevices() only returns wide angle camera as environment facing capture device
#### Expected Behavior:

`mediaDevices.enumerateDevices()` should return all available capture devices instead of just [Wide Angle](https://developer.apple.com/documentation/avfoundation/avcapturedevicetypebuiltinwideanglecamera?language=objc) camera.

Reference for all available capture devices:
https://developer.apple.com/documentation/avfoundation/avcapturedevicetype?language=objc

#### Observed Behavior:

when calling `mediaDevices.enumerateDevices()` with an iPhone it only returns:
```
{
    "deviceId": "com.apple.avfoundation.avcapturedevice.built-in_video:0",
    "facing": "environment",
    "groupId": "",
    "kind": "videoinput",
    "label": "Back Camera"
}
```

#### Steps to reproduce the issue:

1. attach physical iPhone as a test device.
2. call `mediaDevices.enumerateDevices()` to fetch list of available capture devices.

#### Platform Information  
* **React Native Version**: 0.73.6
* **WebRTC Module Version**: 118.0.7
* **Platform OS + Version**:  iOS

brkkyk:
Fixes #1560

### DIFF
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -263,7 +263,7 @@ RCT_EXPORT_METHOD(enumerateDevices : (RCTResponseSenderBlock)callback) {
 #else
     NSMutableArray *devices = [NSMutableArray array];
     AVCaptureDeviceDiscoverySession *videoevicesSession =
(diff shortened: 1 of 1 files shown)


saghul:
Thanks for the patch!

I fear this will break existing apps since there used to be just 2 camera sources.

Can we at least guarantee the first result is the same as before?

brkkyk:
> Thanks for the patch!
> 
> I fear this will break existing apps since there used to be just 2 camera sources.
> 
> Can we at least guarantee the first result is the same as before?

I change the order so that it continues to work the same way. It seems like `findDeviceForPosition` just iterates so it should be fine, what do you think?
https://github.com/react-native-webrtc/react-native-webrtc/blob/master/ios/RCTWebRTC/VideoCaptureController.m#L190

8BallBomBom:
LGTM 🤔
Haven't tested but as long as the original functionality remains, should be good to go.

PR #1565, opened 2024-05-07
webrtc: update WebRTC to M124

saghul:


8BallBomBom:
Will run some tests soon, also looks to not be very disruptive for the New Arch stuff 😄

saghul:
It should be a small update.

Looks like getting the header extension capabilities is now attainable though, I'll work on adding that too.

8BallBomBom:
Didn't seem to experience any issues with basic testing.
Will pull this in with the New Arch branch soon, most likely after release.
Good work as always 😃

PR #1576, opened 2024-06-11
ios: refactor rendering in RTCVideoView

saghul:
The way the renderer is implemented is as follows: there is a UIView which holds on to a RTCMTLVideoView, which does the actual rendering in metal.

Rather than using our own home-grown mechanism to layout the views, just make sure the Metal view takes 100% of the space of our view, and then use the `videoContentMode` property to set it to cover or contain, matching the way a browser would do it.

This greatly simplifies the code because we lonnger care about the video size, the Metal renderer makes the adjustments autmagically.

In addition, the mirror and object-fit properties are only applied once, when they change, not on every layout change, which is unnecessary.

### DIFF
--- a/ios/RCTWebRTC/RTCVideoViewManager.m
+++ b/ios/RCTWebRTC/RTCVideoViewManager.m
@@ -3,10 +3,10 @@
 
 #import <React/RCTLog.h>
 #import <WebRTC/RTCMediaStream.h>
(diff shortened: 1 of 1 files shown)


saghul:
@Calinteodor Can you please review this and take it for a spin with Jitsi Meet?

PR #1582, opened 2024-06-21
fix: Handle invalidate for react-native 0.74+

davidliu:
React-Native no longer calls onCatalystInstanceDestroy in 0.74+ (used to be called when the app reloaded). On 0.74+, this has been replaced by `invalidate()`.

This was causing crashes for consumers who use a custom AudioDeviceModule (as well as perhaps leaking the old ADM for those who use the default).

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -117,6 +117,11 @@ public class WebRTCModule extends ReactContextBaseJavaModule {
     }
 
     @Override
+    public void invalidate() {
+        onCatalystInstanceDestroy();
+    }
+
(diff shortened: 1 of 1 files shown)


saghul:
Can you try releasing it right after it's being basses to the factory?

I was looking into it the other day and since refevremce counting is used, that looked like the right thing to do.

davidliu:
Oh wow, you're right. They release it right after the peer connection factory creation in the example app. That's super counterintuitive. I'll test it out and if it works fine, open another PR.

saghul:
Yeah. I didn't notice it at the time, but realized recently while looking into something else. It kinda makes sense because the PCF "owns" it after initializing...

davidliu:
Closing in favor of #1583.

PR #1584, opened 2024-06-22
fix data channel

danilaplee:
Update RTCDataChannel.ts remove buffer

### DIFF
--- a/src/RTCDataChannel.ts
+++ b/src/RTCDataChannel.ts
@@ -157,7 +157,7 @@ export default class RTCDataChannel extends EventTarget<DataChannelEventMap> {
             let data = ev.data;
 
             if (ev.type === 'binary') {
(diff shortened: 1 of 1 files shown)


saghul:
I think your fix is incorrect. `base64.toByteArray` returns `Uint8Array`, and `RTCDataChannel` specifies `ArrayBuffer` is what's returned for binary data: https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/binaryType

With your change we'd be sending `Uint8Array` instead of `ArrayBuffer` in the event.

What prompted you to need to make this change?

danilaplee:
I experience that Uint8Array has no buffer property

danilaplee:
actually i cannot reproduce this issue anymore seems like maybe it was in the simple-peer library

PR #1588, opened 2024-06-27
ios: fix compatibility with RN >= 0.73

saghul:
Fixes: https://github.com/jitsi/jitsi-meet/issues/14441

### DIFF
--- a/ios/RCTWebRTC/RTCVideoViewManager.m
+++ b/ios/RCTWebRTC/RTCVideoViewManager.m
@@ -2,6 +2,8 @@
 #import <objc/runtime.h>
 
 #import <React/RCTLog.h>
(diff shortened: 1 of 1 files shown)


saghul:
@Calinteodor Can you pl take this for a spin tomorrow?

Calinteodor (approved):

Calinteodor:
> @Calinteodor Can you pl take this for a spin tomorrow?

Tested this commit on jitsi meet, where we use 0.72.14, everything seems to be in order.
Shouldn't we wait until we are aligned with 0.73, too?

saghul:
As long as it works with old versions we are good, because people using this package on their own are already having trouble.

PR #1609, opened 2024-08-07
android: remove no longer used replace rule from manifest

saghul:


8BallBomBom:
Ahhhh, something I corrected in the new branch but didn't back port 🫨

8BallBomBom (approved):

PR #1614, opened 2024-08-15
ios: Add RTCAudioSession convenience methods needed for CallKit

davidliu:
Adds `audioSessionDidActivate` and `audioSessionDidDeactivate` methods for JS usage.

Incoming CallKit calls seem to hold the audio session in a restricted state, which won't let WebRTC activate it on its own (and gets interpreted as an error). These methods need to be called for smooth usage with CallKit.

### DIFF
--- a/Documentation/iOSInstallation.md
+++ b/Documentation/iOSInstallation.md
@@ -26,6 +26,33 @@ Navigate to `<ProjectFolder>/ios/<ProjectName>/` and edit `Info.plist`, add the 
 <string>Microphone permission description</string>
 ```
 
+## CallKit
+
+If your app uses a CallKit integration to handle incoming calls, then your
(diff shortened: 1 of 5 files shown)


saghul:
Not opposing the change, but I'll ask: what's the game plan here?

If you already have some native code for dealing with CallKit, importing WebRTC to get ahold of RTCAudioSession is not a problem.

In what scenario is this necessary and doing the call on the native side would be a hassle?

davidliu:
I'll be honest, I'm not fully familiar with CallKit myself (iOS is outside of my forte), but mostly was thinking about the integration with `react-native-callkeep`, which wraps the CallKit integration and RN devs seem to be isolated from the native code there, but the relevant events are exposed to the JS side. I'm open to other ideas here.

saghul:
Gotcha. If that is the case then yeah this will help.

saghul (approved):

8BallBomBom (approved):

PR #1615, opened 2024-08-19
ios,android: add device/groupId to MediaStreamTrack.getSettings and implement applyConstraints

davidliu:
Adds [missing deviceId/groupId values](https://w3c.github.io/mediacapture-main/#media-track-settings) to the settings map sent back for MediaStreamTracks.

Additionally changes the `_switchCamera` method to an async method, so that it can get the final settings of the camera when done switching.

The one thing that's missing is iOS audio track deviceId/groupId. Can't find any way to associate the available input mics to the one actually in use by WebRTC. Can change `enumerateDevices` to send back a dummy deviceId like we do on Android, since iOS webrtc doesn't really do anything with audio device ids either, but I'm not sure if this might adversely affect anybody by this change.

### DIFF
--- a/Documentation/BasicUsage.md
+++ b/Documentation/BasicUsage.md
@@ -307,8 +307,18 @@ try {
 	// Taken from above, we don't want to flip if we don't have another camera.
 	if ( cameraCount < 2 ) { return; };
 
-	const videoTrack = await localMediaStream.getVideoTracks()[ 0 ];
-	videoTrack._switchCamera();
+	const videoTrack = localMediaStream.getVideoTracks()[0];
+
+	let constraints;
+	if (isFrontCam) {
+		constraints = { facingMode: 'environment' };
+	} else {
+		constraints = { facingMode: 'user' };
+	}
+
+	videoTrack.applyConstraints(constraints);
+	// _switchCamera is deprecated as of 124.0.5
+	// videoTrack._switchCamera();
 
 	isFrontCam = !isFrontCam;
 } catch( err ) {
--- a/Documentation/BasicUsage.md
+++ b/Documentation/BasicUsage.md
@@ -308,15 +308,10 @@ try {
 	if ( cameraCount < 2 ) { return; };
 
 	const videoTrack = localMediaStream.getVideoTracks()[0];
-
-	let constraints;
-	if (isFrontCam) {
-		constraints = { facingMode: 'environment' };
-	} else {
-		constraints = { facingMode: 'user' };
-	}
+	const constraints = { facingMode: isFrontCam ? 'user' : 'environment' };
 
 	videoTrack.applyConstraints(constraints);
+
 	// _switchCamera is deprecated as of 124.0.5
 	// videoTrack._switchCamera();
 
--- a/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
+++ b/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
@@ -263,6 +263,7 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
                 } else {
                     failedDevices.add(name);
                 }
+            }
         }
 
         currentDeviceId = null;
--- a/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
+++ b/android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
@@ -6,6 +6,9 @@ import android.util.Log;
 import android.util.Pair;
 import androidx.annotation.Nullable;
 
+import androidx.annotation.Nullable;
+import androidx.core.util.Consumer;
+
 import com.facebook.react.bridge.ReadableMap;
 
 import org.webrtc.Camera1Capturer;
@@ -19,6 +22,7 @@ import org.webrtc.VideoCapturer;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 public class CameraCaptureController extends AbstractVideoCaptureController {
     /**
@@ -27,6 +31,8 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
     private static final String TAG = CameraCaptureController.class.getSimpleName();
 
     private boolean isFrontFacing;
+    @Nullable
+    private String currentDeviceId = null;
 
     private final Context context;
     private final CameraEnumerator cameraEnumerator;
@@ -38,7 +44,21 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
      * implementation does not do anything but logging unspecific to the camera
      * device's name anyway.
      */
-    private final CameraEventsHandler cameraEventsHandler = new CameraEventsHandler();
+    private final CameraEventsHandler cameraEventsHandler = new CameraEventsHandler() {
+
+        @Override
+        public void onCameraOpening(String cameraName) {
+            super.onCameraOpening(cameraName);
+            String[] deviceNames = cameraEnumerator.getDeviceNames();
+            CameraCaptureController.this.currentDeviceId = null;
+            for (int i = 0; i < deviceNames.length; i++) {
+                if (Objects.equals(deviceNames[i], cameraName)) {
+                    CameraCaptureController.this.currentDeviceId = String.valueOf(i);
+                    break;
+                }
+            }
+        }
+    };
 
     public CameraCaptureController(Context context, CameraEnumerator cameraEnumerator, ReadableMap constraints) {
         super(constraints.getInt("width"), constraints.getInt("height"), constraints.getInt("frameRate"));
@@ -48,7 +68,13 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
         this.constraints = constraints;
     }
 
-    public void switchCamera() {
+    @Nullable
+    @Override
+    public String getDeviceId() {
+        return currentDeviceId;
+    }
+
+    public void switchCamera(Consumer<Exception> onFinishedCallback) {
         if (videoCapturer instanceof CameraVideoCapturer) {
             CameraVideoCapturer capturer = (CameraVideoCapturer) videoCapturer;
             String[] deviceNames = cameraEnumerator.getDeviceNames();
@@ -63,13 +89,20 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
             if (deviceCount == 2) {
                 capturer.switchCamera(new CameraVideoCapturer.CameraSwitchHandler() {
                     @Override
-                    public void onCameraSwitchDone(boolean b) {
-                        isFrontFacing = b;
+                    public void onCameraSwitchDone(boolean isFrontCamera) {
+                        isFrontFacing = isFrontCamera;
+                        if(onFinishedCallback != null) {
+                            onFinishedCallback.accept(null);
+                        }
                     }
 
                     @Override
                     public void onCameraSwitchError(String s) {
-                        Log.e(TAG, "Error switching camera: " + s);
+                        Exception e = new Exception("Error switching camera: " + s);
+                        Log.e(TAG, "OnCameraSwitchError", e);
+                        if(onFinishedCallback != null) {
+                            onFinishedCallback.accept(e);
+                        }
                     }
                 });
                 return;
@@ -77,7 +110,7 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
 
             // If we are here the device has more than 2 cameras. Cycle through them
             // and switch to the first one of the desired facing mode.
-            switchCamera(!isFrontFacing, deviceCount);
+            switchCamera(!isFrontFacing, deviceCount, onFinishedCallback);
         }
     }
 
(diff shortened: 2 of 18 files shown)


saghul:
Thanks for doing this!

Chaging `_switchCamera` to be async gives me pause, as I'm not sure it can break stuff.

Since we are going for completeness here, I think it's time we switch (pun 100% intended!) to `applyConstraints` and do things The Standard Way (TM).

When we can deprecate `_switchCamera` and phase it out.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/applyConstraints

The standard way is more flexible because it's more obvious you want to change to a given `facingMode` or `deviceId`.

I don't remember if `applyConstraints` was a thing when I first introduced `_switchCamera` but it feels like the right time to right a wrong.

saghul:
> Technically speaking, it seems like web doesn't allow switching cameras through applyConstraints, but I think we can skip over that.

Yeah, I haven't tried it on mobile browsers, but I'd say beeing able to do change that constraint is a reasonable expectation  ;-)

davidliu:
@saghul I think this is ready for another review, applyConstraints is now implemented.

Documentation/BasicUsage.md
```diff
@@ -307,8 +307,18 @@ try {
 	// Taken from above, we don't want to flip if we don't have another camera.
 	if ( cameraCount < 2 ) { return; };
 
-	const videoTrack = await localMediaStream.getVideoTracks()[ 0 ];
-	videoTrack._switchCamera();
+	const videoTrack = localMediaStream.getVideoTracks()[0];
+
+	let constraints;
```
saghul, line 312: Maybe write this inline?

```js
const constraints = { facingMode: isFontCam ? 'user' : 'environment' };
```

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -6,6 +6,9 @@
 import android.util.Pair;
 import androidx.annotation.Nullable;
 
+import androidx.annotation.Nullable;
```
saghul, line 9: This is already imported a couple of lines above.

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -27,18 +31,26 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
     private static final String TAG = CameraCaptureController.class.getSimpleName();
 
     private boolean isFrontFacing;
+    @Nullable
+    private String currentDeviceId = null;
```
saghul, line 35: Unnecessary initialization.

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -27,18 +31,26 @@ public class CameraCaptureController extends AbstractVideoCaptureController {
...
+     * {@link CameraEnumerator#createCapturer}.
      */
-    private final CameraEventsHandler cameraEventsHandler = new CameraEventsHandler();
+    private final CameraEventsHandler cameraEventsHandler = new CameraEventsHandler() {
+
+        @Override
+        public void onCameraOpening(String cameraName) {
+            super.onCameraOpening(cameraName);
+            updateActualSize(cameraName, videoCapturer);
+            CameraCaptureController.this.currentDeviceId = findDeviceId(cameraName);
```
saghul, line 51: You also need to find the device ID in `updateActualSize`, do you reckon we could rework this a bit to avoid doing it twice? Maybe first get the ID, then update the size with the ID?

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -48,7 +60,132 @@ public CameraCaptureController(Context context, CameraEnumerator cameraEnumerato
...
+    public void applyConstraints(ReadableMap constraints, Consumer<Exception> onFinishedCallback) {
+        ReadableMap oldConstraints = this.constraints;
+        int oldTargetWidth = this.targetWidth;
+        int oldTargetHeight = this.targetHeight;
+        int oldTargetFps = this.targetFps;
+
+        // Don't save constraints yet, since we may fail to find a fit.
+        Runnable saveConstraints = () -> {
+            this.constraints = constraints;
+            this.targetWidth = constraints.getInt("width");
```
saghul, line 88: Are we guaranteed to have some value for width here? What if it's missing from the passed constraints?

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -48,7 +60,132 @@ public CameraCaptureController(Context context, CameraEnumerator cameraEnumerato
...
+            saveConstraints.run();
+            onFinishedCallback.accept(null);
+            return;
+        }
+
+        // Find target camera to switch to.
+        String[] deviceNames = cameraEnumerator.getDeviceNames();
+        final String deviceId = ReactBridgeUtil.getMapStrValue(constraints, "deviceId");
+        final String facingMode = ReactBridgeUtil.getMapStrValue(constraints, "facingMode");
+        final boolean isFrontFacing = facingMode == null || !facingMode.equals("environment");
```
saghul, line 104: Why not check if it's missing or it equals "user"?

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -48,7 +60,132 @@ public CameraCaptureController(Context context, CameraEnumerator cameraEnumerato
...
+        CameraVideoCapturer capturer = (CameraVideoCapturer) videoCapturer;
+        Runnable changeFormatIfNeededAndFinish = () -> {
+            saveConstraints.run();
+            if (targetWidth != oldTargetWidth ||
+                    targetHeight != oldTargetHeight ||
+                    targetFps != oldTargetFps) {
+                updateActualSize(finalCameraName, videoCapturer);
+                capturer.changeCaptureFormat(targetWidth, targetHeight, targetFps);
+            }
+            if (onFinishedCallback != null) {
```
saghul, line 155: You are not checking for null when calling this callback above.

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -48,7 +60,132 @@ public CameraCaptureController(Context context, CameraEnumerator cameraEnumerato
...
+            changeFormatIfNeededAndFinish.run();
+        }
+    }
+
+    /**
+     * Use applyConstraints instead.
+     * @deprecated
+     */
+    @Deprecated
+    public void switchCamera(Consumer<Exception> onFinishedCallback) {
```
saghul, line 188: Why keep this? Can't we re-implement it in JS using `applyConstraints` internally?

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -184,7 +341,9 @@ private Pair<String, VideoCapturer> createVideoCapturer(String deviceId, String
 
         // Otherwise, use facingMode (defaulting to front/user facing).
         final boolean isFrontFacing = facingMode == null || !facingMode.equals("environment");
```
saghul, line 343: Same as above, let's write this the straightforward way, not sure why I wrote it that way in the past 😅

android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java
```diff
@@ -196,29 +355,33 @@ private Pair<String, VideoCapturer> createVideoCapturer(String deviceId, String
...
             if (!failedDevices.contains(name)) {
                 VideoCapturer videoCapturer = cameraEnumerator.createCapturer(name, cameraEventsHandler);
                 String message = "Create fallback camera " + name;
                 if (videoCapturer != null) {
                     Log.d(TAG, message + " succeeded");
                     this.isFrontFacing = cameraEnumerator.isFrontFacing(name);
+                    this.currentDeviceId = String.valueOf(cameraIndex);
                     return new Pair(name, videoCapturer);
                 } else {
-                    Log.d(TAG, message + " failed");
```
saghul, line 216: I think this log line could be useful, any reason not to keep it?

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -233,14 +234,43 @@ void disposeTrack(String id) {
         }
     }
 
-    void switchCamera(String trackId) {
+    void switchCamera(String trackId, Promise promise) {
```
saghul, line 237: Ditto, can we drop this implementation and do it in JS?

android/src/main/java/com/oney/WebRTCModule/GetUserMediaImpl.java
```diff
@@ -233,14 +234,43 @@ void disposeTrack(String id) {
...
+                }
+            });
+        } else {
+            promise.reject(new Exception("Camera track not found!"));
         }
     }
 
+    void applyConstraints(String trackId, ReadableMap constraints, Promise promise) {
+        TrackPrivate track = tracks.get(trackId);
+        if (track != null && track.videoCaptureController instanceof CameraCaptureController) {
```
saghul, line 258: I guess we might need to tweak this in the future if screen tracks can be tweaked too. Would it make sense to have an implementation in the base class that always returns an error and override it in the camera capturer?

android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
```diff
@@ -882,11 +882,25 @@ public void mediaStreamTrackSetEnabled(int pcId, String id, boolean enabled) {
     }
 
     @ReactMethod
-    public void mediaStreamTrackSwitchCamera(String id) {
+    public void mediaStreamTrackSwitchCamera(String id, Promise promise) {
```
saghul, line 885: Damn it goes deep. Now I'm more convinced I want to kill it :-P

src/MediaStreamTrack.ts
```diff
@@ -107,7 +109,11 @@ export default class MediaStreamTrack extends EventTarget<MediaStreamTrackEventM
             throw new Error('Only implemented for video tracks');
         }
 
-        WebRTCModule.mediaStreamTrackSwitchCamera(this.id);
+        const switchImpl = async () => {
```
saghul, line 112: Alright, this is what I meant. Can we simply use the existing settings, flip the facing mode and call applyConstraints with just that constraint?

src/MediaStreamTrack.ts
```diff
@@ -150,8 +156,32 @@ export default class MediaStreamTrack extends EventTarget<MediaStreamTrackEventM
...
+     * constraints are cleared.
+     */
+    async applyConstraints(constraints?: object): Promise<void> {
+        if (this.kind !== 'video') {
+            throw new Error('Only implemented for video tracks');
+        }
+
+        let video: boolean | object;
+
+        if (constraints === undefined) {
```
saghul, line 176: Do we need this? Looks like we could simply do:

```js
    const normalized = normalizeConstraints(constraints ?? {});
````

File: android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java (hunk 5 above)
Comment (davidliu) [reply android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java:86]: `normalizeMediaConstraints` ensures we have a width or height if either one is missing, so I think this should be fine.

**Comment by davidliu** [comment]
@saghul Addressed the comments, good for another look over.

File: android/src/main/java/com/oney/WebRTCModule/CameraCaptureController.java [hunk 16]
```diff
@@ -196,31 +288,48 @@ private Pair<String, VideoCapturer> createVideoCapturer(String deviceId, String
...
             }
         }
 
+        currentDeviceId = null;
         Log.w(TAG, "Unable to identify a suitable camera.");
 
         return null;
     }
+
+    private static class CreateCapturerResult {
```
saghul, line 324: ❤️

src/MediaStreamTrack.ts
```diff
@@ -107,7 +119,16 @@ export default class MediaStreamTrack extends EventTarget<MediaStreamTrackEventM
             throw new Error('Only implemented for video tracks');
         }
 
-        WebRTCModule.mediaStreamTrackSwitchCamera(this.id);
+        const constraints = deepClone(this._settings);
```
saghul, line 122: Question: should we just set the facingMode and not any other setting, since different cameras might have different supported resolutions?

src/MediaStreamTrack.ts
```diff
@@ -107,7 +119,16 @@ export default class MediaStreamTrack extends EventTarget<MediaStreamTrackEventM
...
+
+        delete constraints.deviceId;
+        constraints.facingMode = this._settings.facingMode === 'user' ? 'environment' : 'user';
+
+        const switchImpl = async () => {
+            await this.applyConstraints(constraints);
+        };
+
+        switchImpl();
```
saghul, line 131: Since we are not really awaiting this, we can simplee call `this.applyConstraints(constraints);` here.

saghul (approved):
LGTM with 2 suggestions, great work!

src/MediaStreamTrack.ts
davidliu: Old behavior of switch camera kept the target resolution, and worst case it'll just find the closest available resolution so it should be fine until we can remove this entirely.

saghul: Sounds good.

PR #1625, opened 2024-09-16
android: don't use minSdkVersion set by app

davidliu:
Client apps could set their minSdkVersion to a lower version (RN 0.73 defaults to 21, and RN 0.74 is at 23), unknowingly sidestepping the minSdkVersion restriction.

They can still explicitly override it by adding the following to their app's AndroidManifest.xml:
```
<uses-sdk tools:overrideLibrary="com.oney.WebRTCModule"/>
```


saghul (approved):
Good catch!

Issue #1631, opened 2024-10-02
Add width/height styles example for RTCView in the documentation


aszmyd:
I've spent several hours trying to figure out why my RTCView is not rendering MediaStream even though everything seemed properly setup. I simply captured own camera from user and wanted to display it using RTCView. No WebRTC connection or some such.

After analyzing example codes, I found out that there were `styles` defined there. So I've added 
```
<RTCView
     style={{ width: "100%", height: "100%" }}
    ...
```

and **that did the trick**. The component was simply somehow 0x0px it seems. 

It could be worth adding some note related to that in the [BasicUsage](https://github.com/react-native-webrtc/react-native-webrtc/blob/master/Documentation/BasicUsage.md#rendering-the-media-stream) section.

Am I missing something here? Even if for some reason my setup (Expo, shadcn/ui) was making it "hidden", it might save a lot of troubleshoot time for other folks if a simple hint like this could be added to the documentation.

8BallBomBom:
Closed in favour of the proposed solution on the linked pr.

PR #1632, opened 2024-10-02
Update BasicUsage.md

aszmyd:
Docs update (https://github.com/react-native-webrtc/react-native-webrtc/issues/1631)

### DIFF
--- a/Documentation/BasicUsage.md
+++ b/Documentation/BasicUsage.md
@@ -342,6 +342,13 @@ Don't forget, the user facing camera is usually mirrored.
 | streamURL | string | 'streamURL' | Required to have an actual video stream rendering. |
 | zOrder | number | 0 | Similar to zIndex. |
 
+> [!NOTE]  
+> If for some reason you don't see the rendered video element, you might want to try adding the `style` parameter to the `RTCView` component. This can help ensure the component has the correct dimensions and positioning. For example:
+> ```javascript
+> <RTCView style={{ width: '100%', height: '100%' }} ...
+> ```
+
+
 ## Controlling remote audio tracks
 
 Remote audio tracks are automatically handled and played through your audio output.


saghul:
I'm not sure about the wording here regarding not seeing video.

We receive many issues like that, but the problem is different usually server side, and this might point people in the wrong direction.

Thoughts? @davidliu @8BallBomBom

aszmyd:
Well, it's worth adding a line of styles to check if it won't fix things for people before troubleshooting servers :) 
I'm fine with changing the wording to whatever you prefer.

8BallBomBom:
Definitely not something you'd think would be needed in the basic usage docs 🤔 
By default you would style components to position and size them as needed for your project.
Typically that is a known basic requirement for when you build anything you want to render on screen.

Honestly this has surprised me, not sure how to respond 🫨

saghul:
> That being said, another thought, maybe a better solution would be to do a default style for the component.
> Which could then also include some documentation stating the component is set to fit the allowed space by default.

I like this.

8BallBomBom:
@aszmyd Do you want to make the discussed changes?
Otherwise, no worries, will open a pr shortly 😄

aszmyd:
> @aszmyd Do you want to make the discussed changes? Otherwise, no worries, will open a pr shortly 😄

I'm happy to try it. Need to dive a bit deeper into how the components are made here but that would be a good learning.

Mrjaco12:
I wanted to say that I was tripped up by this last week. This PR was helpful to me to understand the intent for the `RTCView` component. I think maybe the `objectFit` prop implies some level of default styling that doesn't exist. I'd be happy to help out if I can.

8BallBomBom:
Definitely could be more clear.
Will pick this up when i have some free time.
A lot of good points to consider 🤔

PR #1645, opened 2024-11-04
doc: remove duplicate / confusing information about screen-sharing

saghul:


8BallBomBom:
Old but gold, definitely not required anymore 😅

Issue #590, opened 2019-02-18
Crashes on start in release build


yuensunn:
Used to work perfectly fine (both debug & release build ) in the past. Not sure what I did to trigger this problem.

#### ADB logcat result

> 02-18 17:13:28.746 21624 21646 E rtc     : #
02-18 17:13:28.746 21624 21646 E rtc     : # Fatal error in: ../../sdk/android/src/jni/jni_generator_helper.cc, line 32
02-18 17:13:28.746 21624 21646 E rtc     : # last system error: 88
02-18 17:13:28.746 21624 21646 E rtc     : # Check failed: !clazz.is_null()
02-18 17:13:28.746 21624 21646 E rtc     : # org/webrtc/WebRtcClassLoader
02-18 17:13:28.747 21624 21646 F libc    : Fatal signal 6 (SIGABRT), code -6 in tid 21646 (pool-1-thread-1)

#### Platform information

* **React Native version**: 0.57.7
* **Plugin version**: ^1.69.0
* **OS**:  Android
* **OS version**: 8.0.0

yuensunn:
Update: Everything seems to work fine after clearing npm caches and reinstalling all node modules.

epurban:
I'm having this problem as well. @yuensunn , can you detail any other steps you took to fix this?

yuensunn:
@epurban I can't remember exactly but here are some of the steps that I did

- cleaning the gradle `./gradlew clean`
- deleting the entire `node_modules` folder and `npm install` again
- `npm clean cache`

or maybe you could try upgrading your webrtc package?

epurban:
I'll try those steps now. I'm the latest version 1.69.1. Actually as I write this I notice they released 1.69.2, so I'll try upgrading and see if that fixes anything. You said it came out of seemingly nowhere? How long did you tussle with this issue before you finally resolved it?

epurban:
After upgrading, cleaning the build folder, removing node modules and reinstalling, and clearing npm cache, I still get this same error. it's a bit different from yours, but here it is:

[...]

PR #1655, opened 2024-12-17
android: add libwebrtc consumer proguard rules

davidliu:
Addresses #590 so that downstream consumers don't have to add their own rule (which can be annoying in Expo environments).

----

~~Needs testing. This works elsewhere for AARs, but I need to verify that it works in the react-native build system. Will get to it this weekish, but putting this up here for now.~~

Tested, works fine.

### DIFF
--- a/Documentation/AndroidInstallation.md
+++ b/Documentation/AndroidInstallation.md
@@ -64,14 +64,6 @@ compileOptions {
 }
 ```
 
-## R8/ProGuard Support
-
-In `android/app/proguard-rules.pro` add the following on a new line.
-
-```proguard
--keep class org.webrtc.** { *; }
-```
-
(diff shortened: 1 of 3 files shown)


saghul:
TIL about this, nice! Is there a minimum Gradle plugin version required?

davidliu:
Ooh, that's a good question. Looked it up, and luckily it seems like this has been there for a long while, [since at least AGP 2.0](https://google.github.io/android-gradle-dsl/2.0/com.android.build.gradle.internal.dsl.ProductFlavor.html#com.android.build.gradle.internal.dsl.ProductFlavor:consumerProguardFiles), which should cover pretty much any version of RN.

saghul:
Awesome! If this works, let's have it, it's a great improvement!

saghul (approved):

Issue #1542, opened 2024-04-11
Use speaker of media instead of use speaker of calling


sstatsuya:
I'm using React Native WebRTC to display the livestream, but my device is using the call speaker, making it seem like it's in a call. You can see the photos below.

#### Expected Behavior:
I only want to display the audio and video of the streamer, not to the community.

#### Observed Behavior:
When I adjust the volume bar of the call (highlighted in the red frame), the volume of the stream will increase/decrease. However, when I adjust the volume bar of the media, nothing happens.

![image](https://github.com/react-native-webrtc/react-native-webrtc/assets/70736475/3a493ece-fb84-4eeb-abef-9574edd432d3)
![image](https://github.com/react-native-webrtc/react-native-webrtc/assets/70736475/173483cc-e1a7-43b3-ac1f-d4e3320b972d)

I want to use the media speaker because the call speaker doesn't decrease to zero; it always stays slightly higher than zero. Additionally, the sound quality of calls is worse than that of media. Sometimes I hear voices like on a radio, along with a lot of background noise when using the calling speaker.

#### Platform Information  
* **React Native Version**:  ```0.71.3``` ```0.63.4```
* **WebRTC Module Version**: ```111.0.6```
* **Platform OS + Version**: ```Android``` ```IOS```

Thanks for all your help.

saghul:
You'll need to write a small jmplugin to change the audio category. By default it's in the calls category.

sstatsuya:
> You'll need to write a small jmplugin to change the audio category. By default it's in the calls category.

I'm not proficient in writing native modules, can you give me some references to write this plugin?

davidliu:
You'll also need to change the audio output category in the webrtc module.

In your MainApplication.java:

```
public class MainApplication extends Application implements ReactApplication {

  @Override
  public void onCreate() {
    // do this before WebRTCModule initializes
    WebRTCModuleOptions options = WebRTCModuleOptions.getInstance()
    AudioAttributes audioAttributes = AudioAttributes.Builder()
            .setUsage(AudioAttributes.USAGE_MEDIA)
            .setContentType(AudioAttributes.CONTENT_TYPE_SPEECH)
            .build(),
    options.audioDeviceModule = JavaAudioDeviceModule.builder(this)
        .setAudioAttributes(audioAttributes)
        .createAudioDeviceModule()
    // ...
  }
}
```

[...]

PR #1660, opened 2025-01-02
doc [Android] - set audio category to media

msabaeian:
This PR adds the manual for setting the audio category to media in Android

### DIFF
--- a/Documentation/AndroidInstallation.md
+++ b/Documentation/AndroidInstallation.md
@@ -64,6 +64,55 @@ compileOptions {
 }
 ```
 
+## Set audio category (output) to media
+
+By default, the audio is considered calls and it's not possible to hear it through the phone’s speakers, if you want to change this behavior you need to set the audio category to media so that the audio is not treated as a call stream. To do this:
+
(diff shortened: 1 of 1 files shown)


saghul:
The first sentence is not correct, you can use the call mode and switch devices.

You will need to use a separate plugin for that though. See incall-manager.

msabaeian:
@saghul well even with installing `incall-manager` the call category is still call and not media, unless I'm wrong.

I can change it to this if you prefer:
`The audio is considered calls by default, if you don't want your audio to be treated as a call stream you need to change the category. To set the category:`

maybe it's also good to point to [https://developer.android.com/reference/android/media/AudioAttributes#summary](AudioAttributes) here.

ps: if you don't think it's useful for other people feel free to close it, this is only updating docs based on #1542 ([java](https://github.com/react-native-webrtc/react-native-webrtc/issues/1542#issuecomment-2048975310), [kotlin](https://github.com/react-native-webrtc/react-native-webrtc/issues/1542#issuecomment-2267507173))

saghul:
The proposed new text does look better

saghul (approved):

PR #1661, opened 2025-01-07
ci: update GHA versions

saghul:
Use nvm for choosing the Node version.


saghul:
We need to update the sample project, I'll do that shortly.

PR #1331, opened 2023-02-22
Multiple video frame processors (+ implementation for ios)

wjaykim:
This PR adds support for registering multiple video frame processors.
Plus, it adds missing video frame processor functionality for ios platform as well.

PR #1681, opened 2025-03-08
Multiple video frame processors + implementation for ios

8BallBomBom:
Original pr and commits by [wjaykim](https://github.com/wjaykim) over here #1331
Decided to grab the lot and make sure things were up to date with the master branch.
Saves a lot of the previous mess, my bad.

Everything should be all good but always best to test.
We don't currently have an example for iOS but there is one in the old pr for Android.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/videoEffects/VideoEffectProcessor.java
+++ b/android/src/main/java/com/oney/WebRTCModule/videoEffects/VideoEffectProcessor.java
@@ -5,6 +5,8 @@ import org.webrtc.VideoFrame;
 import org.webrtc.VideoProcessor;
 import org.webrtc.VideoSink;
 
+import java.util.List;
+
 /**
  * Lightweight abstraction for an object that can receive video frames, process and add effects in
  * them, and pass them on to another object.
@@ -12,11 +14,11 @@ import org.webrtc.VideoSink;
 public class VideoEffectProcessor implements VideoProcessor {
     private VideoSink mSink;
     final private SurfaceTextureHelper textureHelper;
-    final private VideoFrameProcessor videoFrameProcessor;
+    final private List<VideoFrameProcessor> videoFrameProcessors;
 
-    public VideoEffectProcessor(VideoFrameProcessor processor, SurfaceTextureHelper textureHelper) {
+    public VideoEffectProcessor(List<VideoFrameProcessor> processors, SurfaceTextureHelper textureHelper) {
         this.textureHelper = textureHelper;
-        this.videoFrameProcessor = processor;
+        this.videoFrameProcessors = processors;
     }
 
     @Override
@@ -39,15 +41,12 @@ public class VideoEffectProcessor implements VideoProcessor {
     @Override
     public void onFrameCaptured(VideoFrame frame) {
         frame.retain();
-        VideoFrame outputFrame = videoFrameProcessor.process(frame, textureHelper);
-
-        if (outputFrame == null) {
-            mSink.onFrame(frame);
-            frame.release();
-            return;
+        VideoFrame outputFrame = frame;
+        for (VideoFrameProcessor processor : this.videoFrameProcessors) {
+            outputFrame = processor.process(outputFrame, textureHelper);
         }
+
         mSink.onFrame(outputFrame);
         outputFrame.release();
-        frame.release();
     }
 }
--- a/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
+++ b/ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
@@ -10,12 +10,23 @@
 #import "WebRTCModule+RTCMediaStream.h"
 #import "WebRTCModule+RTCPeerConnection.h"
 
+#import "ProcessorProvider.h"
 #import "ScreenCaptureController.h"
 #import "ScreenCapturer.h"
 #import "TrackCapturerEventsEmitter.h"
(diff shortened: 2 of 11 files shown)


AleanM:
Sorry i have question, can i use methods for frame processing in swift module. If i install webrtc from videoeffects branch or i  have to wait until accept this pr?

android/src/main/java/com/oney/WebRTCModule/videoEffects/VideoEffectProcessor.java
```diff
@@ -39,15 +41,12 @@ public void setSink(VideoSink sink) {
     @Override
     public void onFrameCaptured(VideoFrame frame) {
         frame.retain();
-        VideoFrame outputFrame = videoFrameProcessor.process(frame, textureHelper);
-
-        if (outputFrame == null) {
```
saghul, line 44: This condition should be restored.

android/src/main/java/com/oney/WebRTCModule/videoEffects/VideoEffectProcessor.java
```diff
@@ -39,15 +41,12 @@ public void setSink(VideoSink sink) {
...
-            frame.release();
-            return;
+        VideoFrame outputFrame = frame;
+        for (VideoFrameProcessor processor : this.videoFrameProcessors) {
+            outputFrame = processor.process(outputFrame, textureHelper);
         }
+
         mSink.onFrame(outputFrame);
         outputFrame.release();
-        frame.release();
```
saghul, line 51: This would result in a leak

ios/RCTWebRTC/VideoCaptureController.h
```diff
@@ -6,6 +6,7 @@
 #import "CaptureController.h"
 
 @interface VideoCaptureController : CaptureController
+@property (nonatomic, readonly, strong) RTCCameraVideoCapturer *capturer;
```
saghul, line 9: This should likely be the base class, so effects can be applied to any capturer.

ios/RCTWebRTC/WebRTCModule+RTCMediaStream.m
```diff
@@ -443,6 +454,31 @@ - (RTCVideoTrack *)createScreenCaptureVideoTrack {
     }
 }
 
+RCT_EXPORT_METHOD(mediaStreamTrackSetVideoEffects:(nonnull NSString *)trackID names:(nonnull NSArray<NSString *> *)names)
+{
+  RTCMediaStreamTrack *track = self.localTracks[trackID];
+  if (track) {
```
saghul, line 460: Better reverse the check and return early if there is no local track.

8BallBomBom: Will do

android/src/main/java/com/oney/WebRTCModule/videoEffects/VideoEffectProcessor.java
8BallBomBom: Will look into it a bit but no worries

android/src/main/java/com/oney/WebRTCModule/videoEffects/VideoEffectProcessor.java
8BallBomBom: Done and done

magiccDev:
Hi @8BallBomBom,

Thank you for this PR!

if you are interested, I would like to sponsor this feature, could you please let me know how I can do that?

8BallBomBom:
Would need to talk with @saghul about that 🤔 
and the original commit was by @wjaykim

wjaykim:
I'll try to make a example use of the processor when possible.
It's been a while after I initially made the pull request(and I'm not working on webrtc related projects now), so I have to realign the current codebases.

saghul:
Thank you! Nothing has significantly changed, at least ;-)

anilgupta83:
I think the changes in VideoEffectProcessor.java, specifically onFrameCaptured function has introduced a bug where frames are not sent back to webrtc resulting in black screen. I modified the code to pick just the first processor, used the old logic and it fixed the problem.

```
        frame.retain();
        VideoFrameProcessor firstProcessor = videoFrameProcessors.get(0);
        VideoFrame outputFrame = firstProcessor.process(frame, textureHelper);
        if(outputFrame==null){
            mSink.onFrame(frame);
            frame.release();
            return;
        }
        mSink.onFrame(outputFrame);
        outputFrame.release();
        frame.release();

```
also, while looping through the list of processors shouldn't we use the last valid frame returned by a processor? Current code uses the original frame.

saghul:
@8BallBomBom are you ready with this?

8BallBomBom:
Should be good to go.
Left out changing the capturer to the base class but that could be addressed.

@anilgupta83 I never managed to reproduce the issue you experienced.

saghul:
@8BallBomBom Thanks! Can you run `npm run format` so the style is checked and then commit the result and land it?

8BallBomBom:
Done and done.

anilgupta83:
> @anilgupta83 I never managed to reproduce the issue you experienced.

Probably something wrong with my processor code, it has changed since then. I have tested the changes you have merged and it works as expected. Kudos!

ios/RCTWebRTC/VideoCaptureController.h
8BallBomBom: Left for now but can be changed if needed.

PR #1703, opened 2025-05-10
improve createOffer types and removed mandatory key from sessionConstraints in docs

mrakesh0608:
1. Removed mandatory key from sessionConstraints in docs, as the code directly reads from sessionConstraints, not from sessionConstraints.mandatory.
2. Align createOffer case as per web standard [w3 docs](https://www.w3.org/TR/webrtc/#dom-rtcofferoptions)
3. Added types to `createOffer`
4. Used typeof options !== 'object' instead of !{} while normalizing offer options, as !{} always evaluates to false

### DIFF
--- a/Documentation/BasicUsage.md
+++ b/Documentation/BasicUsage.md
@@ -232,11 +232,9 @@ That will allow you to enable and disable video streams on demand while a call i
 
 ```javascript
 let sessionConstraints = {
-	mandatory: {
-		OfferToReceiveAudio: true,
-		OfferToReceiveVideo: true,
-		VoiceActivityDetection: true
-	}
+	offerToReceiveAudio: true,
+	offerToReceiveVideo: true,
+	voiceActivityDetection: true
 };
 ```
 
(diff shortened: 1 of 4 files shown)


Documentation/BasicUsage.md
```diff
@@ -232,11 +232,9 @@ That will allow you to enable and disable video streams on demand while a call i
...
-	mandatory: {
-		OfferToReceiveAudio: true,
-		OfferToReceiveVideo: true,
-		VoiceActivityDetection: true
-	}
+	offerToReceiveAudio: true,
+	offerToReceiveVideo: true,
+	voiceActivityDetection: true
 };
 ```
```
mrakesh0608, line 239: Can we remove `sessionConstraints` usage from the docs, since these properties are deprecated ?

saghul: No. They are legacy options but supported.

saghul (approved):

saghul (approved):

Issue #1700, opened 2025-05-04
RN 0.73+ hidden compatibility issue: "Property 'addEventListener' does not exist on type 'RTCPeerConnection'"


dprevost-LMI:
The issue was first raised with [this issue](https://github.com/react-native-webrtc/react-native-webrtc/issues/1544#issuecomment-2823262769), but it was closed following a workaround using tsc change config, silencing the problem instead of fixing the root cause.

User [@IanMolldrem](https://github.com/IanMolldrem) in this [comment](https://github.com/react-native-webrtc/react-native-webrtc/issues/1544#issuecomment-2813652308) reproduces the issue and [provides a path-alternative here](https://github.com/react-native-webrtc/react-native-webrtc/issues/1544#issuecomment-2816034746) that is scoped to the rn-webrtc package instead of relying on the tcs config change that affects your entire project.

Here is a copy of the post since it is the heart of this issue:
> saghul Here's the sample project:
> https://github.com/IanMolldrem/WebRTCRepro
> 
> Just created a new RN project and installed this package and it is occurring as described.
> 
> Repro is evident through the TS errors here:
> https://github.com/IanMolldrem/WebRTCRepro/blob/main/App.tsx
> 
> The workaround detailed in my original issue description still works by uncommenting this:
> https://github.com/IanMolldrem/WebRTCRepro/blob/f6cb28305dbf0534259c43254ff3b80fc0f88ba7/tsconfig.json#L4

There was a mention about resisting this [issue as a fix](https://github.com/react-native-webrtc/react-native-webrtc/pull/1468), but I have no clue if it is the right approach!

dprevost-LMI:
Please keep open

AustinCase:
this is a tough issue. Happily add @ts-ignore, but that took quite a bit of extra time to get going on this lib, when things are already complicated trying to figure it out.

8BallBomBom:
Will be addressed in a future update but you aren't wrong.
Some things could be better and some issues can be annoying to resolve gracefully.

saghul:
124.0.8 is out and fixes this issue.

PR #1708, opened 2025-05-23
WIP: Change event handlers over to getters/setters

8BallBomBom:
Building on top of #1468 with type annotations added again, needs testing.
Might be time to revisit this as a possible solution to #1700 and the many other similar instances.


saghul:
I'm thinking perhaps we should vendor the dependency to avoid all of this.

Thoughts?

8BallBomBom:
A little unconventional but leaves room for potential improvement 🤔 
Also seems the type annotations still don't work...

8BallBomBom:
Will follow up on this shortly.

PR #1720, opened 2025-06-24
package.json: move husky to postinstall

iameli:
So `prepare` is pretty similar to `postinstall` but there are some differences, namely:

> NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.

So the existing husky command has some problems running if you're trying to `npm install` a forked version of react-native-webrtc. Moving it to `postinstall` causes the behavior to be identical for developers working on this repo while allowing forked versions to be able to run the `prepare` script correctly.


iameli:
Hm nope this doesn't actually help.

PR #1763, opened 2025-10-07
ios: fix crash in iOS 26 simulator

saghul:


saghul:
Changed the approach because the problem is with the audio devices enumeration, but still kept the video check for consistency.

PR #1764, opened 2025-10-07
NSDictionary no longer allows inserting nil values on iOS 26 sims

ghost:
This change addresses a crash encountered using Xcode 26.0 / iOS 26 simulators on macOS 15.6.1 or higher.


saghul:
See https://github.com/react-native-webrtc/react-native-webrtc/pull/1763

PR #1788, opened 2026-02-24
ios: Implement ScreenCapturePickerViewManager.showAwaiting

davidliu:
This adds a separate `showAwaiting` method to the ScreenCapturePickerViewManager, which blocks the async call until the broadcast has started, or rejects after a timeout of 5 seconds after dismissal (the broadcast only starts after a 3 second countdown).

Uses this to detect the dismissal: https://stackoverflow.com/a/76621153/295675

The old `show` method only shows the picker, and there's no way of knowing if the broadcast has started (and subsequently be left hanging on whether a subsequent `getDisplayMedia` call will succeed in getting the screen capture or not). This new API allows the SDK consumer to be more informed on whether the user has elected to start the broadcast.

### DIFF
--- a/ios/RCTWebRTC/ScreenCapturePickerViewManager.m
+++ b/ios/RCTWebRTC/ScreenCapturePickerViewManager.m
@@ -7,6 +7,39 @@
 
 NSString *const kRTCScreenSharingExtension = @"RTCScreenSharingExtension";
 
+/**
+ * Clears observers registered by showAwaiting
+ */
+void clearObservers(NSMutableDictionary *promiseDict) {
+    
+    if (!promiseDict) {
+        return;
+    }
+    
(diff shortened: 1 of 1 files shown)


EdgarJMesquita:
This implementation is simpler than ours, the focus observer is a nice touch. 

We ended up creating a custom module that listens to the `iOS_BroadcasrStarted` event from DarwinNotificationCenter.

We also implemented a promise based timeout with 20 seconds in the JS side.


With the focus observer and [this](https://github.com/flutter-webrtc/flutter-webrtc/blob/main/common%2Fdarwin%2FClasses%2FFlutterRTCDesktopCapturer.m#L60-L75), `getDisplayMedia()` can achieve a totally seamless integration, just like it's android counterpart. 
The developer wouldn't need to conditionally render the `RPSystemBroadcastPickerView` or call `show/showAwaiting`.

PR #1793, opened 2026-03-01
feat: add RTCCertificate support for WebRTC

jiyeyuran:
- implement RTCCertificate class and add support for certificate configuration in RTCPeerConnection
- add generateCertificate method to create certificates with fingerprints

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCModule.java
@@ -24,6 +24,11 @@ import com.facebook.react.modules.core.DeviceEventManagerModule;
 import com.oney.WebRTCModule.webrtcutils.H264AndSoftwareVideoDecoderFactory;
 import com.oney.WebRTCModule.webrtcutils.H264AndSoftwareVideoEncoderFactory;
 
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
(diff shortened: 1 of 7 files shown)


jiyeyuran:
@saghul could you help check this PR?

The motivation for this PR is to make transport setup more efficient for mediasoup-based integrations.

A mediasoup transport typically needs to be created first and then connected in a separate step, because the DTLS fingerprints are only available when the transport is connected. If `RTCCertificate` can be provided at `RTCPeerConnection` creation time, the DTLS certificate becomes known up front, and the corresponding fingerprints can be derived immediately.

With that, the fingerprints can be sent together with the transport creation request, which removes the need for a separate transport `connect()` request just to exchange DTLS parameters.

So the main benefit of this PR is that it enables earlier availability of DTLS fingerprints, simplifies the signaling flow, and saves one extra request during mediasoup transport setup.

jiyeyuran:
The `certificates` can also be provided when constructing `RTCPeerConnection`, since it's part of the standard constructor config. https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/RTCPeerConnection#certificates

saghul (approved):
Looking good! I'd like an extra pair of eyes here. @davidliu , @8BallBomBom ?

8BallBomBom:
It looks pretty good and a clean implementation, will run some tests 🤔 
Slightly out of spec when it comes to handling invalid key types but the silent drop and default is acceptable.

8BallBomBom (approved):

jiyeyuran:
Hi, @saghul. Since the comments have been addressed, would you mind taking another look and merging if everything looks good? Thanks!

PR #1810, opened 2026-05-26
Fix crashing android camera after spamming toggle

Magmusacy:


saghul:
Can you get rid of the unnecessary commits?

Magmusacy:
sorry i've made this pr by a mistake, im closing it

saghul:
Disappointed to see division rather than contribution.

Magmusacy:
sorry but this last commit with my fix was tailor made for our usecase and breaks other functionality, which we don't plan on supporting anyway if it wasn't for that I would gladly contribute to your repo ❤️

PR #1812, opened 2026-06-04
ios: trigger broadcast picker from retained view on new arch

Calinteodor:


ios/RCTWebRTC/ScreenCapturePickerViewManager.m
```diff
@@ -30,6 +31,16 @@ - (NSString *)preferredExtension {
 }
 
 RCT_EXPORT_METHOD(show : (nonnull NSNumber *)reactTag) {
+    if (RCTIsNewArchEnabled()) {
+        // On the New Architecture this view is mounted via the Fabric interop
```
saghul: Come to think of it, since we already hold on to a `RPSystemBroadcastPickerView` in `_broadcastPickerView`, why don't we use this code path for both old and new arch? We don't need UI manager to find the view, since we already have a native reference to it.

saghul (approved):

PR #1813, opened 2026-06-16
Hiroshi/fix adm indefinite lock

hiroshihorie:


hiroshihorie:
Opened against the wrong base by mistake. Reopening on the LiveKit fork.

PR #1814, opened 2026-06-17
feat: add nativeReceivedEpochMs / nativeReceivedMonotonicMs to data channel onmessage

harsuc:
### Problem

When measuring one-way latency through a WebRTC data channel, the natural approach is to
compare the sender's broadcast timestamp (embedded in the payload) against the time the
message was received on the other end. The only available hook for that is the JS-side
`onmessage` callback — but that callback fires on the JS thread, which can be scheduled
**hundreds of milliseconds after the OS actually delivered the packet** under UI or bridge
load. Critically, this error is correlated with high-load conditions: the moments where
latency is worst are exactly the moments where the JS timestamp is least accurate.

### Solution

Stamp the receive time in native code — at the point where the OS hands the buffer to the
app — before the native→JS bridge crossing. Two values are captured per message:

| Field | Type | Source | Meaning |
|---|---|---|---|
| `nativeReceivedEpochMs` | `number` | native | Wall-clock epoch ms. Directly subtractable from a server broadcast timestamp for one-way latency. |
| `nativeReceivedMonotonicMs` | `number` | native | Device-monotonic ms (`elapsedRealtimeNanos` / `CACurrentMediaTime`). Accurate for inter-arrival jitter; not comparable across devices. |

Both fields are forwarded through the native event and copied onto the `MessageEvent`
dispatched by `RTCDataChannel`. They are absent on web — consumers must guard with
`typeof ev.nativeReceivedEpochMs === 'number'` before using them.

### Changes

- **Android** (`DataChannelWrapper.java`): captures `System.currentTimeMillis()` and
  `SystemClock.elapsedRealtimeNanos()` as the very first statements in `onMessage`, before
  any buffer copy or encoding work, and emits both as `nativeReceivedEpochMs` /
  `nativeReceivedMonotonicMs` in the event map.

- **iOS** (`WebRTCModule+RTCDataChannel.m`): captures `[NSDate date]` epoch and
  `CACurrentMediaTime()` as the very first statements in
  `dataChannel:didReceiveMessageWithBuffer:` and includes both in the event dictionary.
  Requires `QuartzCore` (already available in all RN apps).

- **JS** (`RTCDataChannel.ts`): when the native fields are present on the raw event,
  copies them onto the `MessageEvent` before dispatching. No change in behaviour when
  the fields are absent (web, or older native builds).

### Backwards compatibility

Purely additive. No existing fields are changed. Consumers that do not read the new fields
are unaffected.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/DataChannelWrapper.java
+++ b/android/src/main/java/com/oney/WebRTCModule/DataChannelWrapper.java
@@ -1,5 +1,6 @@
 package com.oney.WebRTCModule;
 
+import android.os.SystemClock;
 import android.util.Base64;
 
 import androidx.annotation.Nullable;
@@ -59,6 +60,8 @@ class DataChannelWrapper implements DataChannel.Observer {
 
     @Override
     public void onMessage(DataChannel.Buffer buffer) {
+        long epochMs = System.currentTimeMillis();
+        double monotonicMs = SystemClock.elapsedRealtimeNanos() / 1_000_000.0;
         WritableMap params = Arguments.createMap();
         params.putString("reactTag", reactTag);
         params.putInt("peerConnectionId", peerConnectionId);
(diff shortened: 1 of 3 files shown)


saghul:
This is non-standard isn't it?

saghul:
No follow-up, closing.

PR #1820, opened 2026-07-04
make RTCPeerConnection.close set signalingState to "closed"

fippo:
which unlike connectionState and iceConnectionState is apparently not propagated by the native C++ layer


fippo:
(wait a bit, it seems the real bug is elsewhere...)

PR #1822, opened 2026-07-04
resolve RTCView rendering and layout issue

sanchit0160:
The following areas were updated to prevent the black rendering issue:

Constructor Setup: Unclipped children/padding and updated the SurfaceViewRenderer initialization to set hardware scaling and media overlay.
Layout Fallbacks: Added logic in onLayout to handle 0 dimensions by falling back to measured sizes or parent view dimensions, as well as preserving real surface area on the first frame.
Measure & Size Overrides: Implemented onSizeChanged and onMeasure overrides to properly relay dimension changes to the Android native layout system.
Relayouts: Injected requestLayout() and requestSurfaceViewRendererLayout() in callbacks (onFirstFrameRendered, setStreamURL, setVideoTrack) to enforce layout refreshes.

### DIFF
--- a/android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
+++ b/android/src/main/java/com/oney/WebRTCModule/WebRTCView.java
@@ -157,8 +157,20 @@ public class WebRTCView extends ViewGroup {
     public WebRTCView(Context context) {
         super(context);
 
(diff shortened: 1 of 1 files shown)


saghul:
What issues exactly are you fixing?

Note that enabling HW acceleration has created issues in the past, we cannot have it on by default.

PR #1823, opened 2026-07-08
ios(RCTWebRTC): enable Center Stage for devices that support it

Calinteodor:
Need help with testing on an iPhone 17, testing on iPhone 13 Pro went well.

### DIFF
--- a/.github/workflows/ios_ci.yml
+++ b/.github/workflows/ios_ci.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   ios-compile:
-    runs-on: macos-latest
+    runs-on: macos-15
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ios_ci.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   ios-compile:
-    runs-on: macos-latest
```
Calinteodor, line 14: A few days ago gh has promoted macos-latest to macОS 26 image and that's the reason the CI failed.

saghul:
I have an iPhone 17, I can test next week.

saghul (approved):

Issue #1838, opened 2026-09-03
iOS: RTCRtpSender.setParameters({ degradationPreference }) is silently ignored, and getParameters() throws on the JS round-trip


dlippy:
### Summary                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
  On iOS, `degradationPreference` is broken in both directions:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
  1. **Setting it is silently ignored** — the encoder never changes its degradation                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
     behavior (e.g. `MAINTAIN_RESOLUTION` still sheds resolution under load).                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
  2. **Reading parameters throws** — `getParameters()` (and thus the round-trip that                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
     `setParameters()` performs internally) rejects with a JS TypeError, so even a                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
     correct call never applies.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  ### Environment                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
  - react-native-webrtc: 124.0.7                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  - Platform: iOS                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
  - react-native: 0.77.0                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  ### Root cause                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  `RTCRtpParameters.degradationPreference` is an `NSNumber` wrapping the                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
  `RTCDegradationPreference` enum, but the JS bridge speaks in strings.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
[...]

PR #1841, opened 2026-09-21
fix(ios): convert degradationPreference between its name and the enum

sascha-kaleidoscode:
## Summary

On iOS, `setParameters` silently ignores every `degradationPreference` and applies **no adaptation** instead. Any app on iOS asking for `balanced`, `maintain-framerate` or `maintain-resolution` has been getting `disabled`.

Android is unaffected.

Fix for #1838 

## Cause

`RTCRtpSendParameters.toJSON()` sends the preference as a name:

```js
static toNative(format) { return format.toUpperCase().replace('-', '_'); }  // "BALANCED"
```

But the native property is a number — `RTCRtpParameters.h`:

```objc
@property(nonatomic, copy, nullable) NSNumber *degradationPreference;
```

and `WebRTCModule+Transceivers.m` assigned the string straight onto it:

```objc
params.degradationPreference = [options objectForKey:@"degradationPreference"];
```

The property is `copy`, and `NSString` implements `copyWithZone:`, so the assignment succeeds with no warning. The enum is then read out of an `NSString` with `intValue`, which is `0` for any non-numeric string — and `0` is `RTCDegradationPreferenceDisabled`.

So every preference collapses to the same value. `maintain-resolution` appears to work, but only because nothing is ever adapted.

The reverse direction needs fixing too: `parametersToJSON` handed the raw `NSNumber` back, and `RTCRtpSendParameters` runs whatever arrives through `fromNative`, which calls `.toLowerCase()` on it. That throws a `TypeError` on every `getParameters()` — latent until the value is actually set, which it never was.

## Fix

Two conversion helpers on `SerializeUtils`, used in both directions.

## How this showed up

Measured on a video call from an iPhone before the change: the encoder held 720×1280 at 30fps and let quantisation climb to qp 50 rather than reducing either, while leaving a third of its bitrate allowance unspent — because it had been told never to adapt. Setting `maintain-resolution` and setting `balanced` produced identical behaviour, and `getParameters()` read back `null` in both cases.

## Why Android is fine

`SerializeUtils.java` does a real conversion and throws on an unrecognised name, which is presumably why this survived: it works on Android.

```java
rtpParams.degradationPreference =
    RtpParameters.DegradationPreference.valueOf(updateParams.getString("degradationPreference"));
```

## Testing

- Verified against LiveKit's fork of this library (same code) in a mediasoup-based app: `getParameters()` now reads back the preference that was set, where it previously returned `null`.
- `clang-format` applied with the repo's `.clang-format`.
- Not covered by an automated test — I couldn't find an existing pattern for testing the ObjC bridge in this repo. Happy to add one if you can point me at the right place.


saghul:
Is android not affected?

sascha-kaleidoscode:
Correct, Android isn't affected - only iOS. There's a bit more detail under "Why Android is fine" in the description, but short version: the Java bridge does a real `valueOf` conversion and round-trips properly, so it's only the ObjC side that fails quietly.

